PR #188, opened 2025-11-04
Fix the unknown within the schema for chat : android

shubhammalhotra28:
Previously analytics were reporting 'unknown' for model_id because the code was checking resolvedOptions.model (which is often null). Now correctly uses currentModel?.model?.id to track the actual model being used for generation.

Tested and the supabase is getting the correct record for the chat which was recieving null earlier.

## Summary by CodeRabbit

* **Bug Fixes**
  * Analytics submissions now record model IDs (derived from the current model) instead of model names across generation and streaming flows.
  * Analytics logging is consistent across success/failure and streaming events; no changes to runtime behavior or error handling.


shubhammalhotra28:
@coderabbit review

shubhammalhotra28:
@coderabbit review

PR #473, opened 2026-04-14
Adopt llama_params_fit for model loading (rebased #444)

sanchitmonga22:
Reopens the intent of #444 (closed) on top of current `main`, with conflicts resolved and the Greptile-flagged issues fixed.

## Summary
- Replaces the fragile filename-heuristic GPU-layer limiter in `LlamaCppTextGeneration::load_model` with llama.cpp's `llama_params_fit`, which queries actual device memory to decide `n_gpu_layers` and `n_ctx`.
- Adds a configurable memory margin via `config["fit_margin_mib"]` (default 1024 MiB per device).
- Adds a configurable minimum context floor via `config["n_ctx_min"]` (default 2048).
- Caps batch size at 2048 (existing behavior, unchanged).
- Removes the post-load adaptive context ladder (`kLargeModelContextSize` et al.) since `llama_params_fit` now makes that decision based on real memory.

## Commits
1. `8508673` — VyasGuru's original fix (preserved).
2. `c03a9f6` — VyasGuru's merge of stale upstream/main (preserved).
3. `e00c21d` — merge of current `origin/main` with conflicts resolved: adapted `LOGI`/`LOGE` calls to main's `RAC_LOG_INFO("LLM.LlamaCpp", ...)` / `RAC_LOG_ERROR(...)` style, dropped the filename heuristic and adaptive-context ladder.
4. `1eb7542` — Greptile follow-up fixes (below).

## Greptile fixes applied on top
1. **CPU-only build no longer silently drops user `gpu_layers` override.** The user override used to be applied right before an unconditional `n_gpu_layers = 0` in the CPU-only preprocessor block, silently discarding the config value. The override now lives in the `#else` (GPU-build) branch; the CPU-only branch logs a warning when a non-zero user value is being ignored.
2. **`n_ctx_min` is now read from `config["n_ctx_min"]`** — previously the inline "Configurable parameter" comment didn't match behavior.
3. **Fit FAILURE / ERROR capping user context** now emits an explicit log line instead of silently clamping the user-requested `context_size` to 2048.
4. Restored trailing newline at EOF (was stripped by the original PR).

## CPU-only caveat (unchanged from #444)
`llama_params_fit` does not yet account for host memory on CPU-only builds. Upstream PR [ggml-org/llama.cpp#19711](https://github.com/ggml-org/llama.cpp/pull/19711) tracks this. Until that lands, CPU-only builds force `n_gpu_layers=0` and cap context at 4096.

## Test plan
- [ ] `cd sdk/runanywhere-commons && ./scripts/lint-cpp.sh`
- [ ] iOS build (Metal path, `GGML_USE_METAL` defined) — verify fit SUCCESS branch offloads layers correctly.
- [ ] Android build (CPU-only path) — verify `n_gpu_layers=0` is forced and a non-zero user `gpu_layers` logs a warning.
- [ ] Load a 7B+ GGUF on a mobile device and confirm it no longer OOMs at load time.
- [ ] Load a small (<1B) model and confirm context isn't unnecessarily clamped.

Co-authored-by: VyasGuru <71374747+VyasGuru@users.noreply.github.com>

Closes #444

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GPU-layer handling for CPU-only environments and removed misleading pre-load warnings; failures now fall back to safe defaults.

* **New Features**
  * Added config options for fitting margin and minimum context size, plus user overrides for GPU layers and context size that are applied via fitting.

* **Improvements**
  * Enhanced device-fitting workflow with clearer logging, conservative context-size clamping on error, and more reliable model-to-device resource allocation.

<h3>Greptile Summary</h3>

This PR replaces the fragile filename-heuristic GPU-layer limiter in `LlamaCppTextGeneration::load_model` with `llama_params_fit`, which queries real device memory to set `n_gpu_layers` and `n_ctx`, and adds configurable margin (`fit_margin_mib`) and minimum context floor (`n_ctx_min`).

- **P1**: In the GPU build path (`#else` block), when `llama_params_fit` returns `FAILURE` or `ERROR`, `model_params.n_gpu_layers` is correctly reset to 0 — but the immediately-following `if (user_gpu_layers >= 0)` check overwrites it with the user-supplied value without any warning. This silently restores GPU layers that the fit function just determined cannot fit in memory, re-introducing the OOM risk the PR was designed to eliminate.

<h3>Confidence Score: 4/5</h3>

Safe to merge after addressing the GPU-override/FAILURE interaction; the core llama_params_fit integration is sound.

One P1 issue: user gpu_layers override silently undoes the FAILURE/ERROR conservative n_gpu_layers=0 fallback in GPU builds, re-introducing OOM risk on constrained devices. Remaining findings are P2 style/robustness suggestions. The overall architecture improvement (real-memory-based fit vs. filename heuristics) is correct and well-structured.

sdk/runanywhere-commons/src/backends/llamacpp/llamacpp_backend.cpp — specifically the #else GPU-build override block (lines ~344-347)

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
| sdk/runanywhere-commons/src/backends/llamacpp/llamacpp_backend.cpp | Replaces filename-heuristic GPU-layer limiting with llama_params_fit; one P1 issue where a user gpu_layers override silently undoes the conservative FAILURE/ERROR fallback, potentially causing OOM. |

</details>

<h3>Flowchart</h3>

```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart TD
    A[load_model called] --> B[Read user_gpu_layers & user_context_size from config]
    B --> C[Initialize ctx_params with user context if provided]
    C --> D[Build margins, tensor_split, tensor_buft_overrides]
    D --> E[Call llama_params_fit]
    E --> F{fit_status?}
    F -->|SUCCESS| G[Log fitted n_gpu_layers & n_ctx]
    F -->|FAILURE| H[Set n_gpu_layers=0\nCap ctx to 2048]
    F -->|ERROR| I[Log error\nSet n_gpu_layers=0\nCap ctx to 2048]
    G --> J{CPU-only build?}
    H --> J
    I --> J
    J -->|Yes - no GPU backend| K[Force n_gpu_layers=0\nCap ctx to 4096\nWarn if user_gpu_layers>0]
    J -->|No - GPU build| L{user_gpu_layers >= 0?}
    L -->|Yes| M[Override n_gpu_layers=user_gpu_layers\neven after FAILURE/ERROR]
    L -->|No| N[Keep fit result]
    K --> O[llama_model_load_from_file]
    M --> O
    N --> O
    O --> P{model loaded?}
    P -->|No| Q[Return false]
    P -->|Yes| R[Compute final context_size_]
    R --> S[Create llama context]
    S --> T[Return true]
```

<details><summary>Prompt To Fix All With AI</summary>

`````markdown
This is a comment left during a code review.
Path: sdk/runanywhere-commons/src/backends/llamacpp/llamacpp_backend.cpp
Line: 344-347

Comment:
**User override silently undoes FAILURE/ERROR conservative fallback**

When `llama_params_fit` returns `FAILURE` or `ERROR`, the switch block explicitly sets `model_params.n_gpu_layers = 0` and logs "Proceeding with conservative CPU-only defaults." However, the `#else` block immediately overwrites this with `user_gpu_layers` — no additional warning is logged, and the failure outcome is silently ignored. A user who configured `gpu_layers: 24` will have 24 layers placed on-GPU even though fit determined the model cannot fit in device memory, which is the exact OOM scenario this PR aims to prevent.

At minimum, a `RAC_LOG_WARNING` should be emitted when the override is applied after a non-SUCCESS fit result. A stricter fix would skip the override unless fit succeeded:

```suggestion
    if (user_gpu_layers >= 0) {
        if (fit_status != LLAMA_PARAMS_FIT_STATUS_SUCCESS) {
            RAC_LOG_INFO("LLM.LlamaCpp",
                 "Applying user gpu_layers=%d override despite fit %s — risk of OOM",
                 user_gpu_layers,
                 fit_status == LLAMA_PARAMS_FIT_STATUS_FAILURE ? "FAILURE" : "ERROR");
        }
        model_params.n_gpu_layers = user_gpu_layers;
        RAC_LOG_INFO("LLM.LlamaCpp", "Applying user GPU layers override: %d", user_gpu_layers);
    }
```

How can I resolve this? If you propose a fix, please make it concise.

---

This is a comment left during a code review.
Path: sdk/runanywhere-commons/src/backends/llamacpp/llamacpp_backend.cpp
Line: 262-266

Comment:
**`tensor_buft_overrides` may need a sentinel entry**

`llama_max_tensor_buft_overrides()` returns the maximum capacity, but llama.cpp's internal iteration over the overrides array typically terminates on a zero-valued/null sentinel entry (the last element's `pattern` field must be `nullptr`). The vector is value-initialized to zero here, so the first element is already a valid sentinel — but only if no real overrides are ever written. It may be worth adding a brief comment clarifying that the all-zero initialization serves as the "no tensor overrides" sentinel, to make the intent obvious to future readers.

How can I resolve this? If you propose a fix, please make it concise.

---

This is a comment left during a code review.
Path: sdk/runanywhere-commons/src/backends/llamacpp/llamacpp_backend.cpp
Line: 367

Comment:
**Narrowing cast from `uint32_t` to `int` could silently truncate**

`ctx_params.n_ctx` is `uint32_t`; casting it to `int` is safe for any realistic context window, but it is an implicit narrowing conversion. A context value above ~2.1 billion would silently become negative and be picked by `std::min` as the smallest value, producing a near-zero context. Consider using `static_cast<int>` explicitly and/or clamping before the cast:

```suggestion
    context_size_ = std::min({static_cast<int>(std::min(ctx_params.n_ctx, static_cast<uint32_t>(INT_MAX))), model_train_ctx, max_default_context_});
```

How can I resolve this? If you propose a fix, please make it concise.
`````

</details>

<sub>Reviews (1): Last reviewed commit: ["Fix Greptile issues on top of llama\_para..."](https://github.com/runanywhereai/runanywhere-sdks/commit/1eb7542718b59b937a5ba91427d02b06b326cf77) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=28403929)</sub>

> Greptile also left **3 inline comments** on this PR.


sanchitmonga22:
@coderabbitai @greptileai please review this PR.

Context: this is a rebased and fixed-up version of #444 (closed). Key things to double-check:

1. **CPU-only build path** — the `#if !defined(GGML_USE_METAL) && !defined(GGML_USE_CUDA) && !defined(GGML_USE_WEBGPU)` block in `LlamaCppTextGeneration::load_model`. The previous PR silently dropped the user's `gpu_layers` override here; this version logs a warning and forces `n_gpu_layers=0`. Confirm the ordering is right and that GPU-build platforms still honor the override.
2. **`llama_params_fit` contract** — the `tensor_split`, `tensor_buft_overrides`, and `margins` buffers are sized to `llama_max_devices()` / `llama_max_tensor_buft_overrides()`. Flag if any are under-sized or misaligned.
3. **Context-size handling after FAILURE/ERROR** — when `llama_params_fit` fails, we cap `ctx_params.n_ctx` to 2048 and log the cap when the user requested more. Check for off-by-one or unsigned comparison issues.
4. **Config keys added**: `fit_margin_mib` (size_t, default 1024) and `n_ctx_min` (uint32_t, default 2048). Flag if these should live under a structured type rather than loose JSON keys.

Thanks.

PR #505, opened 2026-06-11
Telemetry V2 modality coverage, auth-gated flush, model state and voice fixes

Siddhesh2377:
### Telemetry (commons)
- Emit V2 telemetry for the remaining modalities: embeddings, voice agent turns, VAD, LoRA, and hybrid/cloud STT routing; per-modality payload extraction and JSON serialization to `/api/v2/sdk/telemetry/{modality}`.
- Stop streaming spam: per-token/per-chunk/per-progress events are gated out of telemetry; per-chunk STT stream events are public-only with one per-session summary at stream stop.
- Defer flushes until a JWT exists (pre-auth flushes 401'd and were silently dropped — the HTTP callback is fire-and-forget); 256-event queue cap; auth success kicks the deferred backlog.

### Telemetry dashboard consistency (commons)
- Drop unmapped proto arms instead of sending literal `"unknown"` event rows.
- VLM streamed generation reports `vlm.generation.*` (was mislabeled `llm.generation.*` under modality vlm); the stream's terminal generation event is public-only so each VLM operation produces exactly one telemetry pair.
- Model load/unload events route to their component's modality (`stt.model.load.*` under **stt**, not "model"); download/extraction/delete stay under **model**.
- Capability events (vlm/rag/embeddings/lora) carry `model_id`/`model_name` and duration (via envelope properties — no proto change), so the dashboard's Model and Latency columns populate.
- Uniform terminal status: any `*.completed`/`*.loaded`/`*.deleted`/`*.registered`/… without an error reports Success; `*.failed`/errors report Fail; started/progress/cancelled stay blank.
- Start/stop pairing everywhere: STT live-stream sessions emit a session-level started to pair with the stop summary; VAD stop only emits if a matching start ran; `sdk.init.started` no longer lost (see below).
- First-flush fix: the production "first flush to start timer" branch flushed the very first tracked event during Phase 1 — before the SDK's HTTP adapter was configured — and the batch was silently dropped (`sdk.init.started` lost on every launch). The batch timer now starts at manager creation so the earliest flush happens after the HTTP layer is ready.
- WARN (not silence) when a batch is dropped because JSON serialization failed.

### Auth (commons)
- When the backend rejects a token refresh (expired/rotated refresh token), clear the stale persisted auth state and fall back to full API-key authentication. Previously the device was stuck "authenticated" with dead tokens and every telemetry request 401'd forever. Transport failures (offline) keep tokens.

### Model state
- **Root cause fix:** the Android JNI never installed the populated platform adapter into commons (`rac_set_platform_adapter` was only called by the legacy `racInit`, which the Kotlin SDK's phase1-proto flow never invokes). With a NULL adapter, the registry's filesystem reconcile could never list directories (downloaded models showed a Download button on every relaunch) and commons native logging never reached logcat (an entire class of silent failures). One line in `racSetPlatformAdapter` installs it, mirroring Swift.
- Removed the `.registry.pb` disk-snapshot experiment; the save-time filesystem reconcile (now functional) is the mechanism, with the Phase-2 sweep as backstop. Reconcile misses log with the exact folder checked.
- Lifecycle load rejects models with no local artifact instead of passing a bare model id as a file path to engines; default Silero VAD auto-downloads via validate_availability.

### Voice / STT
- Sherpa backend reports streaming STT support (engines-refactor regression broke STT Live with -236).
- Fix C linkage for `racPlatformRegisterSystemTts`/`racPlatformUnregister` (declared after the extern "C" block close — name-mangled exports crashed app startup).
- Kotlin voice agent: mic driver + per-utterance turn streaming behind `streamVoiceAgent()`, off-main VAD bootstrap/init (main-thread JNI froze the UI), STT stream collection on IO, TTS cancellation passthrough + audio sink, VLM stream decode fix.
- Kotlin `ensureServicesReady`: HTTP/auth retry debounced (30s) and moved off the main thread.

### Example app (Android)
- Voice TTS picker excludes built-ins, registry relink after catalog seed, debug logging in debug builds, auto-load race gate.

## Testing
- Backend endpoints for the new modalities tested live against the dev Railway deployment (200/dedup/422 paths).
- On-device (Android, verified via adb + dashboard): downloaded models show "Use" after relaunch (5/5 linked at seed); telemetry startup 401 gone; no `unknown` rows; started/completed pairs for every modality including system; Model + Latency columns populated for vlm/rag; one row per VLM stream; STT Live restored.

## Known follow-up
- Telemetry batches that hit a transport failure (e.g. DNS outage) are still fire-and-forget dropped; re-queueing needs the HTTP callback to report status back to C++ (small ABI addition, separate PR).

## Release Notes

* **New Features**
  * Added a voice-agent microphone driver with segmented turn detection and streamed native processing.
  * Enabled streaming speech-to-text for compatible Sherpa backends.

* **Bug Fixes**
  * Prevented early voice cancellations during the new “Starting” phase.
  * Improved model availability validation to avoid confusing backend failures.
  * Fixed/paired lifecycle telemetry emissions across voice and STT flows.

* **Improvements**
  * Upgraded telemetry reporting (routing, batching, and richer lifecycle/capability metrics) across STT/TTS, RAG, embeddings, VAD, and LoRA.
  * Improved HTTP/auth recovery and error logging.

* **Documentation**
  * Updated Android example setup to use Maven Local SDK/engine modules and `local.properties` configuration.

* **Chores**
  * Updated example dependency versions and git ignore rules.


sanchitmonga22:
@coderabbitai please do a detaled review

sanchitmonga22:
@cursor please do a detailed review of the PR

Siddhesh2377:
@coderabbitai please do a detaled review

@cursor please do a detailed review of the PR

Siddhesh2377:
Everything's pushed (HEAD 40a0c1334 == remote). The key thing: the senior reviewed against an earlier state — two commits landed after the follow-up they saw (b195c0242, 40a0c1334), and #9 was fixed
  back in a7254b322. Here's the reconciliation with evidence so they can confirm and merge.

  ---
  Heads-up for @Siddhesh2377: please pull — current HEAD is 40a0c1334. The branch has 5 review-fix commits, not just the one follow-up.
  
  #1 — rules.txt ✅  Removed from the tree in 535ae8237 (git ls-files sdk/runanywhere-commons/rules.txt → empty). The copy you have open in the IDE is the untracked local copy — I used git rm --cached
  so it leaves the repo but stays on your disk. It's not in the PR. Say the word if you want it deleted locally too.
  
  #2 — duplicate VAD_STARTED ✅  Confirmed false positive, with evidence:
  - The lifecycle vtable is the engine plugin's ops: model_lifecycle.cpp:570 → loaded->vad_ops = vt->vad_ops. So ref.ops->start is the backend plugin's start (Sherpa/energy/Silero) — plugins don't 
  emit SDKEvents; commons does. 
  - rac_vad_component_start (the only other VAD_STARTED emitter) is never assigned to any vtable slot — grep finds zero function-pointer assignments; it's a standalone C-ABI entry point, not a vad_op.
  So ref.ops->start can't be "backed by" it.
  - Net: the lifecycle path emits VAD_STARTED exactly once. Removing it would re-break VAD telemetry on the Android/JNI path.
  
  #8 — endpoint mismatch ✅  Fixed in b195c0242 (after your review). The sdk_devices table genuinely exists (alembic/002_add_build_tokens_and_sdk_devices.py), so the C header's /rest/v1/sdk_devices is
  right; I fixed the Kotlin fallback to match (dev /rest/v1/sdk_devices, prod /api/v1/devices/register). Plus 40a0c1334 fixed the model-assignments fallback I found alongside it.
  
  #9 — libc++ pickFirsts ✅ Fixed in a7254b322) — pickFirsts += "**/libc++_shared.so" is in examples/android/.../app/build.gradle.kts.
  
  #10 — .stopped arm ✅ Verified: telemetry_manager.cpp:987  includes .stopped in the success branch; vad.stopped/stt.transcription.completed → success, stt.transcription.failed → failure, and
  started/partial/speech.*/metrics correctly stay unset. No misclassification.

  #15 — LoRA base_model_id ✅ Verified — not sourced from adapter ID.  The LoRA publish_capability(kind, op, error) sets no model_id, so base_model_id is currently null (telemetry_json.cpp:301), which
  the backend's nullable column accepts. Design ("base model rides on model_id") is correct; the only gap is that it's never populated — a pre-existing enhancement, not a regression.

  --- 
  
  
  @coderabbitai please do a detailed review

Siddhesh2377:
Everything's pushed (HEAD 40a0c1334 == remote). The key thing: the senior reviewed against an earlier state — two commits landed after the follow-up they saw (b195c0242, 40a0c1334), and #9 was fixed
  back in a7254b322. Here's the reconciliation with evidence so they can confirm and merge.

  ---
  Heads-up for @Siddhesh2377: please pull — current HEAD is 40a0c1334. The branch has 5 review-fix commits, not just the one follow-up.
  
  #1 — rules.txt ✅  Removed from the tree in 535ae8237 (git ls-files sdk/runanywhere-commons/rules.txt → empty). The copy you have open in the IDE is the untracked local copy — I used git rm --cached
  so it leaves the repo but stays on your disk. It's not in the PR. Say the word if you want it deleted locally too.
  
  #2 — duplicate VAD_STARTED ✅  Confirmed false positive, with evidence:
  - The lifecycle vtable is the engine plugin's ops: model_lifecycle.cpp:570 → loaded->vad_ops = vt->vad_ops. So ref.ops->start is the backend plugin's start (Sherpa/energy/Silero) — plugins don't 
  emit SDKEvents; commons does. 
  - rac_vad_component_start (the only other VAD_STARTED emitter) is never assigned to any vtable slot — grep finds zero function-pointer assignments; it's a standalone C-ABI entry point, not a vad_op.
  So ref.ops->start can't be "backed by" it.
  - Net: the lifecycle path emits VAD_STARTED exactly once. Removing it would re-break VAD telemetry on the Android/JNI path.
  
  #8 — endpoint mismatch ✅  Fixed in b195c0242 (after your review). The sdk_devices table genuinely exists (alembic/002_add_build_tokens_and_sdk_devices.py), so the C header's /rest/v1/sdk_devices is
  right; I fixed the Kotlin fallback to match (dev /rest/v1/sdk_devices, prod /api/v1/devices/register). Plus 40a0c1334 fixed the model-assignments fallback I found alongside it.
  
  #9 — libc++ pickFirsts ✅ Fixed in a7254b322) — pickFirsts += "**/libc++_shared.so" is in examples/android/.../app/build.gradle.kts.
  
  #10 — .stopped arm ✅ Verified: telemetry_manager.cpp:987  includes .stopped in the success branch; vad.stopped/stt.transcription.completed → success, stt.transcription.failed → failure, and
  started/partial/speech.*/metrics correctly stay unset. No misclassification.

  #15 — LoRA base_model_id ✅ Verified — not sourced from adapter ID.  The LoRA publish_capability(kind, op, error) sets no model_id, so base_model_id is currently null (telemetry_json.cpp:301), which
  the backend's nullable column accepts. Design ("base model rides on model_id") is correct; the only gap is that it's never populated — a pre-existing enhancement, not a regression.

  --- 
  
  
  @coderabbitai please do a detailed review

Issue #272, opened 2026-01-12
[Bug]: ios swift app is not deleting models once downloaded.


shubhammalhotra28:
### Component

iOS Sample

### Bug Description

The iOS Sample APP doesn't delete the Models once downloaded, if the user goes to the settings page and try to delete them. 
The action doesn't seem to do anything. 

Additionally, for some models it says that download space taken by some models is "zero kb" - which is incorrect.

### Steps to Reproduce

1. Build Sample APP
2. Try to play around chat / STT and TTS
3. On settings page try to delete the models 

### Expected Behavior

Models shoudl be deleted if the user is deleting them, additionally the metadata should be correct

### Environment

iOS : Iphone 16 pro max (iOS version : 26.1)

### Logs/Screenshots

<img width="660" height="1434" alt="Image" src="https://github.com/user-attachments/assets/77e7d9a4-7ad0-462c-9c19-ed72d44f557f" />

Siddhesh2377:
Closing as stale for this repo, on two grounds.

The iOS sample no longer lives here. It was extracted to [RunanywhereAI/runanywhere-ios](https://github.com/RunanywhereAI/runanywhere-ios) in the 0.20.17 restructure, so a sample-app bug belongs there.

The delete path in the current app is also wired correctly. `StorageViewModel.deleteModel` calls the SDK, surfaces the failure instead of swallowing it, and refreshes both views:

```swift
try await RunAnywhere.models.delete(id: model.id)
...
await refreshData()
await ModelListViewModel.shared.loadModelsFromRegistry()
```

That last line is there because the Models tab is a separate singleton with cached rows, which is exactly the "deleted model still shows as installed" symptom this issue describes. Both the Settings and Models entry points route through `RunAnywhere.models.delete(id:)`.

I verified the code, not the running app, so if delete still does nothing on a build of the current `runanywhere-ios` main, please reopen there with the version. The "zero kb" half should go in the same report, since per-model size comes from the SDK's storage analyzer and is worth checking separately.

PR #567, opened 2026-07-19
fix(commons): exclude builtin:// pseudo-models from storage analyzer results

ayaangazali:
## What

Partial fix for #272, the "Zero KB" half. The Models tab refresh half was already fixed on main in 9b77738bc, this handles the weird 0 KB rows that show up in storage and can never be deleted.

## Why

Built-in platform pseudo-models (Apple Foundation Models, System TTS, CoreML Diffusion) register with a `builtin://` local_path and have no files on disk:

- `rac_backend_platform_register.cpp:599,647,695` sets `local_path = strdup("builtin://...")`
- `rac_model_registry_get_downloaded()` (model_registry.cpp:646) returns any model with a non-empty local_path, so these come back as "downloaded"
- `rac_storage_analyzer_analyze()` then reports them as 0-byte downloaded models, and the delete-candidate path offers deletes that can never succeed

That is exactly the 0 KB undeletable rows from #272. I kept the fix inside the storage analyzer rather than filtering `get_downloaded()` itself, since the router intentionally accepts `builtin://` URIs elsewhere (rac_plugin_entry_platform.cpp:49) and other consumers rely on those entries existing.

## How

Single file, `storage_analyzer.cpp`:
- new `is_builtin_pseudo_model()` helper (prefix check on `builtin://`)
- metrics loop skips them and compacts the output array, so `model_count` reflects only real on-disk models
- delete-candidate collection skips them too

The metrics array is calloc'd, so the trailing unused slots are zeroed and the free path (which iterates `model_count`) stays correct.

## Testing

- `storage_analyzer_proto_tests` pass (macOS debug build)
- `test_core` passes
- touched line ranges are clang-format clean

---

Honesty note: I'm a college freshman doing these contributions to learn. I traced the root cause and wrote the logic myself and sanity-checked decisions with Claude Code along the way. If I misjudged the layer this belongs at or missed a consumer of those pseudo-model entries, I'd genuinely like to know. Happy to adjust :)

## Summary by CodeRabbit

* **Bug Fixes**
  * Built-in models are now excluded from storage usage calculations.
  * Built-in models no longer appear as candidates for deletion.
  * Storage analysis results now report accurate model counts when built-in models are present.

### DIFF
--- a/sdk/runanywhere-commons/src/infrastructure/storage/storage_analyzer.cpp
+++ b/sdk/runanywhere-commons/src/infrastructure/storage/storage_analyzer.cpp
@@ -64,6 +64,14 @@ int64_t clamp_non_negative(int64_t value) {
     return value < 0 ? 0 : value;
 }
 
+// Built-in platform pseudo-models (Apple Foundation Models, System TTS,
+// CoreML Diffusion) register with a "builtin://" local_path and have no
+// on-disk artifact, so they are not user-manageable storage entries.
+bool is_builtin_pseudo_model(const rac_model_info_t* model) {
+    return model != nullptr && model->local_path != nullptr &&
+           std::strncmp(model->local_path, "builtin://", 10) == 0;
+}
+
 float used_percent(int64_t total, int64_t used) {
     if (total <= 0 || used <= 0) {
         return 0.0f;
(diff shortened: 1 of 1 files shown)


sanchitmonga22:
Thanks @ayaangazali for adding this!

PR #560, opened 2026-07-17
Chat composer redesign and LLM benchmark timing fix

Siddhesh2377:
## What

Two changes.

**Chat composer (Android example)**
Splits the single input row into two. The top row holds the attach menu on the left and pushes tools, talk, and a new thinking toggle to the right. The bottom row is the input box and send button, so the composer feels more spacious. The thinking toggle is backed by the same setting as Settings > Show reasoning, stays disabled for models that do not support thinking, and keeps the no-thinking default.

**LLM benchmark timing (llamacpp)**
The non-streaming LLM path hardcoded time-to-first-token to 0 and reported end-to-end throughput as tokens/sec, which conflated prompt eval with decode. This brings it in line with the VLM backend: it brackets the first produced token, reports real TTFT, and reports decode-only tokens/sec. Both the returned result and the emitted telemetry pick up the corrected values, so the benchmark screen stops fabricating its decode vs prompt-eval breakdown.

## Notes

- Streaming was already correct and is untouched.
- The struct-based component path is left as is (not on the proto ABI the SDKs use), matching how the VLM path already works.
- Needs a commons Android rebuild and AAR restage to land on device.

---

> [!NOTE]
> **Low Risk**
> Changes are localized to example-app UI and inference telemetry; reported benchmark numbers will shift but generation behavior is unchanged.
> 
> **Overview**
> **LlamaCpp non-streaming LLM metrics** now split prefill from decode like the VLM path: `generate()` records the first streamed token to populate **TTFT** and **decode-only duration**, exposes those on `TextGenerationResult`, and the C API forwards real `time_to_first_token_ms` plus **tokens/sec computed from decode time** (not end-to-end) for `rac_llm_llamacpp_generate` and `generate_from_context`.
> 
> **Android example chat composer** is a two-row layout: attach menu plus tools, talk, and a new **thinking** toggle on top; text field and send/stop on the bottom. Thinking ties to `SettingsRepository.disableThinking`, is disabled when the loaded model lacks `supports_thinking`, and is wired through `ChatViewModel` / `ChatScreen`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 588588a424a0739786f0bd7f155cbcb9463efbd5. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Added Android chat controls to enable or disable model “thinking” when supported.
  - Added Talk mode and Thinking toggle controls to the chat input UI.
[...]

PR #569, opened 2026-07-19
SDK API parity, telemetry field fixes, LLM token-count fixes, and new rcli commands

Siddhesh2377:
This branch collects a few weeks of SDK/commons work. The threads are separate but landed together, so here's what's in it and why.

### SDK public API parity
Aligned the public API surface of Kotlin, React Native, Flutter, and Web to the Swift facade (the source of truth), keeping platform-genuine differences. Added flat `RunAnywhere.x` aliases where an SDK hid a feature behind a capability object. Per-SDK usage docs added under `docs/md/`. Web-search on RN/Flutter/Web and `setHfToken` on Web are deliberately left out — those SDKs route HTTP through native C++, so a JS/Dart HTTP tool fights the design.

### Telemetry: fill the missing fields
Several per-modality telemetry fields were arriving null. Fixed at the engine/commons level so every SDK benefits:
- LLM/VLM `prompt_eval_time_ms` (prefill was never populated on the unary result)
- embeddings `total_tokens` (real WordPiece count from the ONNX provider) and `batch_size`
- LoRA `base_model_id` / `adapter_id` / `adapter_size_bytes` on the failure path (they were dropped)
- RAG `query_token_count` / `context_tokens`

### LLM token undercount (this one actually mattered)
`tokens_per_second` and `tokens_generated` were under-reported by up to ~5x. The count came from streaming-callback invocations, which flush buffered chunks rather than one call per token. Now the unary path uses the decode loop's real count (the streaming path keeps the callback-based count to avoid a vtable ABI change that would ripple into every backend). Auditing the benchmark is what surfaced it — smollm2 reported 4.8 tok/s where the real number was ~45. Also added `sanitize_utf8` at the proto string boundaries, since a model cut off mid-multibyte-character produced invalid UTF-8 that broke `rac_llm_generate_proto` serialization entirely.

### rcli: bench, rag, lora
- `rcli bench` auto-benchmarks every installed model (LLM/STT/TTS/VLM), matching the Android benchmark screen: load, warmup, N trials, median. `RAC_BACKEND_RAG` turned on in the committed rcli presets (it also gates ONNX embeddings).
- `rcli rag query` runs a full ingest + query through the RAG session ABI.
- `rcli lora apply` loads an LLM and attaches an adapter.

### Telemetry tests
- `test_telemetry_extraction` (commons): drives SDKEvent protos through the extraction/routing/JSON pipeline against a mock sink and checks the per-modality endpoint + fields offline. Regression guard for every bug above, including the NaN-confidence one.
- `test_rcli_telemetry_live --live` (rcli): sends real authenticated telemetry per modality and asserts the backend returns 2xx. Opt-in — no-ops without `--live` and creds, so CI stays green.

### Keyless staging environment
`environment = staging` is now absolute and self-contained: whatever URL or API key the app passes, commons overrides the base URL with a baked staging backend URL and clears the key — requests go out unauthenticated and the backend attributes them to the PUBLIC org. The URL never lives in committed source: team builds read it from the git-ignored `development_config.cpp`, release builds substitute the `STAGING_BASE_URL` CI secret into the template (same mechanism as the Supabase dev creds; the secret needs creating in GitHub once). Build with `./run rcli build --env-staging` or `./run sdk commons build-android --env-staging`.

Platform SDKs were audited for places that trusted app-passed config over commons state — every HTTP adapter (Kotlin, Swift, Flutter, RN) now reads the effective base URL from commons and tolerates a missing key instead of silently skipping telemetry or, worse, attaching a stale key as a Bearer (RN did this and every staging event 401'd). Verified end to end on device: Android/Flutter/RN example apps init with `environment: staging` and nothing else, and per-modality events land in the staging console. New commons live test: `test_staging_keyless_live` (build-on-demand, excluded from ctest).

### Also
Android chat composer redesign (two-row layout + thinking toggle, from #560). `idl/sdk_events.proto` gained `GenerationEvent.prompt_eval_time_ms` (field 34); all five language bindings regenerated.

Verified on Linux: rcli builds and runs every modality end to end, telemetry confirmed landing in Supabase, both test suites pass.

## Summary by CodeRabbit

* **New Features**
  * Added CLI commands for authentication, telemetry, RAG queries, benchmarking, and LoRA adapter application.
  * Added keyless development setup with configurable production credentials and staging build support.
  * Added web search tooling and thinking-mode controls in the Android chat experience.
  * Added richer model, device, embedding, generation, vision, and RAG performance metrics.
* **Improvements**
  * Improved telemetry delivery with retries, device-state reporting, and safer JSON handling.
  * Expanded model management and storage APIs across Flutter, Kotlin, React Native, Swift, and Web.
* **Documentation**
  * Added cross-platform API guides, architecture references, and design-system guidance.

---

## Update: per-event device state + device registration overhaul

Follow-up work landed on this branch after review started.

### Every telemetry event now carries live device state
`sdk_binding` (which SDK sent it), battery level/state, low-power mode, total/available RAM, CPU usage since the previous event, and online core count — stamped by the telemetry manager on every tracked event. CPU accounting is read in-process (/proc, mach, GetSystemTimes), no platform hooks needed. Battery/RAM come from the device callbacks where those are thread-safe (JNI, desktop C) and from a push-model cache where they aren't — Dart FFI callbacks are isolate-bound, so Flutter pushes state from its own thread instead of being called into. That distinction was learned the hard way: sampling through the Dart callback from the inference thread crashed Flutter mid-generation.

### Telemetry retry
Failed batches now re-send with exponential backoff (5s/10s, 3 attempts max, bounded queues). Safe at-least-once — the backend dedupes on event id. Covered in test_telemetry_extraction (fail → no early resend → resend after backoff).

### Device registration actually works now
Three separate bugs made devices register as "Unknown / SDK Device":
1. Development builds sent a legacy flat registration JSON (direct-to-Supabase relic) that the current backend 422s. Registration is nested for every environment now, and the Supabase-era dev endpoint is retired.
2. The auth-time placeholder row was never upgraded when the client's persisted is_registered flag was stale (e.g. after a server-side reset). The authenticate response now returns device_registered:false for placeholder rows and commons forces a re-registration immediately after auth succeeds.
3. Device info itself was mostly fabricated. All five SDKs now report real values: marketing device names (Settings.Global on Android, lookup tables on iOS), Build.SOC_MODEL chips, Metal-queried GPU families, sysctl/cpufreq core splits, real battery (battery_plus added for Flutter), stable hardware fingerprints instead of device_id copies, and correct sdk_binding strings (react-native, cli).

### Example apps
Flutter/RN examples take dev credentials at build time (--dart-define / gitignored .env, mirroring Android's local.properties) and use the production env path with a custom base URL — the same proven flow the Android example uses.

### rcli
Now registers full device info on desktop (DMI/procfs/sysctl sources) and passed a live end-to-end check: real payloads verified with all new fields, correct nulls on a desktop (no battery), CPU delta ramping 12→39% during generation.

Backend/console counterpart (schemas, migration 023, analytics read path, SDK badges + device-state timeline in the trace view) is deployed on the monorepo development and staging branches.


shubhammalhotra28:
@coderabbitai - please do a detailed review for this PR! Thanks!

shubhammalhotra28:
### OSS keyless telemetry gate — proven

**Local blast** session `oss-ci-1784770421-30596` → 12/12 HTTP 200  
**Supabase MCP** (Staging analytics `fhtgjtxuoikwwouxqzrn`): all 12 base rows PUBLIC org, `processing_time_ms=42.5`, `sdk_binding=cli`, 10 child tables present, LLM tokens 128/256, required fields non-null  
**CI** https://github.com/RunanywhereAI/runanywhere-sdks/actions/runs/29972563501 → success, session `oss-ci-1784771595-21145`

Workflow PR to land schedule/dispatch on `main`: https://github.com/RunanywhereAI/runanywhere-sdks/pull/586 (blocked on review; needs this branch’s rcli on main first, or merge this PR then 586).

shubhammalhotra28:
@coderabbitai - please do a detailed review.
Thanks!

shubhammalhotra28:
Looks good to merge - making changes in internal PR for the backend - before merging this

shubhammalhotra28:
Just kept Production and Development as the mode of environments for the SDK entirely. Removed Staging completely!
At the SDK front^

shubhammalhotra28:
Looks good to merge - working on looking over the backend changes following this merge!

PR #570, opened 2026-07-19
rcli network driver + example-app brand alignment (#FF6900)

sanchitmonga22:
## rcli control-plane network driver + example-app brand alignment

Combines two SDK-repo changes into one PR.

### Part 1 — rcli control-plane network driver (`sdk/runanywhere-cli`)
Turns `rcli` into a control-plane client that can point at any backend (including localhost) for integration testing and manual debugging.
- Global `--base-url` / `--api-key` / `--environment` flags (+ `RUNANYWHERE_*` env vars), threaded into the SDK config with zero regression when absent.
- `rcli auth login` — real handshake (API key → JWT, device register, model-assignments fetch).
- `rcli telemetry emit --modality <m>` and `rcli telemetry blast` — model-free telemetry through the real network transport with a 12-modality result table.
- Proven end-to-end against a locally-run backend: login registers a device with real hardware facts; blast lands all 12 modalities as verified rows; a bad key exits non-zero. Existing commands unchanged (10/10 CLI tests).

### Part 2 — align the 5 example apps to the brand + design guideline
Every example app had drifted from the brand: their UI accents were the legacy `#FF5500` (Flutter/RN were plain blue) while their logos were already the correct `#FF6900`. This finishes the migration.
- **`examples/DESIGN_GUIDELINE.md`** — canonical brand doc (primary = the logo's #FF6900, gradient →#FB2C36, full palette light+dark, per-platform mapping, the white-on-orange contrast caveat).
- iOS (AppColors + AccentColor + Keyboard/Activity extensions), Android (Color.kt tonal ramp on the brand hue + dead template swatches removed), Flutter (mislabeled-blue primary → brand, seed pinned), Web (CSS tokens + loading SVG + theme-color), React Native (**two competing theme systems consolidated** onto one anchored to #FF6900; legacy barrel deleted; migrated screens now get dark mode).
- Each app's `AGENTS.md` gains a Design System section pointing at the guideline (CLAUDE.md symlinks intact); parent `AGENTS.md` references it.
- Removed the dead `rac_telemetry_manager_parse_response` declaration + stale export.

Grep-proven: zero legacy accent literals remain; #FF6900 present in all five apps.

## Summary by CodeRabbit

* **New Features**
  * Added CLI support for control-plane environments, authentication, device registration, and model assignments.
  * Added telemetry commands for sending events by modality or across all supported modalities, with JSON and human-readable results.
  * Improved theme-aware styling across the React Native app, including light and dark modes.

* **Style**
[...]

PR #571, opened 2026-07-19
rcli: control-plane network driver (base-url/api-key flags + telemetry commands)

sanchitmonga22:
## rcli control-plane network driver

Turns `rcli` into a real control-plane client so it can drive ANY backend (including localhost) for integration testing and manual debugging.

- Global `--base-url` / `--api-key` / `--environment` flags (+ `RUNANYWHERE_*` env vars), threaded into the SDK config with zero regression when absent.
- `rcli auth login` — real handshake (API key → JWT, device register, model-assignments fetch).
- `rcli telemetry emit --modality <m>` and `rcli telemetry blast` — model-free telemetry through the real curl transport, 12-modality result table.

Proven end-to-end against a locally-booted backend: login registers a device with real hardware facts; blast lands all 12 modalities (10 child + 2 base-only) as verified DB rows; bad key → clean 401. Existing commands unchanged (10/10 CLI tests). The `integration/` suite in the monorepo shells out to this.

### DIFF
--- a/sdk/runanywhere-cli/CMakeLists.txt
+++ b/sdk/runanywhere-cli/CMakeLists.txt
@@ -38,6 +38,7 @@ set(RCLI_SOURCES
     src/catalog/model_ref.cpp
     src/commands/cmd_version.cpp
     src/commands/cmd_info.cpp
+    src/commands/cmd_auth.cpp
     src/commands/cmd_backends.cpp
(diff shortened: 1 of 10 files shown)


sanchitmonga22:
Folded into #570 (combined single PR with both the rcli network driver and the example-app brand alignment). Continuing review there.

PR #574, opened 2026-07-20
llm: constrained tool-call decoding, parallel tool calls, real agent-loop wiring

yavuzkaanakyuz:
## Summary

Three additive upgrades to the C++ commons tool-calling core, plus a follow-up commit fixing a real bug found during live on-device testing and wiring parallel tool calls through to Swift. All commons logic lands behind the existing proto ABI, so every SDK (Swift/Kotlin/Flutter/RN/Web) inherits it from this one change.

### 1. Grammar-constrained tool-call decoding
The run loop now derives a decode-time grammar from the live tool set + `tool_choice` policy (new `tool_calling_grammar.{h,cpp}`) and threads it into `rac_llm_options_t.grammar` — the channel both engines already enforce:
- **llamacpp**: raw GBNF via the existing `llama_sampler_init_grammar` sampler chain.
- **qhexrt (Snapdragon NPU)**: the native `toolcall:`/`toolcall_opt:` GrammarKind logit mask.

Under `REQUIRED`/`SPECIFIC` the model can no longer emit a hallucinated tool name or malformed `<tool_call>` envelope; qhexrt additionally covers `AUTO` via its native free-chat-or-one-call kind. Scoped to the default JSON wire format (LFM2 keeps post-hoc validation); post-hoc `validate` stays on as a belt-and-suspenders check.

### 2. Parallel tool calls
Reclaims `ToolCallingOptions` field 15 as `bool parallel_tool_calls` (mirrored on `ToolCallingSessionCreateRequest`, field 20). When set, one model turn may emit multiple tool-call envelopes: the parse ABI returns them all and the run loop validates + executes the whole batch under the existing side-effect admission guard before building **one** follow-up prompt. Default off preserves the historical single-call behavior byte-for-byte. All language bindings regenerated.

**Follow-up fix (this PR, found via live testing on macOS):** the parallel parser had no bound on repeated envelopes. A small model (LFM2-350M in the example app) stuttering — repeating the identical tool call verbatim several times in one turn, a known small-model artifact — was harvested as N "legitimate" parallel calls, burning the `max_tool_calls` budget on duplicates and starving the rest of a compound request (observed: "weather + time" burning 3–4 calls on repeated `get_weather`, never reaching `get_current_time`). Fixed: the parser now stops harvesting the moment a call repeats the immediately-preceding `(name, arguments)` pair, and never harvests past the request's own `max_tool_calls` budget. New regression test (`test_parallel_parse_stops_at_repeated_call`) alongside the existing positive test proving genuinely distinct parallel calls still both parse.

`generateWithTools(...)` in the Swift SDK gained a `parallelToolCalls` parameter (same pattern as `toolChoice`/`forcedToolName`); the example app's chat tool-calling call site now opts in.

### 3. Declarative agent loop wired to the real tool run-loop
`expand_agent_loop` previously never read `AgentLoopConfig.tools` (and `GenerateTextNode` never consumed `max_iterations`), so a tools-bearing config silently ran as a plain tool-less generation. A tools-bearing config now maps to a new `agent_loop` operator that drives `rac_tool_calling_run_loop_proto` with the declared tools and iteration cap. Solutions pipelines have no host tool-executor callback yet, so the operator runs `auto_execute=false` and surfaces a requested call as an explicit structured error naming the tool. Tool-less configs keep the `generate_text` path unchanged.

### 4. Tool-description quality (example app + SDK)
Two small fixes surfaced by manual testing, not architecture changes:
- `search_web`'s description is now directive rather than descriptive — it explicitly names the trigger conditions (recency, post-cutoff events, prices, live data) an `AUTO` tool_choice decision reads to decide whether to call it. Not forced via `tool_choice=REQUIRED`, which would call it on every message regardless of need.
- `get_current_time`'s description and result payload now state explicitly that the returned time is already local and the tool has no location parameter, plus a `note` field restating that next to `utc_offset`. This heads off an observed small-model failure mode where the model re-applies `utc_offset` to an already-local timestamp (e.g. reporting 18:30 when the device's actual local time was 21:30).

## Verification
- New commons tests: `test_tool_calling_grammar` (21/21 — grammar dialects per tool_choice mode, parallel parse, and the repetition-guard regression) and `test_agent_loop_converter` (13/13 — op selection + tool param forwarding).
- Full existing tool-calling / solutions / pipeline suites pass (`ctest -R 'tool_calling|agent_loop'`: 7/7).
- `rac_commons` builds clean (macOS, Debug, backends + tests); touched files clang-formatted.
- **Exercised live**: built and ran the macOS example app end-to-end with a real on-device model (LiquidAI LFM2-350M via llama.cpp), which is how the repetition bug above was actually found and confirmed fixed.
- Not run: iOS/Android device builds and the qhexrt (Snapdragon NPU) grammar dialect (no NPU hardware in this environment) — CI's platform matrix covers those.

## Note for reviewers
This is proposed work from local testing — not merging it myself. Please review before merge.

## Summary by CodeRabbit

* **New Features**
  * Tool-enabled generation can now execute multiple tool calls in parallel.
  * Added optional Apple Health and Calendar tools for iOS, including authorization controls and readable results.
  * Added agent-loop support for solution pipelines with configured tools.
  * Improved tool-call reliability with constrained generation and clearer date and result-grounding guidance.
* **Improvements**
  * Expanded web search guidance and strengthened current-time instructions.
  * Added safer handling and messaging for health-related responses.
* **Bug Fixes**
  * Improved parsing, validation, and execution of multiple tool calls in a single response.


Siddhesh2377:
@yavuzkaanakyuz this needs a rebase on current main before we can take it.

Three conflicts, all in commons tool-calling:

```
CONFLICT (content): sdk/runanywhere-commons/src/features/llm/tool_calling.cpp
CONFLICT (content): sdk/runanywhere-commons/src/features/llm/tool_calling_generation_internal.h
CONFLICT (content): sdk/runanywhere-commons/src/features/llm/tool_calling_run_loop.cpp
```

All three trace to one commit that landed after you branched: `c7bc03199`, "Fix tool-calling / thinking-mode / system-prompt injection (RUN-47) (#589)". It threads `framework_name` through the generate and stream call sites and adds a QHexRT allowlist for thinking suppression. No design disagreement with your work, just edits to the same functions, so the merge should be mechanical.

Two things worth knowing while you rebase:

Thinking mode is being reworked at the IDL layer right now. `disable_thinking` is being replaced with a positive `enable_thinking` that defaults to false, and a `ThinkingPolicy` enum is being added to `model_types.proto`. Your PR reclaims `ToolCallingOptions` field 15 and mirrors it on `ToolCallingSessionCreateRequest` field 20, so both changes regenerate bindings. Whoever lands second regenerates on top. Worth pulling before you regenerate.

We are also holding NPU and iOS changes out of the current merge batch. That affects the qhexrt grammar dialect and the Swift `generateWithTools(parallelToolCalls:)` parameter here. The commons core stands on its own without either, so if splitting those out is easy, the rest can move now. If it is entangled, say so and we will look at it whole.

On the substance: the `expand_agent_loop` fix is the part I would not want to lose. A tools-bearing `AgentLoopConfig` silently running as plain generation is a real bug, not a missing feature. The repetition guard is a good catch too, and the reasoning for finding it live rather than in a test is sound.

Noted that you asked for review before merge rather than merging it yourself.

shubhammalhotra28:
@coderabbitai - please do a detailed review!
Thanks!

shubhammalhotra28:
Closing in favor of a maintained branch owned by the team:

**Branch:** `shubham/tool-calling-agents` (includes all commits from this PR)
**Next:** rebase/merge onto current `main` (this PR was CONFLICTING due to overlapping RUN-80 grammar work), then finish end-to-end and open a replacement PR.

Thanks @yavuzkaanakyuz for the work here — continuing from that tip.

shubhammalhotra28:
Superseded by branch `shubham/tool-calling-agents` (all commits preserved). Closing so work continues on that branch against current main.

shubhammalhotra28:
Superseded by the continuation PR (see linked PR from `shubham/tool-calling-agents`).

PR #588, opened 2026-07-23
Hybrid streaming voice agent (NPU LLM + CPU STT/TTS)

AmanSwar:
## Summary

Makes the RunAnywhere voice agent **hybrid + streaming**: the LLM runs on the Hexagon **NPU** (QHexRT) while STT/TTS/VAD run on **CPU** (sherpa). The single NPU slot holds only the LLM (stays warm all session — no per-turn Whisper↔LLM swap), CPU speech overlaps NPU decode, and the spoken reply **streams out sentence/clause-by-clause** as it is generated.

Device-validated on **S25 (SM8750, Hexagon v79)** via acoustic loopback across app builds v25→v29.

### Core: hybrid streaming pipeline
- **`voice_agent_d7_abi.cpp`** (commons) — refactors `d7_process_utterance` from blocking-generate + one-shot-TTS to streaming: per-token accumulate → clause chunking → per-sentence synth → incremental `AudioFrameEvent(is_final, chunk_index)`. Backward-compatible `out_result` WAV still returned.
- **`voice_agent_tts_chunking.h`** (commons, new) — header-only C++ clause-aware chunk policy (splits on `.!?` always, `,;:` past 50 chars). Header-only to avoid a CMake reconfigure.
- **`StreamingAudioPlayer.kt`** (SDK, new) — incremental `AudioTrack` `MODE_STREAM` + `ENCODING_PCM_FLOAT`; writer thread paces playback via blocking writes.
- **`RunAnywhereVoiceAgent.kt`** (SDK) — `streamVoiceAgent()` taps each `AudioFrameEvent` into the player.
- **`VoiceAgentMicDriver.kt`** (SDK) — removes the old one-shot WAV-blob playback (audio now incremental).

### App bug fixes
- **Setup downloads ALL models** (`VoiceScreen.kt`) — removes break-on-load-failure so a 2nd model failing to load on the single NPU slot no longer aborts the rest.
- **Recommends sherpa CPU Whisper** for the pipeline (`ModelRecommendation.kt`) — frees the NPU slot for the LLM → the hybrid streaming path (not the all-NPU per-turn-swap path). NPU Whisper is still preferred for standalone dictation.

### Robustness (found during device testing)
- **Clipped reply tail** → `StreamingAudioPlayer` drains the `AudioTrack` (polls the hardware playback head to the full written-frame count) before release, instead of `stop()`+`release()` back-to-back which truncated the last ~0.5s.
- **Echo / self-talk loop** → **`VoiceTtsPlaybackGate.kt`** (new) + `VoiceAgentMicDriver.kt`: the whole reply WAV is already synthesized when the blocking turn returns, so the mic-driver reads its duration and **deterministically mutes the mic** for that duration + tail the instant the turn returns — before reading another frame. Kills the device transcribing its own TTS.
- **Phantom-turn loop** (STT hallucinating `[BLANK_AUDIO]`/`[wind blowing]` on silence between utterances) → `RunAnywhereVoiceAgent.kt` mutes phantom-turn audio and `VoiceViewModel.kt` hides them from the UI (`transcriptIsNonSpeech`). The **commons-side** equivalent (skip the turn before the LLM runs) is also included in `voice_agent_d7_abi.cpp`.

## Device validation (S25 / v79)
- Setup downloads + loads all 4 models (sherpa Whisper, Qwen3.5-0.8B NPU, Piper, Silero).
- Listen = **Sherpa Whisper (CPU)**, not NPU.
- Reply audio streams out; `AudioTrack` delivers every synthesized frame (`frames delivered == samples generated`), including multi-clause streamed replies.
- Deterministic mic mute fires (`reply ~1834ms; muting mic through playback`) → only the user's question is transcribed, no self-talk.
- Stop button tears down cleanly (`mic capture stopped`).

## ⚠️ Reviewer notes
- **The commons `.cpp` change needs a native rebuild to take effect.** The fast app build path uses `-x buildLocalJniLibs` (prebuilt `.so`), so the streaming refactor + commons non-speech filter only compile in via `scripts/build/build-core-android.sh`. The Kotlin app-side phantom suppression works without it.
- **Overlap:** `origin/siddhesh/voice-agent-quality` rewrites several of the same files independently — expect a merge conflict to reconcile.
- Layer decision follows the existing proto contract (`VoiceEvent`/`AudioFrameEvent` already carry `chunk_index`/`is_final`), so other SDKs (Swift/Flutter/Python) inherit the streaming path; only mic-in + PCM playback stay platform-specific. The all-CPU path stays byte-identical (additive, no C-ABI break).

## Summary by CodeRabbit

* **New Features**
  * Added incremental, sentence/clause-based voice playback for quicker reply feedback.
  * Enabled streaming LLM→TTS so assistant audio is produced continuously during generation.
* **Bug Fixes**
  * Suppress non-speech/“phantom” transcripts to prevent empty or spurious assistant turns.
  * Voice model preparation is now best-effort across components, continuing even if one fails.
  * STT/TTS routing no longer prefers NPU paths (CPU-first behavior).
* **Improvements**
  * Added mic suppression during TTS playback to reduce self-capture/feedback loops.

### DIFF
--- a/examples/android/RunAnywhereAI/app/src/main/java/com/runanywhere/runanywhereai/ui/screens/models/ModelRecommendation.kt
+++ b/examples/android/RunAnywhereAI/app/src/main/java/com/runanywhere/runanywhereai/ui/screens/models/ModelRecommendation.kt
@@ -144,9 +144,19 @@ object ModelRecommendation {
         val byId = models.associateBy { it.id }
         val preferNpu = hasNpu && tier == HardwareTier.HIGH_END
         return VoicePipeline(
-            stt = pickAsr(preferNpu, hasNpu, budget, byId, models),
+            // The voice PIPELINE keeps STT on CPU (sherpa) even on NPU devices: an
+            // NPU STT would fight the NPU LLM for the single Hexagon slot, forcing a
+            // slow per-turn Whisper<->LLM swap. CPU STT leaves the slot for the LLM
+            // (which stays warm) and routes the turn through the streaming hybrid path
+            // (useNpuSwap=false). NPU Whisper is still preferred for standalone
+            // dictation via recommendedFor(STT); this falls back to NPU only if no
+            // CPU ASR is available.
+            stt = pickAsr(preferNpu = false, hasNpu, budget, byId, models),
             llm = pickLLMs(tier, preferNpu, budget, byId, models).firstOrNull(),
-            tts = pickTts(preferNpu, hasNpu, budget, byId, models),
+            // TTS also stays on CPU (sherpa piper) for the same reason as STT: an NPU
+            // TTS would contend with the NPU LLM for the single Hexagon slot. Only the
+            // LLM runs on the NPU in the hybrid pipeline.
+            tts = pickTts(preferNpu = false, hasNpu, budget, byId, models),
             // VAD is tiny + backend-neutral (ONNX Silero); include it so the pipeline is
             // fully pre-staged. The voice agent also auto-ensures it, so it's a nicety.
             vad = pickCategory(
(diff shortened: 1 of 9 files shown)


Siddhesh2377:
@AmanSwar conflicting with main. Three files, all the Android voice screen and mic driver. Commons side merges clean.

sanchitmonga22:
@coderabbitai review

PR #589, opened 2026-07-23
Fix tool-calling / thinking-mode / system-prompt injection (RUN-47)

AmanSwar:
## Summary

Fixes the tool-calling / thinking-mode / system-prompt-injection issues under epic **RUN-47**, in the engine-agnostic commons C++ layer. Closes **RUN-80**, **RUN-81**, **RUN-83**.

Every behavior change is **on-device validated** (Samsung S25, Hexagon v79, qwen3-0.6b through QHexRT, greedy) in addition to the commons host suites.

## RUN-80 — system prompt causing unnecessary tool calls

**Root cause (device-proven).** Given a tool set but no decision framing, a small model (qwen3-0.6b) emits a tool call for *every* prompt — even "what is the capital of France?" (→ `get_weather(France)`). Neutralizing the aggressive prompt improves quality but does **not** stop the over-calling on its own. The lever is a **tool-decision instruction on the system role**.

Three coordinated parts stop over-calling:

1. **Tool-decision system hint — the actual over-call fix.** On the AUTO decision path, commons appends one benign line to the effective system prompt — *"call a tool only when the user's request genuinely requires it … otherwise reply directly in plain text and do not call any tool"* — **preserving any caller-set system prompt** (device-verified that a caller system prompt alone does not stop over-calling, but caller-prompt + this hint does). Universal (every engine); withheld on REQUIRED/SPECIFIC (a call is forced), NONE (tools vetoed), and pure synthesis turns.

2. **Neutralized the injected prompt** (`tool_calling.cpp`): removed the hardcoded `weather=get_weather / math=calculate / time=get_current_time` RULES, the *"DO NOT compute answers yourself. ALWAYS use the tool"* bias, and the few-shot examples (which also poisoned tool args — the old prompt device-produced a nonsensical `get_current_time(time="Paris")`). Kept the tag/shape guidance the parsers depend on. A strict improvement on every engine.

3. **`tool_choice` gating**: NONE never re-advertises tools; REQUIRED appends one firm directive on the initial turn only; AUTO stays permissive.

### QHexRT grammar now actually engages (bare-Pythonic `[name(args)]`)

The QHexRT `toolcall_opt` grammar constrains output to *free chat OR one well-formed `[name(args)]` call* — but it only engages if the model's first token is `[`. Commons previously prompted the DEFAULT `<tool_call>{json}` convention, so the model emitted a `<`-led tag and the grammar never fired (inert). This PR closes that:

- A new **C-only `RAC_TOOL_FORMAT_PYTHONIC`** flavor renders the tool prompt as bare `[name(args)]` (no tags) and parses it back (a shared `parse_pythonic_call`, reused with the LFM2 Pythonic parser; `detect_format` gains a **strict** last-place `[ident(...)]` predicate that will not false-positive on prose / markdown links / citations / lists).
- It is selected **only for grammar backends** (framework == QHEXRT, probed once per request via `lifecycle_llm_supports_grammar()`), via an internal `rac_tool_call_format_prompt_grammar_proto` build entry + clearing the parse format hint so the parser auto-detects the bare call. Every other engine keeps DEFAULT/LFM2, byte-for-byte.
- **No IDL/proto change** — `PYTHONIC` is a commons-internal C enum with no proto `ToolCallFormatName` value, so the IDL drift gate and the C++/Swift/Kotlin/Dart/TS regen are untouched.
- Device-proven the grammar engages: the exact commons bare-Pythonic prompt + `toolcall_opt` grammar yields clean `[get_weather(location="Tokyo")]` / `[calculate(expression="45 * 12")]` on relevant asks and plain-text abstention on general-knowledge asks.
- **Finding folded in:** on qwen3-0.6b, `<tool_name>`-style *angle-bracket* placeholders in the example prime the model to emit a markdown-ish `[calculate](args)` that the bare parser rejects; **concrete** `[tool_name(param="value")]` placeholders yield clean `[name(args)]` for every tool. The example uses the concrete form (device-verified).

## RUN-81 — LLM thinking-mode behavior

- The `/no_think` prompt directive is now an **allowlist** (`llm_thinking_directive_internal.h`): only QHexRT (which suppresses thinking in-runtime and strips a commons-injected `/no_think`) skips it. Every other/unknown engine keeps injecting.
- Threaded `framework_name` at the proto generate/stream + tool-loop generate sites; the legacy generic C-API path keeps injecting (documented). The universal post-hoc `<think>` strip and per-model thinking tags are untouched.

## RUN-83 — tool-calling templates for top model families

Documented + test-locked the family → format mapping — **LFM2.5 → LFM2** (`[name(args)]` tags); **Qwen3 / Qwen3.5 / Qwen3-VL (v81 1.7B/4B/8B/27B) and 1-bit/ternary Bonsai → DEFAULT** (`<tool_call>{json}`). On QHexRT the grammar path is family-independent (bare-Pythonic), so the flavor only steers the text parser for non-grammar engines. The per-model manifest `format` key wiring is a forge-repo follow-up.

## Multi-engine safety

`tool_calling.cpp` / `rac_llm_thinking.cpp` are shared by all engines, so every engine-specific behavior is gated:
- **Tool-decision hint** → universal (a benign abstention instruction; never forbids a needed call), only on the tool-calling AUTO path.
- **Bare-Pythonic build/parse** → gated on `grammar_backend` (QHexRT-only capability probe). Non-grammar engines keep DEFAULT/LFM2 build + the explicit-format parse, unchanged.
- **Grammar attach** → gated on `supports_grammar`. **`/no_think` skip** → allowlist on `framework_name`.

No IDL/proto regen. No extern-C ABI break (additive struct field; the public `rac_tool_call_format_prompt_proto` is preserved and delegates to a shared impl).

## On-device validation (S25 / Hexagon v79 / qwen3-0.6b via QHexRT, greedy)

| system / path | question | result |
| --- | --- | --- |
| no decision hint (baseline) | capital of France | `get_weather(France)` — over-call |
| **+ tool-decision hint** | capital of France / who wrote Hamlet | plain answer, **no call** |
| **+ tool-decision hint** | weather in Tokyo / 45×12 | `get_weather` / `calculate` — correct |
| **grammar-engage (bare-Pythonic + grammar)** | weather in Tokyo | `[get_weather(location="Tokyo")]` — grammar engages |
| **grammar-engage** | 45 × 12 | `[calculate(expression="45 * 12")]` — grammar engages |
| **grammar-engage** | capital of France / who wrote Hamlet | plain-text abstention (free-chat path) |
| `tool_choice=NONE` | capital of France | normal answer (tools suppressed) |

## Testing

All commons host suites pass (Linux/x86, and the same `linux-debug` preset in CI): `tool_calling_tests`, `tool_calling_run_loop_tests`, `tool_calling_session_proto_tests`, `llm_thinking_tests`.

New coverage: **tool-decision hint gating** (run-loop + session; present on AUTO, caller prompt preserved, withheld on NONE/REQUIRED/synthesis); **bare-Pythonic** detect/parse/render + strict-predicate false-positive guards + multibracket/`]`-in-string/empty-arg/trailing-text edge cases (added after an adversarial review caught non-atomic malloc + a `strrchr` bracket-matching bug — both fixed); **grammar-backend build+parse end-to-end** (run-loop) and **session keep-tools bare-Pythonic** persistence; non-grammar engines asserted unchanged. Prior RUN-81/83 coverage retained.

## Review

A 16-agent adversarial review (each finding independently verified) of the grammar-engage diff caught and this PR fixes: `parse_pythonic_call` non-atomic allocation (could return success with a null args pointer / leak on OOM), `parse_pythonic_format` `strrchr` matching the *last* `]` in the whole output instead of the matching one (`[a()] text [b]` dropped the call), and an unguarded clean_text malloc — all fixed with regression tests. An earlier review round fixed a `tool_choice=NONE`-still-attaches-grammar bug. Refuted findings (a claimed LFM2 leak, a lifecycle race, an "empty-string arg" bug) were verified not-real.

## Summary by CodeRabbit

* **New Features**
  * Added grammar-capability detection for loaded language models.
  * Improved tool calling on compatible engines with constrained decoding and bare Python-style tool-call syntax.
  * Added clearer handling for required, optional, and disabled tool choices.
  * Preserved caller system prompts while improving tool-decision guidance.

* **Bug Fixes**
  * Prevented duplicate thinking-disabling directives on engines that handle them natively.
  * Added validation for required tool calls when no tools are available.

* **Tests**
  * Expanded coverage for grammar-constrained tool calling, prompt behavior, thinking directives, and follow-up generations.


AmanSwar:
Addressed the CodeRabbit review (all host suites still green — `tool_calling_tests`, `tool_calling_run_loop_tests`, `tool_calling_session_proto_tests`, `llm_thinking_tests`):

1. **Session NONE / synthesis-turn grammar coverage** — added. `test_session_grammar_gated_on_framework` now covers `tool_choice=NONE` (no grammar attached) and a tool-call→synthesis sequence (decision turn constrained, synthesis turn unconstrained) on the session path, mirroring the run-loop suite.
2. **Duplicated grammar-clear predicate** — extracted into a shared helper `tool_grammar_constrained_this_turn(...)` (`tool_calling_generation_internal.h`); both the run loop and the session now derive the tool_names-clearing decision from it, so the two parallel implementations stay in lockstep.
3. **Legacy C-API `/no_think` framework gap** (`rac_llm_service.cpp`) — kept as a documented `TODO(RUN-81)`, deliberately out of scope. Threading framework identity through the generic C-API service handle is an ABI-shaped change beyond this epic; the proto/SDK hot paths (`llm_module.cpp`, tool loop) **are** framework-gated, and on this path QHexRT harmlessly strips the redundant token while llama.cpp/onnx still need it — so there's no correctness gap, only a minor redundancy on a non-hot path.

AmanSwar:
Addressed the CodeRabbit review + ran a full production-readiness pass (all commons host suites green — the entire 96-test suite passes, incl. `tool_calling_run_loop_tests`, `tool_calling_tests`, `tool_calling_session_proto_tests`, `llm_thinking_tests`):

**CodeRabbit findings resolved**
1. **SPECIFIC grammar-spec mismatch (Major, run-loop + session).** `tool_choice=SPECIFIC` now narrows the QHexRT grammar spec to the single forced tool (`toolcall_opt:<forced>`) instead of listing every tool — so the grammar *structurally* prevents a wrong-tool call rather than relying on post-hoc policy rejection. This became active once the grammar actually engages. Fixed in the run-loop and session paths, with a test asserting the narrowed spec.
2. **REQUIRED with zero tools (Minor).** Now rejected up front (`INVALID_ARGUMENT`, no generation) in `apply_explicit_tool_choice` (both paths) instead of emitting a "you must call a tool" prompt that fails downstream. Test added.

**Production-readiness review (self-run, 27-agent adversarial + independent verification of every finding).** Verdict: production-ready. The critical invariant — non-QHexRT engines (llama.cpp/onnx/cloud) byte-for-byte unchanged — is verified maintained via the `grammar_backend`/`supports_grammar` gates; no regressions; the RUN-80 hint + grammar-engage + RUN-81 thinking features compose correctly. No new bugs surfaced. Closed the flagged coverage gaps with new tests: SPECIFIC-narrows-grammar, REQUIRED-no-tools, REQUIRED+grammar composition, unknown bare-call validation, and clarified the legacy no-think overload comment.

PR #599, opened 2026-07-27
Centralize SDK defaults in idl/sdk_defaults.proto and fix the drift they had accumulated

Siddhesh2377:
Default values were written out by hand in every SDK, and they had drifted. Ask
the five platform SDKs what the VLM `max_tokens` default is and you got three
different answers, none of which matched the C layer that actually runs
inference. This moves every default into `idl/sdk_defaults.proto` and has the
SDKs read generated constants instead.

Audit that motivated it: `thoughts/shared/research/hardcoded_values_map_2026-07-27.md`.
Design and implementation record: `thoughts/shared/plans/central_default_pool.md`.

## What was actually broken

| Layer | VLM max_tokens | VLM top_k |
|---|---|---|
| `RAC_VLM_OPTIONS_DEFAULT` (C) | 2048 | 0 |
| Swift | 128 | 40 |
| Kotlin / Flutter / Web | 256 | 40 |
| `MLX.swift` no-options fallback | 1024 | 40 |

A Swift caption was capped 16x shorter than the C default. Several of those
copies carried comments naming the file they were supposed to match, which
documents a duplicate rather than preventing one.

Three more of the same shape:

- The three copies of `OkHttpHttpTransport.kt` disagreed on stream chunk size by
  8x (256 KB vs 32 KB) and on read timeout (60s vs 120s).
- Flutter, React Native, and Web each used a VAD floor multiplier of 2.2 against
  the C layer's 2.0. `energy_vad.cpp:230` and Web's `VoiceAgentMicDriver.ts:155`
  compute the same `floor * multiplier` product, so that was a real behavioral
  split.
- Web ran its own voice-agent sampling (200 tokens, temperature 0.7) while the
  other four inherited 96 and greedy decoding from the C++ orchestrator.

And two that were pure transcription: Flutter hand-copied 125 error constants
from `errors.proto` (13 of which anything referenced), and Python hand-copied
130 more under a docstring asking the reader to keep them in sync.

## How it works

`rac_default` annotations plus the convenience post-processors already existed
and covered 11 messages. They did not cover LLM options, VLM options, or
anything without a proto message at all: HTTP timeouts, mic sample rates, worker
handshakes, retry policy. So this extends that mechanism rather than adding a
second one.

`idl/sdk_defaults.proto` declares eight messages carrying 26 annotations:
`Network`, `AudioCapture`, `VoiceAgent`, `Hybrid`, `Worker`, `FFI`,
`Environment`, `Storage`. Platform-specific values still live there, in a message
named for the owner. `WorkerDefaults` is Web-only because nothing else has a
worker handshake; being platform-specific is a reason to name the owner, not to
hide the number in platform source.

**The pool generates no message types.** Nothing puts a `NetworkDefaults` on a
wire, so six message classes in five languages to carry 26 scalars would be
bloat, and every call site would read
`RANetworkDefaults.defaults().requestTimeoutMs` where a constant is clearer. The
file is listed in a new `DECLARATION_ONLY_FILES` set and excluded from all five
message generators; a new generator emits plain constants instead.

Four new generators:

| Generator | Emits | Why |
|---|---|---|
| `generate_cpp_defaults.py` | `rac_defaults_generated.h`, 75 macros | commons runs inference, so a default the C layer disagrees with never takes effect |
| `generate_defaults_pool.py` | Swift/Kotlin/Dart/TS constants, 6 files | the pool, per language |
| `generate_dart_result_codes.py` | `ra_result_codes.dart`, 130 codes | Dart needs `const` for `switch` and `exceptionalReturn:` |
| `generate_python_errors.py` | `_generated_errors.py`, 130 + 9 | Python has no convenience post-processor |

One generator serves four languages instead of four generators, because a flat
constant table has almost no per-language idiom. The argument in
`CONVENIENCE_CODEGEN_DESIGN.md` §0.3 for splitting applies to
`defaults()` / `validate()` / enum accessors, not to this.

Two constraints worth knowing, both found by breaking a build:

- The TypeScript pool uses `Object.freeze`, not `as const`. With `as const` every
  value takes a literal type, so `const x = pool.speechRmsThreshold` infers
  `0.015` and any later assignment fails to typecheck.
- The Dart result codes are generated rather than derived at runtime. A
  `ProtobufEnum`'s `.value` is a field read, so it cannot be a `switch` case or a
  `Pointer.fromFunction(exceptionalReturn:)` argument, and enum names can be
  stripped with `--define=protobuf.omit_enum_names=true`, so the messages are
  built at codegen time.

The Flutter and React Native Android plugins each get their own copy of the
Kotlin pool, because neither depends on the Kotlin SDK and `RADefaults` is not
otherwise on their classpath.

## Behavior changes

Three, all intentional:

1. **VLM `max_tokens` is 2048 everywhere.** Captions on Kotlin, Flutter, and Web
   run 8x longer before truncating; Swift 16x.
2. **VLM `top_k` is 0 (disabled) everywhere.** All four SDKs had set 40.
3. **Web voice replies are 96 tokens and greedy**, matching the four platforms
   that go through the C++ orchestrator.

The VAD floor multiplier is now 2.0 on Flutter, RN, and Web, so speech onset
triggers slightly more readily on those three and matches Kotlin. Web's second
VAD energy threshold was 0.005 where its own mic driver said 0.015; both now read
0.015.

LLM sampling is unchanged (100 / 0.8 / 1.0 was already the C value), and Android
voice behavior is unchanged.

## Also in here

`MLX.swift` picked STT loader classes by substring-matching a blob that mixed the
config's `architectures` with the model id and the containing directory name, so
a folder called `whisper-something` could decide which model class loaded a set
of weights. It now matches the config's architecture fields only and fails
instead of guessing from a name.

`examples/flutter/RunAnywhereAI/android/build.gradle` gains a five-line
`gradle.beforeProject` hook. `jni 1.0.1` skips `apply plugin: 'kotlin-android'`
when the AGP major is 9 or above but still configures the KGP-owned top-level
`kotlin { }` extension, so the project fails to evaluate with "Could not find
method kotlin()". It arrives transitively via `path_provider_android` and
`flutter_secure_storage_android` and 1.0.1 is the newest release, so there is no
version of it that works here. Downgrading to 0.15.2 evaluates fine and then
breaks `path_provider_android`, which needs the jni 1.x Dart API, so the fix
applies the plugin the package assumed it would have. This is what unblocked the
Flutter Android build; it fails the same way on `main`.

## Keeping it fixed

`check_no_hardcoded_defaults.sh` runs in the `centralization` job as a bash-only
check. Verified in both directions: **42 violations against the pre-change tree,
0 after.** It catches every category from the audit, including the three OkHttp
chunk sizes side by side. A `not-a-default: <reason>` marker opts a line out and
requires a reason so silencing it stays reviewable; one use so far, on RIFF
header scratch values in `audio.py` that are overwritten from the fmt chunk.

`verify_default_pool.sh` answers a different question: it reads the pool back out
of all seven emitted targets and diffs them against the proto, **26 defaults x 7
languages = 182 value checks**. The gate proves nobody re-declared a default;
this proves the values the SDKs read are identical everywhere. Codegen that runs
for four languages and silently skips the fifth passes the gate and fails this. I
negative-tested it by injecting the original 2.2 drift into the Dart pool, which
it caught.

## Verification

| Check | Result |
|---|---|
| 8 validation gates | all pass |
| Default pool consistency, 182 checks | pass |
| Kotlin compile / unit tests / ktlint | pass |
| Flutter analyze, all 5 packages | no issues |
| Flutter unit tests | 68 passed |
| Web tsc core+llamacpp+onnx, vitest | pass, 232 passed |
| RN tsc all 4 packages, jest | pass, 50 passed |
| Python pytest | 374 passed, 21 skipped |
| commons, all 176 objects | 0 errors |
| Codegen determinism, sha256 | 7 outputs byte-identical on re-run |
| Android commons, 3 ABIs | built with NDK 27.3.13750724 |
| Kotlin / Flutter / RN example APKs | all built and installed on an arm64 API 36 device |

Not verified, stated rather than implied:

- **`ctest` did not run.** The full commons build fails in vendored `libarchive`
  under gcc's `-Werror=discarded-qualifiers`, before reaching any first-party
  file. The 176 objects were built individually to compensate.
- **Swift does not compile on the Linux box this was written on.** The Swift
  changes are covered by review and by a `-Werror` C probe asserting every macro
  resolves to the intended value. No Swift proto output is stale, since
  `sdk_defaults.proto` deliberately generates none.

Building the example apps caught a regression the per-package analyzer missed:
four sibling Flutter packages also used `RacResultCode`, and `dart analyze`
scoped to `packages/runanywhere` never saw them. All five are checked now.

## Left for later

- **The three forked `OkHttpHttpTransport.kt` copies are still forked** (748 /
  647 / 607 lines). Their constants come from the pool and the gate holds them
  together; de-forking is build-system work across two plugin toolchains.
- **Python's `catalog.py` stays local.** Documented in the module docstring as a
  deliberate choice for a package with no control-plane dependency, with the
  reasoning and what would change it.
- `RAC_VAD_DEFAULT_FRAME_LENGTH` stays a C literal: it is in seconds where the
  proto's `frame_length_ms` is milliseconds, so aliasing it would be wrong.

## Summary by CodeRabbit

* **New Features**
  * Added a centralized defaults pool (code-generated) to drive networking, audio, voice, generation, storage, and worker configuration across platforms.
  * Added generated Python `ErrorCode`/`ErrorCategory` enums.
  * Added a public helper to retrieve the default voice-agent VAD model id.
* **Bug Fixes**
  * Improved WAV decoding by requiring the `fmt` chunk to precede `data` and validating `fmt` chunk length.
  * Propagated VAD calibration defaults/range validation consistently.
* **Chores**
  * Added CI validation to block hardcoded “default-like” literals and to verify generated defaults match the source across languages.


shubhammalhotra28:
Reviewed this PR (all SDK scopes + codegen infra) and pushed one fix for the failing `swift-spm` check.

**Fix pushed (1731c53be):** `MLX.swift:358` references `RAC_VLM_OPTIONS_DEFAULT`, which was a C macro expanding to a struct compound literal — Swift's Clang importer can't import those ("structure not supported"), so MLXRuntime failed to compile. Converted it to a `static const rac_vlm_options_t` in `rac_vlm_types.h`, mirroring how `RAC_LLM_OPTIONS_DEFAULT` already handles this. Values unchanged; verified it compiles as C, C++17, C++20, and that `swift build` now completes with the PR's headers.

**Non-blocking findings worth a look:**
- `verify_default_pool.sh` (the headline cross-language drift check) is never wired into CI — only the hardcoded-defaults scanner is. It runs in <1s; worth adding to the `centralization` job.
- MLX STT loader dispatch now relies solely on `config.json` `model_type`/`architectures`; Whisper MLX conversions shipping OpenAI dims-style configs (no `model_type`) and some Parakeet conversions would now throw `unsupportedSTTModel` where name-based hints previously worked. The `nemo.collections.asr.models` predicate is dead code since the `target` key is never decoded.
- Swift `RALLMGenerationOptions` bare-init defaults drop maxTokens 512→100 (plus temp/topP/topK changes) — intentional per the pool, but a 5x cut on public API worth calling out in the description.
- Generated VLM `defaults()` also flips `streaming_enabled` and `use_gpu` to true (previously unset/false in the hand-written tables).
- Generated `VADConfiguration.validate()` rejects a default-constructed proto (`calibration_multiplier` 0 fails the 1.5–4.0 range with no unset carve-out). Latent — no in-repo callers — but a trap for external users.
- RN/Flutter OkHttp read timeout goes 120s→60s via the pool (native Kotlin was already 60s).
- Pool verifier's `norm_val` still doesn't normalize `RAC_TRUE`→`true`; the first bool field added to the pool will spuriously fail the check.

shubhammalhotra28:
Can be merged with improvements as per the above comment in the follow up PR!

PR #587, opened 2026-07-23
electron: HF page-URL fix + reasoning mode + RAG

AmanSwar:
@
## Summary
Three things on the Electron SDK, built on the existing branch:

1. **HuggingFace page-URL resolution** — pasting `https://huggingface.co/owner/repo` (a web page, not a file) used to download the repo HTML as the "model" and fail to load. `parseHfUrl` now routes HF page/tree/blob URLs to repo resolution (picks the GGUF).
2. **Reasoning (thinking) mode** — `splitThinking`/`stripThinking`/`isThinking` split a model’s `<think>…</think>` from its answer (mirrors commons `rac_llm_thinking`, which the C ABI marks SDK-internal). Exposed on the bridge; demo adds a reasoning toggle + a collapsible reasoning block in chat.
3. **RAG** — `RagSession` over the proto-byte `rac_rag_*` C ABI: create a session on an embedding model (+ optional LLM), ingest documents, ask grounded questions.

## RAG details
- Bound `rac_rag_*_proto` + a `registerModel` binding in `addon.cpp`. Key gap: RAG resolves model **ids** via commons’ global registry, which the path-loading Electron SDK never populated — `registerModel` bridges that.
- Vendored the generated proto codec (one small pure dep, `@bufbuild/protobuf`); a `file:` workspace dep broke on symlink realpath resolution.
- Object-in/object-out `rag*` bridge methods; generic dispatch needed no host changes (Buffers cross the MessagePort as `Uint8Array`, so the addon accepts typed arrays).
- Demo: a **Knowledge** tab (add docs → grounded Q&A with source chunks).

## Testing
- **365/365 unit tests** (RagSession + vendored-codec roundtrip + thinking split + updated surface tests).
- **8/8 on-device e2e** (`e2e_rag.js`) against the running app: grounded answers ("Every 90 days", "the Helios vault located in Reykjavik") with retrieved sources, direct-SDK and UI paths.

Note: the prebuilt `.node` is gitignored (built by the release pipeline); `addon.cpp` is the source of truth.
@

## Summary by CodeRabbit

- **New Features**
  - Added Knowledge (RAG) tools for importing documents, asking questions, viewing sources, and managing indexed content.
  - Added optional Reasoning mode with collapsible model thought displays.
  - Added public Electron APIs for RAG sessions and reasoning-content handling.
  - Hugging Face model pages and file links are now supported when adding models.

- **Bug Fixes**
  - Improved model switching and recovery from stale or unloaded models.
[...]

PR #601, opened 2026-07-28
electron: make the desktop app the demo, and harden it for real use

AmanSwar:
@
Continues the Electron work from #587 (closed, folded elsewhere). Two of that
branchs SDK modules never reached `main` — `speech.ts` and `chat-template.ts`
are in this PR.

## 1. One demo app, not two

`apps/desktop` and `examples/electron/RunAnywhereAI` had diverged badly (1092 vs
679 lines of renderer). Every recent improvement lived only in the former, while
the SDK README pointed at the latter. The productised app now **is** the demo, at
the canonical path; the stale copy is gone.

## 2. Bugs found by using it

| Symptom | Cause |
|---|---|
| Chat "forgot" the previous turns | We sent a hand-rolled `User:/Assistant:` transcript. llama.cpp did not recognise it and wrapped **the whole conversation as one user message**, so every turn looked like the first. Fixed with `formatChat()` — ChatML / Llama-3 / Gemma / Mistral, tagged per catalog entry. |
| Replies were empty or all reasoning | With a correct prompt the model reasons properly and spent the entire 256-token budget doing it. Measured: thinking + 256 tokens → empty; empty `<think>` prefilled → correct in 586 ms. Reasoning-off now genuinely suppresses it; budget raised to 512. |
| Replies attributed the users facts to the model | 0.8B is too small for multi-fact attribution ("I am 21 years old"). Measured 1/3 vs **3/3** for 2B on a recall probe → default model 0.8B → 2B, a system prompt that establishes who is who, temperature 0.7 → 0.3. |
| TTS read markdown aloud ("asterisk asterisk") | New `speakableText()` strips markdown and speaks symbols (`20°C` → "20 degrees Celsius"). |
| The mic button silently stopped responding | `cc.stop()` sat outside `finishTurn`s `try`, so a teardown throw left `busy = true` forever. Whole turn is now guarded, teardown cannot throw, and the orb self-heals from any unexpected state. |
| Models piled up until memory ran out | Slots are per-modality and nothing evicted anything: chats LLM, then a VLM on top, then STT + TTS. A unit-tested policy now releases whatever the current screen does not need — while keeping the three Voice genuinely needs. Measured: the chat→vision step went 2285 MB → **1101 MB**. |
| Tools tab did nothing | A stale memoised LLM handle after eviction; only chat had recovery. Centralised in `withLlm()`. Tools also **execute** now instead of only parsing a call. |

## 3. Production hardening

- **Model downloads are verified** against the SHA-256 the HF origin API reports, while the bytes come from a CDN — catches corruption and a poisoned object. A mismatch deletes the file *and* the `.part`.
- **Disk-space guard** before a multi-GB download, instead of dying on ENOSPC.
- **Atomic store writes** (temp + fsync + rename). Previously a crash mid-write truncated `conversations.json` and the app silently started empty. A corrupt store is now backed up, not discarded.
- **Crash handlers** (`uncaughtException`, `unhandledRejection`, `render-process-gone` → reload), a `will-navigate` guard, and the **self-test IPC is gated behind `RA_SELFTEST`** — `runanywhere-test-done` reached `app.exit()` in shipped builds.
- **Settings migration** so a persisted copy of a superseded default is upgraded, without ever touching a prompt the user customised.
- **Model licences surfaced** — Gemma / Llama / NVIDIA weights carry use restrictions, shown with a link before download.
- The "100% on-device" claim is now accurate ("Inference runs on this device") — the app does fetch models over the network.

## 4. Catalog refreshed to 2026

SmolLM and Qwen2.5 are gone. Every URL was HTTP-verified (200 + GGUF magic) and both new defaults were **downloaded, loaded and generated with** on device: Qwen3.5 (0.8B/2B/4B/9B), Gemma 4 E2B/E4B, LFM2.5, Ministral 3, Phi-4-mini, Nemotron 3 Nano, plus VLMs.

## 5. UI

Rebuilt on an agreed direction: a quiet shell holding a floating content panel, light **and** dark, product/workbench navigation split, and assistant replies typeset as **prose** rather than coloured bubbles. The palette follows `DESIGN_GUIDELINE.md` — the app had been using `#FF5500`, which that document explicitly bans.

## Testing

- **435 SDK + 24 app unit tests** (new: chat templates 17, download integrity 11, model policy 11, store 13, speech 12, licences 5)
- System checks on device: 8/8 app, 7/7 model release with real process memory, 5/5 voice robustness including a forced mid-turn failure
- CI now **compiles the native addon** (previously never built in CI, so a broken binding shipped undetected) and runs the apps tests

## Known, not fixed

- **Knowledge/RAG loads a second copy of the LLM** — it bypasses the shared slot inside the SDK, so the release policy cannot reach it. Needs an SDK change.
- No cancel for an in-flight generation; there is no cancel API in the stack. The voice UI lets you abandon a turn instead.
- The voice state machine is DOM-closure bound, so it is covered by a system test rather than unit tests.
@

## Summary by CodeRabbit

* **New Features**
  * Rolled out an updated RunAnywhere AI desktop UI with redesigned navigation, improved model modality retention, and a refreshed voice experience.
  * Added multi-template chat formatting, speech-friendly text output, and license metadata for downloadable models.
  * Added Windows desktop launchers and an icon generator for the example.
* **Bug Fixes**
  * Improved SDK downloads with disk-space prechecks and SHA-256 integrity verification (rejects mismatches and avoids persisting corrupted downloads).
* **Documentation**
  * Rewrote the Windows runbook with clearer GPU behavior, self-test instructions, and packaging guidance.
* **Tests**
  * Expanded hermetic CI/unit tests, including native addon build verification and added coverage for chat templates, speech output, downloads, and persistence/migrations.


AmanSwar:
Worked through every bot finding. Fixed, with the reasoning:

**CodeQL HIGH — `js/polynomial-redos` in `chat-template.ts`**
My own regression: I fixed this exact class in `thinking.ts` earlier and then reintroduced a second copy of the regex here. Rather than write a third implementation, `thinking.ts` now owns the semantics — `stripAllThinking()` is an `indexOf` scan that removes *every* block, and `chat-template.ts` calls it. 60k repeated `<think>` openers now hash in ~1ms; there is a regression test asserting it.

**`sha256File` buffered a whole model into memory** — real, and worse than it looks: catalog entries reach ~7GB, past the max Buffer length. Now streamed with an 8MB high-water mark, and `finalize` moved into the async continuation. Test hashes ~96MB while a 5ms interval ticks, asserting the loop keeps turning.

**Gemma `hasTurnMarkup`** — the bot suggested returning `false` to match the comment. That would have been the wrong half to fix: `build_prompt` in `engines/llamacpp/llamacpp_backend.cpp` really did not know `<start_of_turn>`, so a correctly formatted multi-turn Gemma prompt fell through to the single-user-message path and the model lost its history — the very bug this PR exists to fix. Taught the backend the marker instead, and added a test that asserts every template `formatChat` emits is recognised, so the two lists cannot drift.

**System-only prompt dropped (gemma/mistral)** — real; `pending` was only flushed when a user turn existed. Now emitted as the opening user turn. Tests for both.

**Late rejection deletes a newer handle** — real race; cleanup is now guarded by identity (`handles[modality] === memo`).

**`STALE_MODEL_RE` too broad** — real; `/model.*load/` matched `"model failed to load: not enough memory"`, so an OOM would immediately retry a multi-GB load. Anchored on the specific host messages.

**`activeChatTemplate()` on a cold catalog** — real, and the same failure mode as the memory bug: a message sent before the catalog resolved formatted a Llama-3/Gemma prompt as ChatML. It now awaits a shared `ensureCatalog()`; `buildPrompt` is async and all three call sites await it.

**Models tab persisted `source`, the picker persisted `id`** — real; they disagreed for custom models and the saved choice would not resolve on restart. Id is now the single identity everywhere, with `modelSource(id)` mapping to the loadable source at load time.

**Cancelled voice turn hijacks the new one** — real, and the user hit this twice in testing. `abandoned` was a single boolean that the next turn reset. Extracted `turn-guard.js`: each turn takes a number, `current()` can never become true again once superseded, and every mid-chain write plus the `finally` block is gated on it. Six unit tests, including one that replays the exact cancel-then-speak-again sequence.

**`console-message` signature** — real; Electron 43 passes a details object with a *string* level, so `level >= 2` was always false and renderer warnings/errors were never surfaced.

**`resolveNativePath` was Windows-only** — now derives the variant from `process.platform`/`process.arch` and fails fast with a readable message naming the prebuilds that are present, shown in a dialog instead of dying silently.

**Also fixed:** `persist-credentials: false` on both jobs that run `npm install`; `.voiceerror` uses the themed `--red` (the hardcoded `#ffb4b4` was near-invisible on the light panel); the orphaned VAD `Detection` label; and every stale `apps/desktop` / `run-demo-gpu.cmd` reference in the README, `main.js` header, `make-icon.js` and the settings pane.

**CI:** `native-addon-build` was on `windows-latest`, which no longer ships the VS 2022 the `windows-release` preset generator needs. Pinned to `windows-2022` and switched to the repo's `setup-toolchain` action, mirroring the working `rcli-windows` job. The `swift-spm` / `python-*` failures were not from this branch — main had already fixed them; merged `origin/main` (clean).

**Skipped:** generating `ChatTemplate` / `ChatTurn` from `idl/*.proto`. These are Electron-local prompt-formatting helpers that never cross the C ABI or reach another platform — the guideline targets structured values shared across SDKs. Happy to revisit if we ever expose templating through commons.

444 SDK unit tests and 30 demo-app unit tests pass.

examples/electron/RunAnywhereAI/renderer.js
```diff
@@ -568,33 +953,247 @@ function captureController() {
...
+    // idle, or any state we did not anticipate: self-heal and listen.
+    if (busy) { busy = false; turns.cancel(); }
+    stopPlayback();
+    beginListening();
+  });
+  orb.addEventListener('keydown', (e) => { if (e.key === ' ' || e.key === 'Enter') { e.preventDefault(); orb.click(); } });
+
+  // Leaving the tab must release the mic — otherwise the OS keeps showing the
+  // recording indicator and the capture keeps running in the background.
+  voiceCleanup = () => {
```
shubhammalhotra28, line 1189: **Voice turn continues after leaving the tab.** `voiceCleanup` stops capture/playback but does not invalidate a turn that is transcribing or generating. That turn can later update the Voice DOM and start TTS while the user is on another tab. Cancel the turn, clear the watchdog, and reset `busy` during cleanup; add a regression test for tab navigation mid-turn.

sdk/runanywhere-electron/src/thinking.ts
```diff
@@ -57,6 +57,32 @@ export function stripThinking(text: string): string {
...
+ * backtracks polynomially on input with many repeated `<think>` openers
+ * (js/polynomial-redos), and model output is untrusted input.
+ */
+export function stripAllThinking(text: string): string {
+  if (!text) return '';
+  let out = '';
+  let rest = text;
+  // Terminates: every iteration drops at least the open tag from `rest`.
+  for (;;) {
+    const open = firstOpen(rest);
```
shubhammalhotra28: **This scanner is still quadratic for repeated complete thinking blocks.** Each loop calls `firstOpen(rest)`, which searches the full remaining suffix for both tags. Input such as `"<think>x</think>".repeat(N)` therefore rescans increasingly overlapping suffixes and can stall prompt construction. Advance one cursor through the source instead, and cover the repeated closed-block case in the performance regression test.

shubhammalhotra28: This could be a follow up though if u would like ot if it's easier to fix in this itself. fine with either.

AmanSwar:
Both done in this PR — thanks, they were real and the first one was a genuinely good catch.

**`stripAllThinking` was still quadratic.** You were right, and my perf test was the reason it slipped through: it used `'<think>'.repeat(60000)` — all *unterminated* openers, which exit after a single scan (0.6ms) without ever exercising the loop. Repeated closed blocks are the case that does. Measured before the fix:

| blocks | before | after |
|---|---|---|
| 2,000 | 20 ms | 0.6 ms |
| 8,000 | 179 ms | 1.6 ms |
| 32,000 | **2,932 ms** | **4.7 ms** |
| 128,000 | — | 8.8 ms |

Two separate causes, both as you described: `firstOpen` ran one `indexOf` per tag, so every iteration rescanned the whole remainder for whichever tag was absent; and carrying a shrinking `rest` string recopied the remainder each time. It now walks one cursor over the original string and slices only the pieces it keeps. `nextOpen` searches the `<think` prefix both tags share and disambiguates, so every `indexOf` resumes from a strictly increasing offset — amortised linear. `splitThinking` uses the same scanner now, so there is one implementation.

The new test asserts the scaling ratio rather than a wall-clock number. I checked it fails against the old code: 4x the input cost **15.9x** the time (183ms -> 2909ms), so it trips both assertions. Also added coverage for near-miss prefixes (`<thinker>`, `<thinkers>`), which the shared-prefix scan has to reject.

**Voice turn continues after leaving the tab.** Confirmed — `voiceCleanup` stopped capture and playback but never invalidated a turn still transcribing or generating, so it would later write into the Voice DOM and start speaking while the user was elsewhere. Cleanup now supersedes the turn through the same `turn-guard` the cancel path uses (`turns.cancel()`), clears the watchdog and resets `busy`; every remaining guard in that turn then fails closed, so it cannot write text, synthesize, play audio, or reset the orb. Two regression tests cover it: tab-navigation mid-turn writes nothing, and a stale turn landing afterwards cannot clear the state of the next visit.

446 SDK + 32 demo-app unit tests pass.

PR #603, opened 2026-07-29
refactor(qhexrt): remove model catalog registration

### LINKED ISSUE #523: Move NPU catalog refresh bootstrap into SDK APIs
## Overview

CodeRabbit flagged that PR #522 repeats NPU catalog bootstrap logic in the example apps: each app registers all logical HNPU rows, catches per-model failures, and refreshes the model registry. That works for the current PR, but it means example apps carry SDK orchestration knowledge instead of making one thin SDK call.

Impact: Medium
Priority: P2
Effort: Medium

## Problem Statement

The current PR intentionally registers logical HNPU rows from Android, Flutter, and React Native example apps so token changes immediately re-seed private QHexRT/HNPU bundles. However, the register/count/catch/refresh sequence is duplicated across platforms:

- Android: examples/android/RunAnywhereAI/app/src/main/java/com/runanywhere/runanywhereai/data/ModelBootstrap.kt
- Flutter: examples/flutter/RunAnywhereAI/lib/core/services/model_catalog_bootstrap.dart
- React Native: examples/react-native/RunAnywhereAI/src/services/ModelCatalogBootstrap.ts

This makes the examples less thin and increases the chance that future SDK behavior changes drift across platforms.

## Current State

PR #522 applies quick stability fixes in the example apps:

- Android catches per-model failures and now guards final registry refresh failures.
- Settings token save/clear refreshes the NPU catalog after updating auth.
- Flutter and RN refresh their NPU catalogs after token changes.

The remaining structural issue is API shape: the SDK should provide a single catalog re-registration or NPU refresh helper.

## Proposed Solution

Expose a shared SDK-level entry point, for example one of:

```text
RunAnywhere.refreshNpuCatalog(models)
RunAnywhere.reregisterCatalog(models)
RunAnywhere.registerCatalog(models, refresh = true)
```

The SDK API should own:

- registering each provided catalog row,
- per-row failure handling/reporting,
- registry refresh,
- consistent result/count reporting,
- platform bindings for Kotlin, Flutter, and React Native.

Example apps should then only call the SDK API after auth changes.

## Implementation Plan

- [ ] Design a small cross-platform result type for catalog refresh counts/errors.
- [ ] Add the native/shared SDK API or platform wrapper API.
- [ ] Wire Kotlin, Flutter, and React Native SDK bindings.
- [ ] Replace example-app register/count/catch loops with one SDK call.
- [ ] Add focused tests for success, per-row failure, and registry refresh failure behavior.
- [ ] Update example app docs where needed.

## Success Criteria

- Example apps no longer implement multi-step NPU catalog bootstrap logic.
- Android, Flutter, and RN all use the same SDK-level contract.
- Token save/clear still refreshes HNPU catalog visibility.
- Per-model failures are surfaced consistently without crashing the UI.

## Related

- Review source: https://github.com/RunanywhereAI/runanywhere-sdks/pull/522#discussion_r3525486804
- Follow-up from PR #522.

shubhamsinnh:
## Summary

This PR simplifies QHexRT model registration to match the current architecture:

- Removes the engine- and SDK-owned QHexRT model catalog.
- Removes device-specific model selection and architecture-folder URL rewriting.
- Removes QHexRT catalog registration helpers from the Kotlin, Flutter, and React Native packages.
- Keeps `probeNpu()` and `isArchitectureSupported()` as optional device-support checks.
- Registers exact, caller-provided model URLs through each core SDK's existing `RunAnywhere.registerModel(...)` API.

## Why

PR #532 followed the older QHexRT flow: seed logical catalog entries, detect the device architecture, filter models, and rewrite URLs to architecture-specific folders.

Based on the feedback on #532, current QHexRT models are published as dedicated artifacts. QHexRT should therefore not maintain a model catalog, choose a model, or modify its URL. The caller selects the model and registers its exact link through the core SDK.

This keeps model registration in the shared, backend-neutral path. QHexRT remains responsible for backend registration and optional device-capability reporting.

## What changed

- Deleted the native QHexRT catalog implementation, build wiring, and catalog tests.
- Removed catalog policy and device-aware registration APIs from Kotlin, Flutter, and React Native QHexRT packages.
- Updated Android, Flutter, and React Native examples to use explicit URL registration through the core SDK.
- Removed QHexRT-specific HF authentication checks; authentication requirements now come from model metadata.
- Updated documentation and removed tests for the retired catalog APIs.

## Behavior after this PR

- QHexRT does not seed or own a model catalog.
- QHexRT does not select models or rewrite model URLs.
- Applications register dedicated QHexRT model URLs through `RunAnywhere.registerModel(...)`.
- Applications may use `probeNpu()` as an optional preflight before presenting or registering NPU models.
- Existing core download, registry, lifecycle, and compatibility behavior remains unchanged.

## Breaking changes

The QHexRT catalog lookup, model eligibility, HF-auth policy, and device-aware model-registration APIs have been removed. Callers using those APIs should move to explicit model registration through the core SDK.

## Validation

All PR checks pass, including:

- Kotlin/Android SDK and example tests
- Flutter analysis and tests
- React Native typecheck, lint, and tests
- Native Android, iOS, macOS, Linux, Windows, Web, and WASM builds
- Repository validation, security, and centralization checks

## Related

- Fixes #523
- Supersedes the catalog-based approach proposed in #532

## Summary by CodeRabbit

* **New Features**
  * Model registration now supports richer metadata, including download size, context length, source, and description.
  * Android, Flutter, and React Native examples can register explicit QHexRT model URLs through the core model registry.
  * QHexRT availability and architecture support can be checked before registration.

* **Breaking Changes**
  * Removed QHexRT device-specific model registration and model catalog helper APIs.
  * QHexRT now focuses on NPU probing, architecture support, and backend lifecycle management.


shubhamsinnh:
Hi @shubhammalhotra28 

This PR follows the architectural direction from your feedback on #532: -
Based on your feedback on #532, QHexRT no longer maintains or automatically adds an NPU model catalog. Users now register the exact model link directly, while the NPU check is only used to confirm device support.
The Android, Flutter, and React Native examples have been updated, and all CI checks are passing. 

Please let me know if this now aligns with the intended approach. Thanks!

Siddhesh2377:
@shubhamsinnh conflicting with main across 8 files. The two qhexrt catalog ones are modify/delete, so those need a call rather than a rebase.

shubhamsinnh:
Hi @Siddhesh2377 and @shubhammalhotra28,

The conflicts with `main` and the remaining review feedback have been addressed. All current review threads are resolved, the latest CI run is green, and the PR is mergeable.

Please take another look when convenient, thanks!

sanchitmonga22:
Thank you for taking this on. The direction is right, it is what was asked for in #532, and the refactor itself is clean. I am closing it, but I want to be precise about why, because I checked the obvious objection and it turned out to be wrong.

**First, the thing I expected to block this does not block it.** I assumed `pin_hf_ref_to_arch` was load-bearing, so deleting it would break every HNPU URL. That is not true. Commons already does the same rewrite: `maybe_resolve_logical_bundle_ref()` in `register_model_from_url.cpp:180` calls `qhexrt_resolve_bundle_variant()` from `qhexrt_bundle_policy.h:70`, which probes the device and inserts `v75`/`v79`/`v81`. Your PR does not touch either file, and Kotlin, Flutter and RN all pass `framework = QHEXRT`, so the hook still fires. You were right that the catalog copy is redundant.

**What actually blocks it is the eligibility mask that happened to live in the same file.** I probed all 49 HNPU manifest URLs against HuggingFace, four paths each:

| device arch | catalog rows with no published manifest |
|---|---|
| v79 (SM8750 / S25, our default target) | 26 of 49 |
| v75 | 23 of 49 |
| v81 | 5 of 49 |

`kModelPolicies` skips those rows before any network call. Flutter and RN still register the full list after this PR, so on a v79 device about half the catalog would try to register and fail with per-row errors. My probe matched the mask almost exactly, which tells me the mask encodes the real published tree rather than stale guesses.

Three smaller ones:

- `ModelCatalog.npuCatalog` ends up with zero production consumers. `seedCatalog()` registers only `ModelCatalog.models`, which excludes it, so the NPU picker is empty on a fresh install. Flutter and RN kept their registration in this same PR, so I read this as an oversight rather than intent. The two tests that would have caught it are deleted in the same diff, which is why CI stayed green. Green CI is not evidence here.
- `requires_hf_auth` goes away and `requiresHfAuth()` falls back to tag sniffing. Gated repos do exist in the shipping catalog: `lfm2_5_1_2b_thinking_HNPU` returns 401 on every arch.
- `context_length` was propagated only through the deleted `toQHexRTRegistrationRequest()`. `ModelRegistration.url(...)` has no such parameter. `ModelCatalog.kt` documents what happens without it: generation overruns the KV ring and decode fails with rc=-130. Registering a model that then fails at inference is worse than not showing it.

**One repo change you could not have known about.** `examples/android/RunAnywhereAI/` was extracted to github.com/RunanywhereAI/runanywhere-android earlier today and now consumes the SDK from Maven Central. Eleven of your 48 files target a tree that no longer exists here.

**If you want to land this, here is the shape I would merge.** Split it: keep the catalog removal and the Flutter/RN changes in this repo, but first decide where per-model arch eligibility lives. My preference is having `resolve_variant` report "no manifest for this arch" and having the caller skip the row, which kills the duplicate table for good instead of moving it. That is a better version of your idea and I would review it quickly. The Android part becomes a separate PR on runanywhere-android once the core change lands.

Your other observation in this PR was correct and worth keeping: the catalog should not own product data. That is still the plan.

PR #604, opened 2026-07-29
feat(connect): add macOS host support for mobile clients (remediated)

shubhammalhotra28:
## Description
Supersedes / remediates [#556](https://github.com/RunanywhereAI/runanywhere-sdks/pull/556) (`host-feature`) onto current `dev`.

Adds macOS Connect hosting with Android / iOS / Flutter / React Native clients, then hardens the review blockers:
- request-ID cancellation (Stop cancels one generation without disconnect/rediscovery)
- commons session-cap + failed transport-init safety
- platform transport/concurrency fixes (Kotlin / Flutter / RN / Swift)
- example-app Stop/LoRA/banner/hosted-state corrections
- Wire regen with trailing-whitespace strip + focused cancel/framing tests

Original Connect feature commits are preserved; `origin/dev` was merged (not rebased).

## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Documentation update
- [x] Refactoring

## Testing
- [x] Lint passes locally *(focused Connect surfaces; full matrix via CI)*
- [x] Added/updated tests for changes

### Platform-Specific Testing (check all that apply)
**Swift SDK / iOS Sample:**
- [ ] Tested on iPhone (Simulator or Device)
- [ ] Tested on iPad / Tablet
- [x] Tested on Mac (macOS target) *(Swift `ConnectCancelLifecycleTests` + commons C++ Connect tests)*

**Kotlin SDK / Android Sample:**
- [x] Tested on Android Phone (Emulator or Device) *(JVM unit: `ConnectCancelFrameTest`; device LAN interop still manual)*
- [ ] Tested on Android Tablet

**Flutter SDK / Flutter Sample:**
- [ ] Tested on iOS *(unit: `connect_cancel_frame_test`; device LAN interop still manual)*
- [ ] Tested on Android

**React Native SDK / React Native Sample:**
- [ ] Tested on iOS *(unit: `ConnectFraming.test.ts`; device LAN interop still manual)*
- [ ] Tested on Android

**Playground:**
- [ ] Tested on target platform
- [ ] Verified no regressions in existing Playground projects
**Web SDK / Web Sample:**
- [ ] Tested in Chrome (Desktop)
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] WASM backends load (LlamaCpp + ONNX)
- [ ] OPFS storage persistence verified (survives page refresh)
- [ ] Settings persistence verified (localStorage)

### Focused automation already green locally
[...]

PR #600, opened 2026-07-27
docs: revamp README with architecture diagram, eight SDKs, engines, and Console

adityabatra072:
## Description
This PR revamps the project README to better showcase the RunAnywhere SDK ecosystem and overall architecture.

### Changes
- Redesigned the README structure and messaging.
- Added a high-level architecture diagram.
- Expanded documentation to cover all eight SDKs.
- Added sections for inference engines, QHexRT, and RunAnywhere Console.
- Improved Quick Start examples and SDK installation guides.
- Updated feature matrix, supported models, example apps, and repository layout.
- Improved navigation, formatting, and overall readability.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [x] Documentation update
- [ ] Refactoring

## Testing
- [ ] Lint passes locally
- [ ] Added/updated tests for changes

### Platform-Specific Testing (check all that apply)
**Swift SDK / iOS Sample:**
- [ ] Tested on iPhone (Simulator or Device)
- [ ] Tested on iPad / Tablet
- [ ] Tested on Mac (macOS target)

**Kotlin SDK / Android Sample:**
- [ ] Tested on Android Phone (Emulator or Device)
- [ ] Tested on Android Tablet

**Flutter SDK / Flutter Sample:**
- [ ] Tested on iOS
- [ ] Tested on Android

**React Native SDK / React Native Sample:**
- [ ] Tested on iOS
- [ ] Tested on Android

**Playground:**
- [ ] Tested on target platform
- [ ] Verified no regressions in existing Playground projects
**Web SDK / Web Sample:**
- [ ] Tested in Chrome (Desktop)
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] WASM backends load (LlamaCpp + ONNX)
- [ ] OPFS storage persistence verified (survives page refresh)
- [ ] Settings persistence verified (localStorage)

## Labels
Please add the appropriate label(s):

**SDKs:**
- [ ] `Swift SDK` - Changes to Swift SDK (`sdk/runanywhere-swift`)
- [ ] `Kotlin SDK` - Changes to Kotlin SDK (`sdk/runanywhere-kotlin`)
- [ ] `Flutter SDK` - Changes to Flutter SDK (`sdk/runanywhere-flutter`)
- [ ] `React Native SDK` - Changes to React Native SDK (`sdk/runanywhere-react-native`)
- [ ] `Web SDK` - Changes to Web SDK (`sdk/runanywhere-web`)
[...]

PR #608, opened 2026-07-31
fix(flutter): stop deleteAllModels from reporting success when listing fails

ayaangazali:
## What

`RunAnywhereDownloads.deleteAllModels()` built its delete request from `list()`, which reports a failed storage lookup as an empty list. When the lookup failed, the request went out with **no model ids**, commons had nothing to delete, and the caller received `StorageDeleteResult(success: true)` having deleted nothing.

This changes only the destructive path to read through `getStorageInfoResult()`, which carries `success` / `error_message` (`storage_types.proto`), and to surface a failed lookup as `StorageDeleteResult(success: false)` with the underlying message.

## Why

A silent no-op is the worst outcome for a destructive call: a caller that asks to free space is told it succeeded while every model is still on disk, and there is nothing in the return value to retry on. `list()` already logs the failure, but the log is not visible to the caller and the `[]` is indistinguishable from "no models are downloaded".

`list()` keeps its tolerant empty-list contract — display callers such as the example app's storage view (`storage_view.dart`) legitimately want a list they can render. Only the delete path now requires a successful lookup before it acts.

## Scope

- One file, +17/-2. No API signature changes, no new types, no behavior change when the lookup succeeds.
- No unit test added, per CONTRIBUTING's E2E-through-the-example-apps convention.

## Testing

From `sdk/runanywhere-flutter/packages/runanywhere` (Flutter 3.44.6):
- `flutter pub get` — OK
- `flutter analyze --no-pub` — `No issues found!`

Rebased onto current `main` (`3becbae55`) and re-gated after the rebase.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when deleting all downloaded models by validating storage information first.
  * Deletion now reports a failure if storage details cannot be retrieved, rather than indicating success without removing models.

### DIFF
--- a/sdk/runanywhere-flutter/packages/runanywhere/lib/public/capabilities/runanywhere_downloads.dart
+++ b/sdk/runanywhere-flutter/packages/runanywhere/lib/public/capabilities/runanywhere_downloads.dart
@@ -281,12 +281,23 @@ class RunAnywhereDownloads {
     if (!DartBridge.isInitialized) {
       throw SDKException.notInitialized();
     }
-    final metrics = await list();
+    // Read the models through the result-bearing API rather than `list()`:
+    // `list()` reports a failed lookup as an empty list, which would become a
+    // delete request carrying no model ids. Commons no-ops on that and reports
+    // no error, so the caller is told a destructive operation succeeded when
+    // nothing was examined, let alone deleted.
+    final infoResult = await getStorageInfoResult(
+      StorageInfoRequest(includeModels: true),
+    );
+    if (infoResult.hasError()) {
+      return StorageDeleteResult(error: infoResult.error);
+    }
+
     // `delete_files` was renamed with an inverted polarity to
     // `keep_files_on_disk` (idl/storage_types.proto) — see [delete].
     return DartBridgeStorage.instance.deleteProto(
       StorageDeleteRequest(
-        modelIds: metrics.map((m) => m.modelId),
+        modelIds: infoResult.info.models.map((m) => m.modelId),
         clearRegistryPaths: true,
         unloadIfLoaded: true,
         allowPlatformDelete: true,


ayaangazali:
The `kotlin-android` failure is pre-existing breakage on `main`, not from this diff — this PR changes one Dart file (`runanywhere_downloads.dart`) and no Kotlin.

The job fails on a single ktlint violation in a file this PR does not touch:

```
sdk/runanywhere-kotlin/src/main/kotlin/com/runanywhere/sdk/foundation/connect/AndroidConnectTransport.kt:449:9
Declarations and declarations with comments should have an empty space between.
(standard:spacing-between-declarations-with-comments)
```

It was introduced by #604 (merged 2026-07-30) and is present on `main` at `3becbae5`, whose own `pr-build` run is failing for the same reason. Any branch rebased onto current `main` inherits it.

I've opened a one-line fix for it separately so this and the other affected PRs can go green — I'll link it here shortly. Not re-pushing this branch to chase an unrelated failure.

ayaangazali:
The one-line fix for that ktlint violation is #609 — merging it should turn `kotlin-android` green here and on the other affected PRs.

ayaangazali:
Both failures came from the base, and that base is now fixed: #609 merged and `main`'s own `pr-build` is green at `932f6b6f`.

This branch was still sitting on the pre-#609 `main`, so it kept inheriting the two lint violations. Rebased onto current `main` and force-pushed — no change to the diff itself (still the one Dart file, +17/-2).

Re-gated after the rebase: `flutter analyze --no-pub` on `packages/runanywhere` — `No issues found!`

ayaangazali:
The rebase did fix the lint failures — `kotlin-android` now gets all the way past ktlint into the test phase, and `swift-spm` is no longer failing. What's left is a flaky timing test, not this diff.

```
98 tests completed, 1 failed
```

That test asserts a wall-clock budget for coroutine cancellation to propagate across 5 consumers on `Dispatchers.Default`:

```kotlin
val torn = withTimeoutOrNull(250) { waitFor { unregisterCount.get() == 1 } } ?: false
...
assertTrue("cancel-to-native unregister fired in ${elapsed}ms (budget 250ms)", elapsed < 250)
```

A 250 ms deadline on a shared CI runner is inherently timing-sensitive. Supporting evidence that it is intermittent rather than broken: `kotlin-android` is green on #600 and #590 and on `main` at `932f6b6f`, and this PR changes exactly one Dart file (`runanywhere_downloads.dart`) with no Kotlin, so it cannot affect Kotlin coroutine teardown. A different timing test in the same job (`VoiceAgentStreamAdapterTest > last detach tears down the C registration`) flaked the same way on my #582 earlier.

Not pushing a code change for this. Re-triggering the workflow with an identical diff; if it flakes again, a maintainer re-run should clear it.

ayaangazali:
Gentle nudge on this one, since #611, #612 and #613 have since merged past it. This branch is rebased on current `main`, all checks are green, and the diff is still the single Dart file (+17/-2).

Recap of the bug in one line: `deleteAllModels()` built its request from `list()`, which reports a failed storage lookup as an empty list, so a failed lookup produced a delete request with no model ids and a `success: true` result that deleted nothing.

Happy to rebase again or adjust the approach if you would rather handle the failed-lookup case differently.

ayaangazali:
Rebased onto current `main` (a9b542f). This was not a no-op rebase: #605 changed the shapes this fix depends on, so the previous revision no longer compiled.

`StorageInfoResult` and `StorageDeleteResult` both dropped `success` / `error_message` in favour of `optional SDKError error`. Analyzing the old revision against current `main` gives:

```
error • The getter 'success' isn't defined for the type 'StorageInfoResult' • undefined_getter
error • The named parameter 'success' isn't defined • undefined_named_parameter
error • The named parameter 'errorMessage' isn't defined • undefined_named_parameter
```

The fix now propagates the lookup's own error instead of synthesising one:

```dart
final infoResult = await getStorageInfoResult(
  StorageInfoRequest(includeModels: true),
);
if (infoResult.hasError()) {
  return StorageDeleteResult(error: infoResult.error);
}
```

The defect itself is unchanged and still reproduces at `main`: `deleteAllModels()` builds its request from `list()`, which reports a failed lookup as `[]`, so the request carries no model ids, commons no-ops, and the caller gets a result with no error after a destructive call that deleted nothing.

`flutter analyze --no-pub` on `packages/runanywhere` is clean, and all 26 checks were green on the previous revision. One file, +13/-2. No rush from my side, just keeping it current and compiling.

Siddhesh2377:
Thanks @ayaangazali, solid catch!

PR #611, opened 2026-08-01
fix(web): keep the modality default module when a lower-priority slot survives

ayaangazali:
## What

`registerModuleCapabilities` and `unregisterModuleCapabilities` in `ModalityProtoAdapter.ts` each carried their own copy of the `defaultModule` fallback chain. The register copy walks 13 slots; the unregister copy walked 6, missing `embedding`, `rag`, `diffusion`, `structured-output`, `tool-calling`, `lora` and `voice-agent`.

The result: unregistering the module that happened to own the default cleared `adapterState.defaultModule` even when another module still occupied one of the seven missing slots. `ModalityProtoAdapter.tryDefault()` then returns `null` and the SDK behaves as though no adapter is installed while a perfectly usable one is still registered.

Both copies are replaced with a single `recomputeDefaultModule()` helper driven by one ordered list, so the two paths cannot drift apart again. The register-side priority order is unchanged.

## Why this is reachable, not hypothetical

`registerWasmModule` documents that a module "may be registered against any subset of capabilities", and the existing unit tests already register single-capability modules that way, including `registerWasmModule(['rag'], ragModule, ['onnx'])`. Registering an `llm` module plus a `rag` module and then tearing down the `llm` one is enough to hit it.

## Testing

From `sdk/runanywhere-web`:
- `npm run typecheck -w packages/core` — passes.
- `npx vitest run tests/unit/runtime/EmscriptenModule.test.ts` (from `packages/core`) — 9 passed.

I added one regression case next to the existing teardown test and checked it both ways: with the source fix reverted it fails with `AssertionError: expected null not to be null`, and passes with the fix applied. I would rather it earn its place than just ride along.

Local note in case it helps anyone else: running the web unit tests needed `@bufbuild/protobuf` resolvable from `sdk/shared/proto-ts`, whose `dist` requires it but which has no `node_modules` of its own. A repo-root `npm install` currently fails on an unrelated peer-dependency conflict in the React Native example, so I worked around it locally rather than changing anything committed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved default backend selection when modality-specific backends are registered or removed.
  * Preserved an available retrieval-augmented generation backend when the higher-priority language model backend is removed.
  * Ensured fallback behavior remains consistent across supported modality types.

### DIFF
--- a/sdk/runanywhere-web/packages/core/src/Adapters/ModalityProtoAdapter.ts
+++ b/sdk/runanywhere-web/packages/core/src/Adapters/ModalityProtoAdapter.ts
@@ -66,6 +66,39 @@ const MODALITY_CAPABILITIES: ReadonlySet<string> = new Set<ModalityCapabilityNam
   'voice-agent',
 ]);
 
+/**
+ * Slot preference used to pick the aggregate `defaultModule`. `llm` is the
+ * historical anchor; the rest are fallbacks so `tryDefault()` keeps returning
+ * a usable module while any modality slot is still occupied. Shared by
+ * register and unregister so the two cannot drift apart.
+ */
+const DEFAULT_MODULE_PRIORITY: readonly ModalityCapabilityName[] = [
+  'llm',
+  'vlm',
+  'stt',
+  'tts',
+  'vad',
+  'embedding',
+  'segmentation',
+  'rag',
+  'diffusion',
+  'structured-output',
+  'tool-calling',
+  'lora',
+  'voice-agent',
+];
+
+function recomputeDefaultModule(): void {
+  for (const slot of DEFAULT_MODULE_PRIORITY) {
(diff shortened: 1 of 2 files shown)


ayaangazali:
`kotlin-android` failed on the known flaky timing test, not on this change:

```
98 tests completed, 1 failed
```

That test asserts a 250 ms wall-clock budget for coroutine cancellation to propagate across five consumers on `Dispatchers.Default`, which is inherently sensitive on a shared runner.

The clearest evidence that neither diff causes it: the exact same test failed in the same way on #611, which changes only TypeScript and contains no Kotlin at all. It also failed once on #608 (a Dart-only change) and then passed on an identical re-run. `kotlin-android` is green on `main`.

Re-triggering with an unchanged diff rather than pushing a fix for it.

PR #613, opened 2026-08-01
fix(web): abort the OPFS writable stream when the createWritable path fails

ayaangazali:
## What

`writeBytesToOPFSFile` tries `createWritable` first and falls back to a sync-access handle when that path throws. If `createWritable()` itself succeeded but `write()` or `close()` threw, the stream was left open: the catch block only logged and fell through to the fallback.

## Why it matters

An open writable stream holds an **exclusive lock** on the OPFS file. So the fallback's `createSyncAccessHandle()` call, on that same handle, then fails as well and the write is lost. The leak defeats the exact recovery the fallback exists to provide, turning a recoverable `write()` failure (quota exceeded, transient storage error) into a failed persist.

The sync-access path directly below already guards its own `close()` with `try/finally`, so this is the one branch missing the release.

Fix is to abort the stream before falling through. `abort()` itself throws if `close()` already ran, so that call is guarded in turn rather than trading one unhandled rejection for another.

## Testing

From `sdk/runanywhere-web`:
- `npm run typecheck -w packages/core` — passes.
- Full unit suite from `packages/core`: **54 files, 233 tests, all passing.**
- `npx eslint` on both changed files — clean.

I added one regression test and checked it both ways: with the source fix reverted it fails with `expected "spy" to be called once, but got 0 times`, and it passes with the fix applied. It drives the real public entry point (`OPFSBridge.writeFileToOPFS`) through an injected persistent root, with a fake handle whose `write()` rejects, and asserts both that the writable is aborted and that the sync-access fallback then succeeds.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file writing reliability when browser storage limits are reached.
  * Ensured writes can fall back successfully after a streaming write fails.

### DIFF
--- a/sdk/runanywhere-web/packages/core/src/Infrastructure/OPFSBridge.ts
+++ b/sdk/runanywhere-web/packages/core/src/Infrastructure/OPFSBridge.ts
@@ -239,12 +239,25 @@ async function writeBytesToOPFSFile(
   // require a sync-access handle (i.e. no main-thread restrictions on
   // Chrome / Firefox).
   if (typeof (handle as { createWritable?: unknown }).createWritable === 'function') {
+    let writable: Awaited<ReturnType<FileSystemFileHandle['createWritable']>> | undefined;
     try {
-      const writable = await handle.createWritable({ keepExistingData: false });
+      writable = await handle.createWritable({ keepExistingData: false });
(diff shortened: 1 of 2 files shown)


shubhammalhotra28:
Thanks for the fix. @ayaangazali

PR #615, opened 2026-08-03
fix(commons): check the loaded state with the resolved model id

ayaangazali:
## What

`add_model_candidate_from_info` resolves the model id once, with a fallback for when the registry entry has no id of its own:

```cpp
const std::string id = current->id && current->id[0] != '\0' ? current->id : requested_id;
if (id.empty()) { ...return; }
```

Every later use reads that resolved `id` (the warnings, `row.model_id`, `row.storage_key`). The loaded-state probe was the one exception and passed the raw `current->id`:

```cpp
bool has_loaded_state = known_loaded_state(handle, current->id, &is_loaded);
```

## Why it matters

`known_loaded_state()` null-checks its `model_id` argument and returns false, leaving `is_loaded` untouched. So in exactly the case the fallback exists for, the probe reports no known state and `is_loaded` stays false, which means this guard never fires:

```cpp
if (is_loaded && !allow_loaded_models) {
    add_warning_once(plan, seen_warnings, "Model is loaded and cannot be safely deleted: " + id);
    return;
}
```

A model that is currently loaded then lands in the delete plan even though the caller asked not to touch loaded models. The "loaded state unavailable" warning is emitted, but it does not stop the deletion.

It is reachable through the requested-ids path (`rac_model_registry_get` then straight into this function), which does not check `current->id` first. The downloaded-models path is unaffected because it already skips entries with a null id.

Passing `id.c_str()` is safe: the early return above guarantees it is non-empty, so this only removes the null case rather than changing any behaviour that works today.

## Testing

Configured with the `macos-debug` preset and built with Ninja:
- `test_storage_analyzer_proto` — 0 test(s) failed
- `test_storage_service_proto_abi` — 18 checks, 0 failures
- `test_core --run-all` — 18 passed, 0 failed

No new test: the failure needs an `is_model_loaded` callback plus a registry entry with an empty id, which the existing analyzer fixtures do not model, and the change is a one-argument correction rather than new logic. Happy to add a fixture if you would prefer one.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model detection when registry metadata is missing an ID.
  * Prevented incorrectly treating already loaded models as new candidates.

### DIFF
--- a/sdk/runanywhere-commons/src/infrastructure/storage/storage_analyzer.cpp
+++ b/sdk/runanywhere-commons/src/infrastructure/storage/storage_analyzer.cpp
@@ -342,7 +342,12 @@ void add_model_candidate_from_info(rac_storage_analyzer_handle_t handle,
     }
 
     bool is_loaded = false;
-    bool has_loaded_state = known_loaded_state(handle, current->id, &is_loaded);
+    // Use the resolved id, not the raw one: `id` already falls back to
+    // `requested_id` when `current->id` is null or empty, and every other use
+    // below reads `id`. Passing the raw pointer makes known_loaded_state()
+    // return false for exactly that case, which leaves `is_loaded` false and
+    // silently skips the loaded-model guard.
+    bool has_loaded_state = known_loaded_state(handle, id.c_str(), &is_loaded);
     if (!has_loaded_state) {
         add_warning_once(plan, seen_warnings, kLoadedStateUnavailableWarning);
     }


sanchitmonga22:
Closing this one on the premise rather than the patch — I traced it carefully and the fallback path can't actually be reached.

`save_model_info_impl` rejects `!model->id` (`model_registry.cpp:337`) and keys the map on `model->id` (`:343`). So `rac_model_registry_get` can only ever return an entry whose id already equals the requested id — the resolved-id fallback has no case where it changes the outcome.

Separately, the layer this was meant to protect doesn't consult it: `jni_storage_is_model_loaded` (`core/src/jni/runanywhere_commons_jni.cpp:4083`) ignores `model_id` entirely and hardcodes `RAC_FALSE`, and Swift and Electron pass NULL.

One process note, meant helpfully: the PR body reports a run of `test_storage_service_proto_abi`, and that target doesn't exist in this repo. Worth double-checking before citing a test run — it's the kind of thing that costs a reviewer real time to chase.

**The bug behind this is real, though, and bigger than the patch.** `is_model_loaded` is not wired to any lifecycle store — the JNI stub above literally lies to commons. Wiring that up for real, and aligning `storage_analyzer.cpp:1209` (the site that actually *deletes files* based on loaded state), is a genuinely valuable fix. If you want it, I'll review it properly.

ayaangazali:
Thanks for tracing it rather than just closing it, and you are right on the premise. `save_model_info_impl` rejecting `!model->id` and keying on `model->id` does make the resolved-id fallback unreachable, so the patch was guarding a case that cannot occur. I was reasoning from the shape of the code rather than from reachability, and that is the actual mistake.

On the test target: you are right and I should not have cited it. `test_storage_service_proto_abi` does not exist here, and citing a run of it was careless in a way that costs a reviewer time. I have corrected my own notes so it cannot resurface, and I will only cite a gate I have actually watched pass in this repo.

Yes please on the real one. My reading matches yours: `is_model_loaded` has no source of truth behind it. The JNI callback hardcodes `RAC_FALSE` (`core/src/jni/runanywhere_commons_jni.cpp:4083`), `jni_storage_unload_model` is a no-op that returns `RAC_SUCCESS`, and Swift and Electron leave the slot NULL, so `model_is_loaded` (`storage_analyzer.cpp:164`) bails on `!handle->callbacks.is_model_loaded` and the delete path treats every model as not-loaded. That is what makes `:1235` dangerous: `unload_if_loaded` never fires, and the recursive delete runs underneath a model the engine still has open.

The shape I would propose, unless you would rather it went differently:

Commons already knows the answer and is not being asked. `model_lifecycle` keeps `g_loaded`, and `LoadedModel` carries `model_id` and `state`. So rather than asking each host to report loaded state through a callback they cannot answer correctly, have the analyzer consult the lifecycle store directly, and keep the callback only as a fallback for a host that loads outside commons. That fixes Android, Swift and Electron in one place instead of three, which also matches the layering rule in AGENTS.md.

Two things I want to check with you before writing it:

1. Whether the callback should stay in the struct at all once commons answers authoritatively, or be left in place for compatibility.
2. `unload_model` has the same hole. Consulting the lifecycle store tells us a model is loaded, but actually unloading it from inside the delete path means calling into the lifecycle unload, which is a heavier change than the read side. I would rather do the read side first so the delete path stops lying, then the unload wiring separately, unless you want them together.

Happy to go straight to a PR on the read side if that split works for you.

PR #619, opened 2026-08-04
NeuRT: Apple Neural Engine as a first-class engine (coreml -> neurt), and a CI-buildable shell

sanchitmonga22:
## Description

Integrates **NeuRT** — the Apple Neural Engine half of the sibling `neurun` repo — as a first-class engine, and makes it reachable from the SDK, the CLI and the iOS example app.

The engine formerly called `coreml` is now `neurt`, named for the runtime that implements it rather than for Apple's framework — the same convention `qhexrt` follows. **CoreML is deliberately kept** where it names Apple's framework rather than our engine: the `coreml` device runtime (`runtimes/coreml`), the `MODEL_FORMAT_COREML` / `INFERENCE_FRAMEWORK_COREML` wire values, and the `rac_diffusion_coreml_*` C ABI. The engine↔runtime name collision that made this area hard to read is gone.

### What changed

| Area | Change |
|---|---|
| `engines/coreml` → `engines/neurt` | Renamed; gained a working **LLM slot** (ANE text generation) alongside the existing diffusion slot |
| Routing | `INFERENCE_FRAMEWORK_COREML` → `RAC_ENGINE_ID_NEURT` name pin, so ANE models reach NeuRT even though its priority (100) sits below MLX's (110) |
| Bundle policy | An ANE bundle can be registered by URL; manifest selection rejects `PRECISION.json` / `*.iodesc.json` / `tokenizer.json` sidecars |
| `rcli` | The Neural Engine is actually reachable from the CLI |
| iOS example | LFM2.5-2.6B registered on all three accelerators (CPU / GPU-MLX / ANE) so one model can be compared on identical hardware |
| Display | The Apple engine shows as **NeuRT**, resolved centrally in commons so every SDK inherits it from one place |

### CI

The Apple jobs were failing before this PR for a structural reason worth calling out: `RAC_BACKEND_NEURT` defaults ON on Apple, both modalities are implemented in `neurun`, and `neurun` is a **separate private repo** no CI host can clone — so `macos-debug`, `macos-release`, `ios-device`, `rcli-macos`, `swift-spm` and `python-macos` all died at CMake configure on a `FATAL_ERROR`.

Fixed by adopting the **two-mode pattern `qhexrt` already uses** for exactly this situation ("absent on CI hosts", in its own words):

- **neurun present** → routable engine, unchanged.
- **neurun absent** → only the shell compiles: an all-NULL vtable whose `capability_check` returns `RAC_ERROR_BACKEND_UNAVAILABLE`, so registration is **refused** rather than accepted-and-useless.

The old `FATAL_ERROR` was guarding something real — "an engine that registers, claims Apple availability, and then serves nothing" — and a stub that *refuses registration* closes that hole directly, which is strictly stronger than refusing to build.

**Packaging keeps the hard failure**, because that is where shipping a stub would actually hurt: `build-core-xcframework.sh` refuses unless `RAC_ALLOW_NEURT_STUB=1`, which only the PR build sets. `release.yml` deliberately does not, so a release without the checkout fails loudly instead of shipping an SDK whose ANE engine rejects every registration.

### One behavioural fix worth a look

Renaming the display string to "NeuRT" silently broke **parsing** of `"Core ML"`. `rac_inference_framework_from_string` resolves a value from any of its three representations — wire string, analytics key, display name — and neither `INFERENCE_FRAMEWORK_COREML` nor `coreml` has the space, so previously-persisted records started returning `RAC_ERROR_NOT_FOUND`. Added back as an explicit legacy alias, with both spellings pinned in the test.

## Type of Change
- [x] Bug fix
- [x] New feature
- [x] Refactoring

## Testing
- [x] Lint passes locally

Verified both build modes locally:
- `-DNEURT_ROOT=/nonexistent` → full build clean; `test_plugin_entry_neurt` reports *"ok: stub neurt engine is not advertised or routable"*.
- Real checkout → `rac_neurt_core` + `rac_neurt_llm_ops` build; `test_plugin_entry_neurt` reports *"ok: routable neurt engine serves DIFFUSION + GENERATE_TEXT and pins"*; `test_bundle_policy` 38/38.

`test_plugin_entry_neurt` is deliberately **not** skipped in stub mode — it asserts both arms of the `RAC_NEURT_ROUTABLE` switch, so CI keeps real coverage of the shell rather than skipping the engine entirely.

**Not covered by CI:** actual ANE inference. That needs the private `neurun` checkout, so it stays a local/device gate.

### Platform-Specific Testing
**Swift SDK / iOS Sample:**
- [x] Tested on Mac (macOS target)

## Summary by CodeRabbit

- **New Features**
  - Added NeuRT support for Apple Neural Engine text generation and Core ML image generation.
  - Added local model loading for bundles, files, URLs, and Hugging Face references.
  - Added engine selection for catalog models, including NeuRT, Core ML, and ANE aliases.
  - Registered LFM2.5 model variants across ANE, CPU, and GPU backends.
- **Improvements**
  - Updated Apple SDK packaging and labels to use NeuRT.
  - Improved access to private and gated Hugging Face repositories.
  - Increased the iOS app version to 0.37 and enabled additional memory capacity.


sanchitmonga22:
@coderabbitai please review the PR in DETAIL! thanks!

sanchitmonga22:
## CodeRabbit review — all 20 comments addressed

17 review comments + 3 issue comments (CodeRabbit's auto-summary, the review request, and its ack). **100% accounted for: 14 fixed, 2 verified invalid, 1 deferred with reasoning.** No GitHub issues opened, per instruction.

### Defects the review caught that this branch had introduced

Calling these out separately because they were *not* pre-existing:

1. **Would have broken CI again** — `build-core-xcframework.sh` listed `librac_neurt_llm_ops.a` / `librac_neurt_core.a` unconditionally, but those targets exist only in the routable build, so the stub path this PR adds would fail in `prepare_archive_input()` on a missing file. Both merge functions now include the pair only when present. **Presence, not the `RAC_ALLOW_NEURT_STUB` flag**, is the test — it stays correct however the build was configured, and the shell has no unresolved `U _g_neurt_llm_ops` because its vtable slot is NULL.
2. **`"neurt"` in the telemetry framework dimension** — that is the *engine id*, not a framework value. `RAC_FRAMEWORK_COREML`'s analytics key is `coreml`, and `rac_inference_framework_from_string("neurt")` returns `NOT_FOUND`. Reverted.
3. **A comment asserting the `FATAL_ERROR` behaviour this PR removed.**

### Fixed (14)

| # | Fix |
|---|---|
| R17 | stub-mode xcframework merge (above) |
| R3 | `PACKAGING_ONLY_SYMBOLS` is now per-configuration — unconditional filtering could mask a genuinely missing export in the **distribution** archive, the exact failure the gate exists to catch |
| R7 | local CLI model ids are path-unique (two bundles with the same basename collided and overwrote) |
| R8 | a `.mlpackage` supplied **as** the path is detected, not only when nested — it *is* a directory, so it took the dir branch and silently dropped the CoreML/NeuRT pin |
| R9 | `--engine` now reaches the load request in `bench`/`embed` (was parsed, then ignored for catalog models) |
| R10 | telemetry framework value (above) |
| R14 | HF bearer on the pinned-manifest request — a gated repo 401'd, refs came back empty, and the **whole folder** got registered |
| R2 | stale `NEURT_ROOT` comment |
| R15 | canonical `RAC_ENGINE_ID_NEURT` instead of a string literal |
| R1, R11, R12, R13, R16 | doc/matrix consistency (engine count, capability matrix, xcframework list, `NeuRT` heading, python `backends()` semantics) |

### Verified invalid — with evidence, not assertion (2)

- **R5** (download-size guard): does not apply here, two independent ways. The value is **discarded** on this path — `hf.co/…` is a *folder* ref, so `register_from_hf_folder` prefers the resolver's live tree total over the caller's. And even if it reached the guard, the check is an **80% floor**; the real folder is 3,257,702,729 B = 3.034 GiB against a 3.213 GiB declaration = **94.4%**. (The silero-vad precedent that *does* trip it is 46%.) The secondary suggestion — gate the entry on device capability — was also declined: that is business logic the root `AGENTS.md` forbids in an example app, and no other large entry is gated.
- **R6**: the doc is right that neurt declares `RAC_RUNTIME_ANE` — the code declares it and `test_plugin_entry_neurt` asserts it. The comment conflated "no registered `ane` adapter under `runtimes/`" with "the engine doesn't declare it"; runtime declarations are advisory metadata, as the next paragraph in that same file states.

### Deferred, not silently dropped (1)

**R4 — model catalog data belongs in the SDK.** Legitimate, and it correctly cites the layering rule — but it is not a defect this PR introduced. `ModelCatalogBootstrap.swift` is 1,504 lines / 82 registration calls; this PR adds 3. The same demo catalog lives in the app layer on **all five platforms**, and there is **no SDK-side catalog to move into** — the Swift SDK ships only registration primitives.

The real fix is a cross-platform ABI change: catalog data + a `rac_catalog_seed_*` entry point in commons, a catalog proto with regenerated bindings for 5 languages, one `seedDefaultCatalog()` per SDK, and deletion of 5 example catalogs plus their tests. Doing it for only the 3 LFM2.5 entries would split the catalog across two layers — worse than either end state. Left unchanged rather than half-done.

PR #620, opened 2026-08-04
fix(commons): require archive extensions to match the URL path suffix

shubhamsinnh:
## Description

Corrects archive classification so supported extensions are recognized only at the end of a URL path.

Previously, an extension appearing in a backup filename, checksum filename, or parent path could incorrectly mark a normal file as an archive. That could route a valid download through extraction and produce a misleading failure. Query strings, fragments, and uppercase extensions remain supported.

This change keeps archive handling consistent for every SDK that relies on the shared native layer.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring

## Testing

- [x] Required CI checks pass

### Platform-Specific Testing

Not applicable. This is a shared native behavior change with no user-interface impact.

## Labels

- [ ] `core` - Maintainer permissions required

## Checklist

- [x] Project style guidelines followed
- [x] Self-review completed
- [x] Documentation updated or confirmed unnecessary

## Screenshots

Not applicable.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved archive file detection for paths containing query parameters and fragments.
  * Prevented false positives when archive extensions appear before the end of a path.
  * Added reliable handling for uppercase and alternate archive extensions.

* **Tests**
  * Expanded coverage for query strings, fragments, alternate archive formats, and invalid embedded extensions.

### DIFF
--- a/sdk/runanywhere-commons/src/infrastructure/model_management/model_types.cpp
+++ b/sdk/runanywhere-commons/src/infrastructure/model_management/model_types.cpp
@@ -21,6 +21,16 @@
 #include "rac/core/rac_logger.h"
 #include "rac/infrastructure/model_management/rac_model_types.h"
 
+namespace {
+
+bool ends_with(const std::string& value, const char* suffix) {
+    const size_t suffix_length = std::strlen(suffix);
+    return value.size() >= suffix_length &&
+           value.compare(value.size() - suffix_length, suffix_length, suffix) == 0;
+}
+
+}  // namespace
+
 // =============================================================================
 // ARCHIVE TYPE FUNCTIONS
 // =============================================================================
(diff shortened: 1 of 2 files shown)


shubhamsinnh:
@shubhammalhotra28 archive format detection was matching extensions mid-URL, tightened it to only look at the actual path suffix. happy to adjust the regex if you prefer a different approach.

sanchitmonga22 (approved):
Nice find. `rfind(ext) != npos` matched the extension anywhere in the URL, so `model.zip.backup` classified as a ZIP and `x.tar.gz.txt` as a tarball. Requiring the extension to be the terminal suffix is right, and it brings this in line with the three other archive-detection sites in the repo that already did suffix matching.

One consequence worth stating plainly since it is intended: a URL like `https://host/x.zip/download` no longer auto-extracts. I scanned the shipping catalogs and found nothing relying on that shape, so this should be invisible in practice.

`electron (windows-2022)` and `swift-spm` are red here, and both are red on main independently of this PR.

PR #621, opened 2026-08-04
fix: deep-copy model artifact descriptors

shubhamsinnh:
## Summary

- deep-copy `expected_files` and every multi-file artifact descriptor in `rac_model_info_copy()`
- keep `file_descriptor_count` consistent with descriptor ownership
- normalize owned pointers before allocation so the copy never aliases the source

## Root cause

`rac_model_info_copy()` copied `artifact_info` as a struct, then cleared its descriptor pointer without clearing `file_descriptor_count`. Consumers iterate the array using that retained count, which could dereference a null pointer. A non-null descriptor pointer paired with a zero count could also remain aliased after the struct copy and be freed twice.

## Impact

Model assignment caching and other paths using `rac_model_info_copy()` now preserve multi-file artifact metadata with independent ownership. Because the fix is in shared C++ commons, it benefits all platform SDKs.

## Validation

- `git diff --check`
- reviewed allocation/free symmetry against `rac_expected_model_files_free()` and `rac_model_file_descriptors_free()`

C++ lint was not run because `clang-format` is unavailable in the local environment. The existing local CMake build directory was incomplete and could not be regenerated within the command window.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model information copying to preserve artifact metadata, including file patterns, descriptions, and descriptors.
  * Ensured copied model details remain complete and consistent when metadata is missing or resources cannot be allocated.
  * Fixed model artifact copies so they remain independent from the original data.
  * Improved handling of empty or inconsistent artifact metadata during model save and retrieval.

### DIFF
--- a/sdk/runanywhere-commons/src/infrastructure/model_management/model_registry_convert.cpp
+++ b/sdk/runanywhere-commons/src/infrastructure/model_management/model_registry_convert.cpp
@@ -9,6 +9,7 @@
  */
 
 #include "model_registry_internal.h"
+#include "model_types_internal.h"
 
 #include <cstdlib>
 #include <cstring>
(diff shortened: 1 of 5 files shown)


shubhamsinnh:
@shubhammalhotra28 this one swaps shallow copies for deep copies on the model descriptor path, prevents a dangling pointer if the source list moves.  Happy to make any changes if required.

sanchitmonga22 (approved):
This found a genuine bug and I am glad you filed it. The shallow artifact copy lost multi-file metadata silently, and worse, it left `file_descriptors` and `file_descriptor_count` disagreeing, which is a null deref waiting at `model_registry_manifest.cpp:172`.

I reworked the implementation before merging. Your patch hand-rolled a third copy of the same walk, and the repo already had two; a fourth divergent copy is how the next version of this bug gets written. It now routes through one shared helper used by both `rac_model_info_copy` and `deep_copy_model`. I also closed an OOM path your version widened, where a failed allocation left the pointer set with no NULL check downstream.

There is a regression test now, and it earns its place. Reverting the fix and re-running it gives `FAIL @ test_model_artifact_copy.cpp:154` followed by SIGSEGV, so it catches the actual crash rather than just asserting the happy path. Full suite is 102/102 green.

`electron (windows-2022)` is red, with the same `dumpbin /EXPORTS` exit 157 signature the last four runs on main show. Not from this change.

Thanks for the report and for the patience on the rewrite.

PR #631, opened 2026-08-05
test(qhexrt): enumerate lfm2_5_2_6b under v75 and v81

ayaangazali:
## What

`test_native_catalog_owns_arch_and_auth_policy` checks per-arch support against exact set membership:

```cpp
ASSERT_EQ(rac_qhexrt_catalog_model_supports_arch(id.c_str(), RAC_QHEXRT_HEXAGON_ARCH_V75),
          v75.count(id) == 0 ? RAC_FALSE : RAC_TRUE);
```

`lfm2_5_2_6b` is enumerated only in the `v79` set, but the catalog grants it every supported arch:

```cpp
{"lfm2_5_2_6b", kAllSupportedArches, false},   // kV75 | kV79 | kV81
```

and the accessor answers straight off that mask:

```cpp
return policy != nullptr && (policy->arch_mask & arch_mask(arch)) != 0 ? RAC_TRUE : RAC_FALSE;
```

So for V75 and V81 the catalog returns `RAC_TRUE` while the test expects `RAC_FALSE`. Two assertions contradict the policy they exist to pin.

Looks like a seam between the two PRs: #624 added v79 support and enumerated it, then #628 widened the policy to v75/v81 (its title and the row's comment both say so) without extending the sets.

The union feeds the count assertion, and the id was already in `v79`, so `all.size()` is unchanged by this and `ASSERT_EQ(all.size(), rac_qhexrt_catalog_model_count())` still holds.

## Why it went unnoticed

The target is gated on the private Hexagon backend:

```cmake
if(RAC_BACKEND_QHEXRT AND TARGET rac_backend_qhexrt AND NOT TARGET test_qhexrt_model_catalog)
```

so it is not built by the macOS or Linux CI lanes, and `main` stays green with the contradiction in place. It surfaces on a build that actually enables the qhexrt backend, which is the Android NPU configuration this row is for.

## Testing, and the gap

**I could not execute this test.** Building it needs `RAC_BACKEND_QHEXRT=ON` and the private QHexRT artifacts, which I do not have. So rather than imply a run I did not do, here is what I verified by reading the three relevant places, quoted above: the policy mask, the mask constant (`kAllSupportedArches = kV75 | kV79 | kV81`), and the accessor. Those three make the outcome deterministic without running anything.

The edit itself adds two string literals to existing initialiser lists, and I confirmed the id now appears exactly once in each of the v75, v79 and v81 sets.

If you would rather narrow the policy than widen the test, say so and I will flip it the other way. I assumed the policy is correct because #628's title is the v75/v81 eligibility and the row's comment spells out that v75 and v81 ship the two-part hybrid decode split while v79 ships the monolithic decode.

## Summary by CodeRabbit

* **Tests**
  * Added coverage for the `lfm2_5_2_6b` model in additional QHexRT model catalog test sets.


sanchitmonga22:
Closing as already-landed: a byte-identical hunk went in via `ac9aebbcd5` (#641) four days after you opened this. Thanks for catching it independently.

One thing worth knowing for next time: this patch wouldn't have gone green on its own. `cosmos3_edge_text`, `cosmos3_edge_vlm` and `cosmos3_edge_diffusion` had the same V79 mismatch, and #641 had to fix all four together.

**Follow-up I'd genuinely like:** the arch expectations are three hand-copied `unordered_set` literals, which is why they drift. Deriving them from the policy table instead would make this class of mismatch impossible. Related — the `all.size() == model_count()` guard is structurally blind to a mask widening, so it can't catch the failure it looks like it's guarding. Either of those would be a real improvement.

PR #633, opened 2026-08-06
fix(rcli): complete the FNV-1a digest behind local model ids

ayaangazali:
## What

`id_for_local_path` pins a local model's id to its full path, because the basename alone collides:

> The basename alone collides — every `.../model.mlpackage` sanitizes to the same string — so pin the id to the whole path with an FNV-1a digest.

Two things stopped it being that digest.

**1. The offset basis is a digit short.**

```cpp
uint64_t hash = 1469598103934665603ULL;
```

The FNV-1a 64-bit basis is `14695981039346656037` (`0xcbf29ce484222325`), 20 digits. The literal is that value with the trailing digit dropped, 19 digits. The prime on the next line is the correct `1099511628211`, so the intent is unambiguous.

**2. Half the hash never reaches the id.**

```cpp
for (int shift = 28; shift >= 0; shift -= 4) {
```

That emits 8 nibbles, so only bits 31-0 of a 64-bit hash are used. A 64-bit value needs shifts 60 down to 0.

## Why it matters

The digest exists so two bundles sharing a basename get distinct ids. Truncating to 32 bits shrinks that guarantee from 2^64 to 2^32, and paths differing solely above bit 31 collide outright. When they do, the second registration lands on the first one's id.

Concretely, for two different paths with the same basename:

```
/Users/me/models/model.mlpackage
  before: local-model.mlpackage-eb4d5469          (8 hex)
  after : local-model.mlpackage-f57ae22d33b5f963  (16 hex)

/var/lib/ra/model.mlpackage
  before: local-model.mlpackage-555bb248          (8 hex)
  after : local-model.mlpackage-3377f1976fc23a32  (16 hex)
```

## The one behavioural consequence, stated plainly

This changes the id computed for a given path, so a local model registered by an earlier build resolves under its old id. Ids are derived from the path on every run rather than stored as an input, so there is nothing to migrate; the effect is that a stale registry row from a previous build would no longer be matched. Given `rcli` is a developer CLI I judged that acceptable, but it is your call, and it is unavoidable for either half of the fix since both change the digest.

## Testing

Configured with `-DRAC_BUILD_CLI=ON -DRAC_BUILD_TESTS=ON -DRAC_DESKTOP_ADAPTER=ON` and built with Ninja:

- `test_rcli_unit --run-all` — **25 passed, 0 failed**

No existing test covers `id_for_local_path`, and I did not add one: asserting a specific digest string would pin the algorithm's output rather than its property, and the useful property (two paths sharing a basename get different ids) already held before this change. The constants above are checkable by eye against the FNV-1a spec.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved identifiers for locally stored models by using the complete hash value.
  * Reduced the risk of duplicate identifiers for models with similar or lengthy paths.
  * Standardized identifiers to a consistent 16-digit hexadecimal format for more reliable model references.

### DIFF
--- a/sdk/runanywhere-cli/src/catalog/model_ref.cpp
+++ b/sdk/runanywhere-cli/src/catalog/model_ref.cpp
@@ -98,14 +98,16 @@ std::string id_for_local_path(const std::string &path) {
 
   // The basename alone collides — every `.../model.mlpackage` sanitizes to the
   // same string — so pin the id to the whole path with an FNV-1a digest.
-  uint64_t hash = 1469598103934665603ULL;
+  uint64_t hash = 14695981039346656037ULL;  // FNV-1a 64-bit offset basis
   for (const unsigned char c : full) {
     hash ^= c;
     hash *= 1099511628211ULL;
   }
   static constexpr char kHex[] = "0123456789abcdef";
   std::string digest;
-  for (int shift = 28; shift >= 0; shift -= 4) {
+  // 16 nibbles: emit the whole 64-bit hash. Stopping at shift 28 kept only the
+  // low 32 bits, so paths differing above that bit shared an id.
+  for (int shift = 60; shift >= 0; shift -= 4) {
     digest.push_back(kHex[(hash >> shift) & 0xF]);
   }
   return "local-" + base + "-" + digest;


ayaangazali:
These failures are a GitHub Actions outage, not this diff. Every failing job died on the same step, before any of them reached the code:

```
Failed to resolve action download info. Error: Service Unavailable
##[error]Service Unavailable
```

That is identical across `rcli-macos`, `kotlin-android`, `linux-asan`, `centralization` and `Gitleaks Secret Scan`, with nine further jobs cancelled as collateral. `rcli-macos` never got as far as building, so nothing here exercised the change.

Two other signals that it is platform-side: another branch (`siddhesh/sdk-fixes-v2`) has four runs failing in the same window, and `Legacy file blocklist` passed on this branch during it. `main` has no failing checks.

A run is still in progress, so I am leaving it to finish rather than force-pushing over it. If it settles red on the same error I will re-trigger with an identical diff.

For what it is worth locally: `test_rcli_unit --run-all` is 25 passed, 0 failed with this change, configured with `-DRAC_BUILD_CLI=ON -DRAC_DESKTOP_ADAPTER=ON`.

ayaangazali:
Adding the decisive detail, since these keep getting read as code failures.

Each of the six failing jobs contains exactly **one** step, and it is `Set up job`:

| job | steps |
|---|---|
| centralization | `failure: Set up job` |
| Gitleaks Secret Scan | `failure: Set up job` |
| linux-asan | `failure: Set up job` |
| rcli-macos | `failure: Set up job` |
| android-arm64 | `failure: Set up job` |
| kotlin-android | `failure: Set up job` |

There is no checkout step, no build step and no test step in any of them. The runner failed while resolving the actions themselves:

```
Failed to resolve action download info. Error: Service Unavailable
```

So the branch was never fetched onto the runner, and nothing in this diff was compiled or executed. A one-line change in `model_ref.cpp` cannot affect `Set up job`, and it certainly cannot affect Gitleaks or kotlin-android.

Still not pushing a speculative fix. These need a re-run once the platform is healthy, which is not something I can trigger from a fork.

ayaangazali:
Re-run cleared the outage: the six `Set up job` failures are gone and 11 checks are green. One new failure came up, `rcli-windows`, and it is also environmental.

The Build step died linking a target this diff does not touch:

```
[1081/1233] Building CXX object sdk/runanywhere-cli/CMakeFiles/rcli_core.dir/src/catalog/model_ref.cpp.obj
...
[1213/1233] Linking CXX executable sdk\runanywhere-commons\tests\test_core.exe
FAILED: [code=1] sdk/runanywhere-commons/tests/test_core.exe
C:\mingw64\bin\c++.exe ... -o sdk\runanywhere-commons\tests\test_core.exe ...
Access is denied.
```

Three things there: `model_ref.cpp` (the only file I changed) compiled fine at step 1081; the failure is `test_core.exe` in commons tests, which this PR does not touch; and `Access is denied` from the linker invocation is a filesystem lock on the output, not a compile or link error in the code. Other targets kept linking after it (1214 through 1218), which is what a one-off locked output looks like rather than a broken build.

`rcli-windows` is also green on all nine of my other open PRs, including #625 which touches commons, so it is not a systemic break.

Re-triggering with an identical diff again. If Windows locks a different output next time, that is the runner and not this change.

sanchitmonga22 (approved):
Reproduced the digests in the PR body exactly. Both bugs are real: the digest was truncated to 32 bits and the FNV offset basis was typo'd, so local model ids collided far more readily than intended. All three `rcli-*` jobs green. Thanks — nice catch.

PR #638, opened 2026-08-07
idl: reserve LLMGenerationResult tag 7 after deleting ttft_ms

ayaangazali:
## What is wrong

#605 deleted `LLMGenerationResult.ttft_ms` (tag 7) so that `TokenUsage.ttft_ms` is the single home for the metric, but did not reserve the vacated tag:

```proto
message LLMGenerationResult {
    ...
    double generation_time_ms  = 6;
    optional string framework  = 9;
    reserved 10;  // was string finish_reason -- wire-type flip to enum, fresh tag below
```

Tag 7 is simply absent.

## Why that is wrong

The same commit reserved the equivalent tag in the sibling message, so the intent is not in question:

```proto
// vlm_options.proto, VLMResult
reserved 8;  // was time_to_first_token_ms -- TokenUsage.ttft_ms is the
             // canonical spelling (usage = 15)
```

`LLMGenerationResult` already uses the pattern for tag 10. Without the marker a later field can claim 7, and payloads written by a current build carry a `double` there, so an older reader decodes the new field as the stale TTFT rather than failing.

The proposal doc for this change flags it directly: *"Before deleting any TTFT field, add `reserved` for llm_options tag 7, llm_service tag 7 and vlm_options tag 8 — the after-text has none."* Two of the three are now handled: `vlm_options` got the marker, and `llm_service.LLMStreamFinalResult` needs nothing because the whole message was deleted. `llm_options` is the one left.

## What this does

Adds the marker and commits the regenerated bindings it produces.

## Correction to an earlier claim in this PR

The first revision of this description said no regeneration was needed because `reserved` is compile-time only. **That was wrong, and `idl-drift-check` correctly failed on it.**

protoc embeds the reserved ranges in the serialized `FileDescriptorProto` that generators emit, so three files legitimately change: `llm_options.pb.cc`, `llm_options_pb2.py`, and `llm_options.pb.swift`. My check had been to grep the generated Swift for the word "reserved" and find none, but the ranges are descriptor bytes rather than readable text, so absence of the word proved nothing. In the Swift name-map bytecode the added entry is `\u{c}\u{7}\u{1}` (reserved-range opcode, start 7, count 1), sitting next to the `\u{c}\u{a}\u{1}` already present for tag 10.

This revision also collapses the tag 7 comment to one line. protoc attaches only the first line of a trailing comment to a `reserved` statement, so the continuation leaked onto the following field and generated Swift and TS documented `framework` as `/** spelling (usage = 25) */`. With the comment on one line, Kotlin, Dart and proto-ts regenerate byte-identical to `main`, which is why those files are absent from the diff.

## Testing

- `protoc --proto_path=idl idl/llm_options.proto` compiles clean.
- Injecting `double sneaky_reuse = 7;` into a copy fails as intended, confirming the marker now enforces:

```
llm_options.proto:116:14: Field "sneaky_reuse" uses reserved number 7.
```

- Re-ran `./idl/codegen/generate_all.sh` on the committed tree with the versions CI pins (protoc 35.1, wire-compiler 5.5.1, grpcio-tools 1.71.2, protoc_plugin 25.0.0) and `git status --porcelain --untracked-files=all` comes back empty, which is the condition the drift gate asserts.

One gap worth stating plainly: I could not run the `sdk/shared/proto-ts` `tsc` build locally, because `npm install` there resolves against the monorepo workspace root and hits an unrelated `react-native-worklets` peer conflict. That step is a no-op for this change though, since `proto-ts` sources regenerate byte-identical to `main`.

## Summary by CodeRabbit

* **API Changes**
  * Removed the deprecated time-to-first-token field from generation results.
  * Preserved compatibility by reserving the former field number.
  * Time-to-first-token metrics remain available through the canonical token usage data.
  * Updated the Python SDK’s generated API metadata to reflect this change.


ayaangazali:
Pushed 1cc933c — this fixes the `Verify generated code matches IDL` failure, which was mine.

I had claimed in the description that `reserved` needs no regeneration. That was wrong. protoc embeds reserved ranges in the serialized `FileDescriptorProto`, so `llm_options.pb.cc`, `llm_options_pb2.py` and `llm_options.pb.swift` all legitimately change. My check had been grepping the generated Swift for the word "reserved", but the ranges are descriptor bytes rather than text, so finding nothing proved nothing.

The commit also collapses the tag 7 comment to a single line. protoc attaches only a trailing comment's first line to `reserved`, so the continuation leaked onto the next field and generated Swift and TS were documenting `framework` as `/** spelling (usage = 25) */`. With that fixed, Kotlin, Dart and proto-ts regenerate byte-identical to `main`, so only the three descriptor-carrying files appear in the diff.

Regenerated with the versions CI pins (protoc 35.1, wire-compiler 5.5.1, grpcio-tools 1.71.2, protoc_plugin 25.0.0); re-running `generate_all.sh` on the committed tree leaves the working tree clean. I have corrected the description so the inaccurate claim is not left standing.

sanchitmonga22:
Closing this one, but I want to be clear it was **right** — and it's the most dangerous of the batch, which is worth explaining.

#661 landed `reserved 3, 4, 7, 8, 17, 18, 19;` at `idl/llm_options.proto:94-96` — a strict superset of what this PR reserves, with the descriptor bytes committed. So the outcome you wanted is in `main`.

The trap: this PR's `.proto` change **auto-merges cleanly** into a duplicate `reserved 7`, and protoc rejects duplicate reserved ranges. So merging it now would be a silent build-breaker rather than a harmless no-op. Flagging that explicitly because nothing in CI would have told us before the merge button.

Two things you documented that nobody else in this repo had written down, and that I'd like to keep:
- protoc embeds reserved ranges in the descriptor (so they are not merely advisory)
- protoc attaches only the *first* comment line to a `reserved` statement

**Concrete follow-up if you want it:** `main` reserves the *numbers* but not the *names*. And more usefully — a validation gate that fails when a field number disappears from a message without a matching `reserved` marker would have caught the seven holes #605 left behind. That gate is a genuinely valuable contribution and I'd merge it.

PR #641, opened 2026-08-08
UI revamp

sanchitmonga22:
## Summary

Full UI/UX revamp of the four RunAnywhere example apps — iOS, macOS, Android, Web — plus the
underlying SDK/commons fixes their voice, download, and VLM flows exposed along the way.
58 commits, 285 files, ~35K insertions / ~10K deletions across
`examples/{ios,android,web}`, `sdk/{runanywhere-commons,runanywhere-swift,runanywhere-kotlin,runanywhere-web}`.

Brand identity (`#FF6900`, the logo) is unchanged per `examples/DESIGN_GUIDELINE.md`; everything
else — layout, typography, motion, components, information architecture — was rebuilt or fixed.

### iOS / macOS (`examples/ios/RunAnywhereAI`, `sdk/runanywhere-swift`)
- New design-system foundation: motion tiers, type scale, layout, color — shared by iOS and macOS.
- macOS got a real sidebar shell, a menu bar, a Form-based Preferences window, and a proper Mac
  type scale, replacing the modal-drawer/phone-in-a-window layout it shipped with.
- Chat: one scroll driver, real message actions (copy/retry/edit/delete), metrics that don't
  disappear mid-render, markdown rendered as blocks instead of a per-token guess.
- Voice: barge-in, partial-transcript marking, the voice agent's silence-marker fabrication fixed
  (it was answering "You said: The." on room noise), TTS `speak()` now spans playback rather than
  just synthesis.
- Downloads: background `URLSession`, verified with a real 1.9GB transfer surviving app
  suspension. The `cancel(byProducingResumeData:)` bug's real cause was a 401-byte cache key
  hitting `NAME_MAX` and failing `ENAMETOOLONG` silently behind a `try?` — not the async-timing
  hypothesis it was first assumed to be.
- VLM: camera session lifecycle, chat-with-image/files, MLX Qwen2-VL 2B found broken on Metal
  (was the recommended vision model) and dropped from recommendations in favor of a working one.
- New `AudioRouteMonitor` SDK API (`AVAudioSession` on iOS, CoreAudio on macOS) surfaces which
  speaker/mic the system has selected, and flags virtual/aggregate devices that silently produce
  digital silence as input.
- Fixed the XCFramework build script bundling two Debug+Release copies of every abseil object
  (1204 duplicate-symbol link errors) — a config-filter bug, not a code regression.

### Android (`examples/android/RunAnywhereAI`, `sdk/runanywhere-kotlin`)
- Cold start cut from 14.1s to 0.5s; boot screen rebuilt.
- One motion system replacing three separate VU-meter implementations; a real drawn waveform.
- Chat: visible copy/retry/edit/delete on every turn; markdown matches web's delimiter-run
  parsing (fixes streaming italic/bold flicker).
- Model picker: models listed first, sizes no longer mislabelled.
- Download: foreground service with rate/ETA/retry from `DownloadEvent.Progress`; fixed a
  preempted-transfer-reported-as-completed race in `ModelDownloadService`.
- Voice: fixed audio capture always selecting `VOICE_COMMUNICATION`, which the emulator (and
  possibly some devices) initializes successfully and then silently yields all-zero samples —
  now gated on whether the device actually has an `AcousticEchoCanceler`.
- VLM: photo-picker image chat verified working end-to-end (streamed reply, Sources card for
  grounded file answers).

### Web (`examples/web/RunAnywhereAI`, `sdk/runanywhere-web`)
- Text contrast ramp fixed to clear WCAG AA at every step; every control reachable by touch/keyboard.
- Truthful state everywhere a previous build faked or hid one: Solutions readiness, engine
  registration failure, the "no model" card no longer hiding the composer, Settings controls
  actually named and functional.
- Documents: real drag/paste file support, one honest error surface.
- Downloads: stop/resume from where it stopped; the acceleration badge demoted from a
  z-index-140 overlay into a real drawer row.
- Fixed the actual root cause of plain-file (`.gguf`) downloads failing app-wide with zero
  network requests issued: `rac_download_plan_proto` returned without writing its out-buffer
  (an Asyncify unwind signature) because the size probe is synchronous; routed through
  `ccall({async:true})`. Verified with 6 models downloaded byte-exact against HF content-length.
- Fixed VLM streaming being dead in every browser session: the worker's proto-event trampoline
  registered both LLM's `void`-returning callback and VLM's `bool`-returning callback as the same
  Emscripten function signature (`'viii'`), when VLM's needs `'iiii'` — a WASM function-table type
  mismatch that trapped on every call.
- Voice: barge-in now measured working on real speaker→mic acoustics (not just a loopback), VAD
  fixed (`detectStream` drained the whole input before yielding; the lifecycle configure step
  rejected the loaded Silero model).

### Commons (`sdk/runanywhere-commons`)
- `download_orchestrator.cpp`: a failed HEAD content-length probe was treated as fatal instead of
  as an optimization, killing the whole download plan on any platform-transport fault.
- `rac_stt_stream.cpp` fallback segmenter retuned from measurement (ambient-multiplier gate was
  9.5dB above typical speech in a real room, so 8 seconds of clear speech produced no window at
  all) — verified against 88/88 commons tests, whose own fixtures were rewritten to hand the
  segmenter a measured room tone instead of pure zero-silence.
- Voice agent: stopped publishing an engine's non-speech marker (`[BLANK_AUDIO]`, `[ Silence ]`)
  as a real transcript, which was making the agent answer questions nobody asked.

## Cross-cutting

- All 31 review comments on this PR addressed (fixed or rebutted with evidence where the
  reviewer's premise didn't hold against the current code) — no follow-up issues filed.
- Voice, download, and VLM flows measured hands-on across all four apps with real speaker/mic
  acoustics where the platform allows it (iOS Simulator audio output is not detectable at the
  host mic — measured, not assumed — so iOS voice acoustics need a physical device).
- Merged `main` (through #651) with one real conflict in `RunAnywhereTextGeneration.kt`; resolved
  in favor of the branch's `sanitizeStreamMetrics` path, which is a strict superset of main's
  decode-time fix (#634) plus a batch-buffered-backend guard main's version didn't have — verified
  by running `RunAnywhereTextGenerationStreamTest` (6/6 passing) rather than assumed.

## Testing

- [x] `sdk/runanywhere-commons`: `ctest` 88/88
- [x] `sdk/runanywhere-swift`: `swift build` clean; iOS Simulator + macOS `xcodebuild` both green
- [x] `sdk/runanywhere-kotlin`: unit tests green; Android debug build green
- [x] `sdk/runanywhere-web` + `examples/web`: lint + typecheck clean (production `build` needs an
  ONNX Runtime+Dawn WebGPU archive not available in this environment; `npm run dev` unaffected)
- [x] Hands-on device/simulator/emulator validation of voice, download, and VLM flows on all four
  targets, with real audio where the platform supports it

### Platform-Specific Testing
**Swift SDK / iOS Sample:**
- [x] Tested on iPhone (Simulator)
- [x] Tested on Mac (macOS target)

- [x] Tested on Android Phone (Emulator)

**Web SDK / Web Sample:**
- [x] Tested in Chrome


Siddhesh2377:
@sanchitmonga22 this conflicts with main now, and that was my #634 merge, sorry. Only `RunAnywhereTextGeneration.kt` collides. Take main's version of that hunk and you're clear.

PR #649, opened 2026-08-09
chore: release 0.20.14

Siddhesh2377:
Release 0.20.14. Merging this with the `release:patch` label pushes `v0.20.14`, which triggers `release.yml` to build and publish every artifact.

This is the first release that should publish **without a manual upload**: #644 gave CI the `neurun` checkout it needs to build the real NeuRT engine, and untracked the local-natives marker that broke remote consumption.

### What's in it

- **`RunAnywhereNeuRT`** — the Apple Neural Engine backend as a standalone Swift package, alongside core, LlamaCPP, ONNX, and MLX. Registered in the iOS example and verified on a running app.
- The `.runanywhere-local-natives` fix, so a fetched package resolves the release zips instead of looking for local `Binaries/`.
- CI builds NeuRT routable, pinned to neurun `1c44fd15` — the same commit these checksums were built against.

### Verified locally

Routable Apple build (real NeuRT, not a stub) produced all 9 zips. Then a link audit over the shipped archives:

- **Slices** — all six xcframeworks carry `ios-arm64`, `ios-arm64-simulator`, and `macos-arm64`.
- **Cross-archive symbols** — every undefined `rac_*` reference in every archive is defined by another archive in the set, on both the iOS and macOS slices. (The one flag, `descriptor_table_rac_options_2eproto`, is protobuf internals that RACommons defines itself.)
- **`_rac_plugin_entry_neurt`** — defined in all three NeuRT slices. Commons hard-references this symbol and the app links with `-all_load`, so a missing definition would break every consumer's link.
- **Checksums** — all six zips match the values committed in `Package.swift`, which is what SPM verifies on a consumer machine and what the publish job re-checks against its own rebuild.
- **SPM link** — `RunAnywhere`, `LlamaCPPRuntime`, `ONNXRuntime`, `NeuRTRuntime`, `MLXRuntime` all build.

Plus: 86 Swift tests, 0 failures, 0 skipped; SwiftLint `--strict` clean; version-coherence gate green.

### Known, not fixed here

- Installing by SemVer (`from: "0.20.14"`) still fails: the manifest revision-pins the `PrismML-Eng/mlx-swift` and mlx-audio-swift forks, and SPM refuses a version-pinned package that depends on a revision-pinned one. Consumers use a commit or branch ref until those forks publish tags.
- The NeuRT archive is still linked into `ONNXRuntime` as well as `NeuRTRuntime`, because commons unconditionally references `rac_plugin_entry_neurt`. Dropping it from ONNX needs the static-register shim made conditional in C++ — a follow-up, not a manifest change.

## Summary by CodeRabbit

* **Release**
  * Released SDK suite version **0.20.14** across Flutter, Kotlin, Python, React Native, Swift, and Web platforms.
  * Updated backend packages and runtime integrations to the corresponding 0.20.14 versions.
  * Refreshed installation instructions, dependency requirements, package metadata, and release checksums.
* **Documentation**
  * Added 0.20.14 changelog entries and updated documented version references throughout the SDK guides.


Siddhesh2377 (commented):
pass

PR #626, opened 2026-08-05
fix(flutter): propagate tool-calling cancellation failures

shubhamsinnh:
### Problem

`DartBridgeToolCalling.cancelSession()` documents that it returns `false` when the native cancellation call fails. The implementation discarded the `rac_result_t` returned by `rac_tool_calling_session_cancel_proto` and returned `true` for every non-throwing FFI invocation. Ordinary native failures are returned as result codes rather than Dart exceptions, so callers could receive a successful cancellation result even when the native call reported failure.

### Root cause

The bridge treated the absence of an FFI exception as operational success instead of checking the structured native result code.

### Change

- Capture the cancellation return value.
- Return `true` only when it equals the generated `RacResultCodes.success` constant.
- Preserve the existing exception handling and warning log.

This is limited to the Flutter tool-calling bridge. It does not change the public API, native ABI, cancellation semantics, or behavior of other SDKs. No documentation or migration change is required because the implementation now matches the existing documented return contract.

### Expected and actual behavior

- Expected: `cancelSession()` returns `true` for `RAC_SUCCESS` and `false` for a non-success native result or an FFI exception.
- Before: every non-throwing native call returned `true`, regardless of its result code.
- After: the returned Boolean reflects the native result code.

### Related issue

No linked issue. This was found by comparing the bridge's documented return contract with its implementation. Searches for the native symbol and tool-calling cancellation found no open issue or overlapping pull request.

### Risk and reviewer guidance

Risk is low: the change is a direct comparison against the generated success constant already used throughout the Flutter bridge. Please verify the Boolean mapping at `cancelSession()` and the generated result-code import.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring

## Testing
- [ ] Lint passes locally — Flutter, Dart, Melos, and FVM are not installed on this host, so local Flutter analysis could not be run.
- [ ] Added/updated tests for changes — no unit test was added because repository guidance asks contributors not to add unit tests unless specifically requested; this is a focused bridge result-code correction.

Validation completed:

- `git diff --check origin/main...HEAD` — passed.
- Confirmed the Dart FFI binding returns an integer `rac_result_t`.
- Confirmed `RacResultCodes.success` is the generated success constant with value `0`.
- Reviewed the one-file diff for unrelated changes and prohibited attribution references — none found.

Recommended CI validation:

- Flutter analyzer and repository validation gates.

### Platform-Specific Testing (check all that apply)
[...]

PR #654, opened 2026-08-09
fix(web): check the tool-session cancel result before latching it

ayaangazali:
## What is wrong

`dispatchCancel` throws away the result code from `_rac_tool_calling_session_cancel_proto` and latches the cancel regardless:

```ts
try {
  module._rac_tool_calling_session_cancel_proto(sessionHandle);
  cancelDispatched = true;
} catch (err) {
  logger.warning(`session_cancel_proto failed: ...`);
}
```

The export returns `rac_result_t`, declared in `EmscriptenModule.ts` as `(sessionHandle: bigint): number`. A WASM export returns a non-zero code on failure, it does not throw, so the `catch` never sees it.

## Why that is wrong

`cancelDispatched` is the guard that stops the cancel being attempted again:

```ts
if (
  sessionHandle !== 0n &&
  !cancelDispatched &&
  typeof module._rac_tool_calling_session_cancel_proto === 'function'
)
```

So a cancel that the native side rejected is recorded as delivered. `dispatchCancel` is reachable from three places in this function (the `abort` listener, the already-aborted check after session create, and a `queueMicrotask` on the mid-stream abort path), and every one of them becomes a no-op afterwards. The user aborted, the session was never cancelled, and nothing tries again.

This is the same defect that was just fixed for Flutter in #626, which replaced the discarded call with an explicit success comparison:

```dart
-      fn(sessionHandle);
-      return true;
+      return fn(sessionHandle) == RacResultCodes.success;
```

The Dart doc comment for that method already stated the contract this Web path was not honouring: *"Returns false when the native cancellation call fails."*

## What this does

Latches only on `rc === 0` and warns otherwise, leaving a later abort free to retry. `0 = success` matches the convention already used in this package, for example `VoiceAgentStreamAdapter` treating `rc !== 0` from `_rac_voice_agent_set_proto_callback` as failure.

I deliberately left the worker-backed `dispatchCancel` earlier in the file alone. It goes through `host.infer('tool.sessionCancel', ...)`, whose failures surface as a rejected promise that is already caught and logged, so it is a different mechanism rather than the same bug.

## Testing

I could not run the unit suite on this machine. `vitest run tests/unit/Public/Extensions/RunAnywhere+ToolCalling.test.ts` fails to load with `Cannot find module '@bufbuild/protobuf/wire'` after a clean `npm install` in `sdk/runanywhere-web`. That failure is pre-existing and not from this change: I re-ran it with my diff stashed and got the identical "Test Files 1 failed, Tests no tests" result. Flagging it rather than implying the suite is green locally.

Worth noting that the existing test's fake returns `0` from that export, so it exercises the success path and its expectations are unchanged by this diff.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session cancellation reliability.
  * Failed cancellation attempts are now logged and remain eligible for retry.
  * Cancellation requests that fail during communication or processing can be retried automatically.
  * Successful cancellation requests are tracked correctly, reducing the risk of duplicate or missed cancellation handling.

### DIFF
--- a/sdk/runanywhere-web/packages/core/src/Public/Extensions/RunAnywhere+ToolCalling.ts
+++ b/sdk/runanywhere-web/packages/core/src/Public/Extensions/RunAnywhere+ToolCalling.ts
@@ -887,8 +887,17 @@ export const ToolCalling = {
         typeof module._rac_tool_calling_session_cancel_proto === 'function'
       ) {
         try {
-          module._rac_tool_calling_session_cancel_proto(sessionHandle);
-          cancelDispatched = true;
+          // The export returns rac_result_t (0 = success). Latching
+          // `cancelDispatched` on a non-zero code would record a cancel that
+          // never happened and stop every later abort from retrying it.
+          const rc = module._rac_tool_calling_session_cancel_proto(sessionHandle);
+          if (rc === 0) {
+            cancelDispatched = true;
+          } else {
+            logger.warning(
+              `session_cancel_proto returned ${rc}; leaving the cancel undispatched so a later abort can retry`,
+            );
+          }
         } catch (err) {
           logger.warning(
             `session_cancel_proto failed: ${err instanceof Error ? err.message : String(err)}`,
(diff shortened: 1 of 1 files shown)


sanchitmonga22 (approved):
Thanks for this. Merging with one correction pushed on top.

I checked every branch of the native cancel function and they all return `RAC_SUCCESS`, so there is no live path today where a rejected cancel gets latched as delivered. That makes this hardening rather than a bug fix, and I edited the comment to say so. The guard is still worth having: the moment that function grows a real failure branch, the old code would have silently lied about it. I also swapped the bare `0` for the named result symbol.

One thing left undone on purpose: the PR description still describes this as a live defect. I did not rewrite your description, so read the merged comment as the accurate version.

`swift-spm` is red here and on main, unrelated to this change.

PR #659, opened 2026-08-10
fix(commons): optimize VLM streaming hot path & fix VAD session leak (#647)

### LINKED ISSUE #647: Core C++ hot-path has significant performance bottlenecks and memory leaks that hurt throughput and RAM on mobile
I spent some time reading through the `runanywhere-commons` C++ core deeply — specifically the inference hot path, lifecycle management, and stream dispatch layers. Really impressive architecture overall, but I found several issues that are compounding on each other and causing unnecessary overhead on every single generated token. These aren't edge cases — they hit on every LLM/VLM/Voice generation call.

Sharing the highest-impact ones below. Happy to discuss or help with fixes if useful.

## Issue 1: Full protobuf serialize + deserialize on every streaming token

**Files**: `src/features/llm/rac_llm_stream.cpp`, `llm_module.cpp:840+`, `vlm_module.cpp`, `stt_module.cpp`

Right now every single token that comes out of the LLM backend goes through:
1. String append to `ctx->full_text`
2. Full `LLMStreamEvent` protobuf construction + `SerializeToArray`
3. Raw bytes copied across the FFI boundary (JNI / Dart FFI / JSI / WASM)
4. Full proto decode on the SDK side

At 100-200 tok/s during generation, this is ~200 heap allocations per second just for the stream event plumbing. The `thread_local` proto reuse helps, but the output vector and the SDK-side decode still allocate fresh every time.

A flat 32-byte struct (seq + event_kind + token_len + token chars) would eliminate all proto overhead from the hot loop. Proto can stay for cold-path results and errors.

## Issue 2: Global mutex taken on every token dispatch + lifecycle queries all serialize through one lock

**Files**: `src/features/llm/rac_llm_stream.cpp:68`, `src/core/model_lifecycle.cpp:49-60`

The stream dispatch path does this per token:
```cpp
std::lock_guard<std::mutex> lock(g_mu());  // taken 100+ times/sec during generation
auto it = g_slots().find(handle);
// snapshot callback
```

And the model lifecycle has:
```cpp
std::mutex g_lifecycle_mutex;         // ALL modalities share this
std::mutex g_load_admission_mutex;    // ALL loads serialize here
```

So even checking `is_loaded()` on the STT component while LLM is generating contends on the same mutex. In the voice pipeline (STT→LLM→TTS), this means each stage can't overlap at all — they're fully serialized through shared locks even when there's no actual data dependency.

The `destroy_loaded_model` path is worse — it does `cv.wait()` while holding `g_lifecycle_mutex`, blocking every other lifecycle query until the model's active refs drain to zero.

## Issue 3: Unbounded string growth during generation + voice sessions

**Files**: `llm_module.cpp:840`, `voice_agent_d7_abi.cpp`, `model_assignment.cpp:49`

Three separate unbounded growth patterns:

- `ctx->full_text += cleaned` — grows with every token, triggers multiple reallocations during a single 4K-token response (~16KB with doubling means 32KB peak)
- `handle->conversation_history` in voice agent — **never capped, never evicted**. A 30-min voice session accumulates all transcripts + responses forever
- `g_cached_models` / `g_cached_model_proto_bytes` — static vectors whose capacity never shrinks even after the cache is cleared

These aren't leaks in the traditional sense but they mean the process RSS only ever goes up during a session.

## Issue 4: Static stream registry maps hold dead sessions and never release bucket memory

**Files**: `rac_stt_stream.cpp:162,169,180`, `rac_vad_stream.cpp:90,95`, `rac_tts_stream.cpp:52,57`, `rac_diarization_stream.cpp`, `rac_diffusion_stream.cpp`

Each modality has 2-3 `static unordered_map` instances tracking active stream sessions:
```cpp
static std::unordered_map<uint64_t, StreamSession> m;
```

Problems:
- If a component is destroyed without proper quiesce (crash, Android activity kill, lifecycle bug), entries stay forever
- Even after proper erasure, `unordered_map` never releases its bucket array — the internal hash table only grows over the process lifetime
- Each `StreamSession` holds audio frame buffers, serialization state, callback pointers

On a device running voice all day (dictation app, assistant), this is a slow steady leak of 1-5KB per session that never comes back.

## Issue 5: Every registered model stored twice — C struct + full proto serialization

**Files**: `model_registry.cpp`, `model_registry_internal.h`

The model registry keeps two complete copies of every model's metadata:
1. A `rac_model_info_t*` C struct where each field is individually `rac_strdup`'d (8-12 malloc calls per model)
2. A `std::string` holding the full serialized protobuf bytes (200-500 bytes per model)

With a typical 30-50 model catalog that's ~50KB of pure duplication. The modern SDK paths already use the proto-byte ABI exclusively — the C struct exists only for legacy compatibility.

## Approximate Impact If All Fixed

Based on the hot-path analysis:

- **Token generation overhead**: drops from ~5 mutex acquisitions + 2 proto serializations per token to effectively zero overhead on the critical path. On a mid-range Android phone this could mean 5-15% higher effective tok/s since the CPU isn't spending cycles on plumbing.
- **Voice pipeline latency**: STT→LLM→TTS stages could overlap instead of fully serializing. Time-to-first-audio in voice could drop by the STT post-processing time (~50-150ms) since LLM can start while STT cleanup runs.
- **Steady-state RAM (30-min voice session)**: estimated 150-300KB savings from capping conversation history + fixing stream registry leaks + eliminating double model storage. On a 4GB device running a 2GB model, that's not huge in absolute terms but it's the difference between staying comfortably in budget vs tickling the Android low-memory killer.
- **Peak RAM during generation**: pre-allocating `full_text` and eliminating per-token proto allocations removes the sawtooth allocation pattern. Fewer GC pressure spikes on Android, fewer jetsam events on iOS.

None of these are individually catastrophic, but they compound — especially in the voice pipeline path where all 5 hit simultaneously on every turn.

Adityakk9031:
## Summary
Fixes performance bottlenecks and memory leaks identified in issue #647.

### Key Fixes
1. **VLM Streaming Hot Path Optimization (\lm_module.cpp\)**:
   - Added \	hread_local\ reuse for \VLMStreamEvent\ proto and scratch byte vector in \dispatch_vlm_stream_event()\, eliminating ~6 heap allocations per token at steady state.
   - Eliminated O(N^2) streaming text re-serialization on intermediate tokens in \generated_stream_token_trampoline()\ by sending delta tokens instead of full accumulated text on intermediate events. Full accumulated text is dispatched upon completion.
   - Added \ctx.text.reserve(2048)\ before starting generation to prevent string reallocations.

2. **VAD Stream Registry Memory Leak Fix (\
ac_vad_stream.h\, \
ac_vad_stream.cpp\, \ad_module.cpp\)**:
   - Declared and implemented \
ac_vad_stream_component_teardown(handle)\ to sweep and erase orphaned sessions from \g_sessions()\ when a VAD component is destroyed.
   - Wired teardown into \
ac_vad_component_destroy()\ in \ad_module.cpp\, matching STT/TTS cleanup patterns.

3. **Assignment Cache Memory Reclamation (\model_assignment.cpp\)**:
   - Added \.shrink_to_fit()\ on \g_cached_models\ and \g_cached_model_proto_bytes\ inside \clear_cache_internal()\.

Fixes #647

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved voice activity detection component cleanup to prevent lingering streaming sessions and callbacks.
  * Enhanced shutdown reliability by ensuring active stream dispatch is fully quiesced before release.

* **Improvements**
  * Streamlined vision-language streaming events for more consistent terminal status reporting.
  * Reduced redundant streaming payload data and improved handling of text accumulated during streaming.

### DIFF
--- a/sdk/runanywhere-commons/include/rac/features/vad/rac_vad_stream.h
+++ b/sdk/runanywhere-commons/include/rac/features/vad/rac_vad_stream.h
@@ -142,6 +142,11 @@ RAC_API rac_result_t rac_vad_stream_stop_proto(uint64_t session_id);
  */
 RAC_API rac_result_t rac_vad_stream_cancel_proto(uint64_t session_id);
 
+/**
+ * @brief Teardown all VAD streaming sessions for a handle.
+ */
+RAC_API void rac_vad_stream_component_teardown(rac_handle_t handle);
+
 #ifdef __cplusplus
 } /* extern "C" */
 #endif
--- a/sdk/runanywhere-commons/include/rac/features/vad/rac_vad_stream.h
+++ b/sdk/runanywhere-commons/include/rac/features/vad/rac_vad_stream.h
@@ -143,9 +143,22 @@ RAC_API rac_result_t rac_vad_stream_stop_proto(uint64_t session_id);
 RAC_API rac_result_t rac_vad_stream_cancel_proto(uint64_t session_id);
 
 /**
- * @brief Teardown all VAD streaming sessions for a handle.
+ * @brief Clear the stream callback slot AND erase every streaming session
+ *        still bound to @p handle.
+ *
+ * Commons-internal: called only from rac_vad_component_destroy() in
+ * vad_module.cpp. Deliberately NOT decorated with RAC_API — it is not part of
+ * the exported SDK ABI (see sdk/runanywhere-commons/exports/RACommons.exports
+ * and scripts/validation/commons/check_rac_api_exports.sh), and SDK bridges
+ * must keep using rac_vad_unset_stream_proto_callback() +
+ * rac_vad_proto_quiesce() for their own teardown.
+ *
+ * Superset of rac_vad_unset_stream_proto_callback(handle): sessions that were
+ * never stopped/cancelled would otherwise outlive the component in the session
+ * registry, holding a dangling handle that a recycled component address makes
+ * indistinguishable from a live session.
  */
-RAC_API void rac_vad_stream_component_teardown(rac_handle_t handle);
+void rac_vad_stream_component_teardown(rac_handle_t handle);
 
 #ifdef __cplusplus
 } /* extern "C" */
--- a/sdk/runanywhere-commons/src/features/vad/rac_vad_stream.cpp
+++ b/sdk/runanywhere-commons/src/features/vad/rac_vad_stream.cpp
@@ -277,7 +277,7 @@ void rac_vad_stream_component_teardown(rac_handle_t handle) {
         return;
     std::lock_guard<std::mutex> lock(g_mu());
     g_slots().erase(handle);
-    for (auto it = g_sessions().begin(); it != g_sessions().end(); ) {
+    for (auto it = g_sessions().begin(); it != g_sessions().end();) {
         if (it->second.handle == handle) {
             it = g_sessions().erase(it);
         } else {
--- a/sdk/runanywhere-commons/src/features/vad/rac_vad_stream.cpp
+++ b/sdk/runanywhere-commons/src/features/vad/rac_vad_stream.cpp
@@ -272,6 +272,20 @@ rac_result_t rac_vad_unset_stream_proto_callback(rac_handle_t handle) {
     return RAC_SUCCESS;
 }
 
+void rac_vad_stream_component_teardown(rac_handle_t handle) {
+    if (handle == nullptr)
+        return;
+    std::lock_guard<std::mutex> lock(g_mu());
+    g_slots().erase(handle);
+    for (auto it = g_sessions().begin(); it != g_sessions().end(); ) {
+        if (it->second.handle == handle) {
+            it = g_sessions().erase(it);
+        } else {
+            ++it;
+        }
+    }
+}
+
 // Public quiesce helper. Mirrors
 // rac_vlm_proto_quiesce / rac_llm_proto_quiesce. Spin-waits until every
 // in-flight dispatch_vad_stream_event invocation has returned. Callers
--- a/sdk/runanywhere-commons/src/features/vad/vad_module.cpp
+++ b/sdk/runanywhere-commons/src/features/vad/vad_module.cpp
@@ -699,7 +699,7 @@ extern "C" void rac_vad_component_destroy(rac_handle_t handle) {
     // in addition to the proto_activity_slot cleared above. Without this, the
     // wire-seq + stale user_data UAF triggers when the handle heap address is
     // reused by a fresh component (rac_vad_component_create).
-    rac_vad_unset_stream_proto_callback(handle);
+    rac_vad_stream_component_teardown(handle);
     // Spin-wait for any in-flight
     // dispatch_vad_stream_event() invocation on another thread before freeing
     // the component. Mirrors rac_vlm_component_destroy / rac_llm_component_destroy.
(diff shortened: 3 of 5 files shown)


Adityakk9031:
@sanchitmonga22 have a look

josuediazflores:
Nice work turning the #647 findings around quickly. I reviewed the diff against main (`89de67b`) and the core mechanics all check out:

- The `thread_local` proto + scratch vector in `dispatch_vlm_stream_event` matches the established LLM/STT pattern (`rac_llm_stream.cpp:452`, `rac_stt_stream.cpp:1688-1689`), and `ctx.text.reserve(2048)` mirrors the LLM path.
- The O(N^2) fix is right, and I verified it is safe: grepping every SDK's non-generated source plus all example apps finds zero consumers of `streaming_text`, so dropping it from per-token events breaks nothing in-tree.
- The VAD teardown is correctly placed: it erases the callback slot and all handle-owned sessions under one lock, before `rac_vad_proto_quiesce()`, exactly where the old unset call sat, so the quiesce ordering that prevents the UAF is preserved. This also fixes a subtle correctness issue beyond the leak: the handle-wide fallback scan in `dispatch_vad_stream_event` (`rac_vad_stream.cpp:410`) could stamp a stale orphaned session's `request_id` onto a new component reusing the same heap address. With sessions swept on destroy, that path is closed.

Two non-blocking suggestions:

1. **Symbol linkage.** STT solves this same problem with internal `rac::stt::begin/end_stream_component_teardown` rather than a public C symbol. `rac_vad_stream_component_teardown` is declared `RAC_API` in the public header but is not in `exports/RACommons.exports` (the curated list CMake wires into Apple builds), and nothing outside the library calls it. Making it an internal `rac::vad::` function would match the STT pattern and avoid a public declaration whose symbol is not exported; alternatively, add it to the exports file.
2. **`streaming_text` on the terminal event.** The new COMPLETED/FAILED `SDKEvent` ships the entire accumulated text once per generation. That is already a big improvement over per-token, but since nothing reads the field and LLM's `emit_llm_generation_completed` sends metrics only (tokens, durations, tok/s, no text), dropping it here too would make the VLM completion shape consistent with LLM and keep generated content out of the telemetry/log sinks entirely.

Tiny nits: there is a trailing-whitespace-only line after `ctx->terminal_sent = true;`, and the new `populate_envelope(...)` ternary line probably exceeds the commons `.clang-format` column limit.

sanchitmonga22 (approved):
Two real problems found here, and the VAD one is the more serious of the pair. Thank you.

The orphan session leak matters because it does not just leak. Recycled handles cross-attribute request ids, so telemetry lands against the wrong session. That is the kind of bug that makes later debugging actively misleading. The O(N-squared) in the VLM streaming hot path is also real.

I pushed three fixes on top before merging:

1. Removed `RAC_API` from `rac_vad_stream_component_teardown`. Adding it broke the symbol-export validation gate.
2. Dropped the new terminal `SDKEvent` emission, which double-counted telemetry against the existing terminal event.
3. Dropped an unrelated `shrink_to_fit` hunk to keep the diff to the fix.

One behavior change a reader should know about: VLM per-token `GenerationEvent.streaming_text` on the public SDKEvent bus is now always empty. Nothing in this repo reads it and the terminal event still carries the full text, but it is a public bus, so I am recording it here rather than letting it pass unnoticed.

`electron (windows-2022)` and `swift-spm` are red on main too and unrelated.

Issue #635, opened 2026-08-07
tokens_per_second fallback includes prefill in Swift and Web, understating generation speed


ayaangazali:
`tokens_per_second` is computed in each SDK when the backend does not return one. commons excludes prefill from the denominator and states why in `sdk/runanywhere-commons/src/features/llm/llm_module.cpp`:

```cpp
// Tokens/sec over decode time only — including prefill (TTFT) in the
// denominator systematically understates generation speed.
```

Python (`stream_metrics.py`) and Electron (`stream.ts`) follow that. Three SDK fallbacks do not, so the same run reports a lower rate on those platforms than on native. With 100 tokens, 1s prefill and a 5s span: 25 tok/s from commons, 20 tok/s from the drifted paths. The gap widens with prompt length.

Kotlin is fixed in #634. The remaining three:

**1. `sdk/runanywhere-swift/Sources/RunAnywhere/Public/Extensions/LLM/RunAnywhere+TextGeneration.swift:253`**

```swift
result.tokensPerSecond = final.map { Double($0.tokensPerSecond) }
    ?? (totalLatency > 0 ? Double(tokenCount) / (totalLatency / 1000) : 0)
```

The local `ttft` is in scope, used two lines below for `result.ttftMs`. One-line fix.

**2. `sdk/runanywhere-web/packages/core/src/Public/Extensions/RunAnywhere+TextGeneration.ts:456`** (`aggregateStream`)

```ts
tokensPerSecond: result.tokensPerSecond
  || (totalLatencyMs > 0 ? tokenCount / (totalLatencyMs / 1000) : 0),
```

`ttftMs` is in scope on the next line. One-line fix.

**3. `sdk/runanywhere-web/packages/core/src/Public/Extensions/RunAnywhere+TextGeneration.ts:291`** (`finalLLMResult`) — not a one-liner

```ts
const generationTimeMs = final?.totalTimeMs ?? performance.now() - startedAt;
...
tokensPerSecond: final?.tokensPerSecond
  ?? (generationTimeMs > 0 ? (tokensGenerated / generationTimeMs) * 1000 : 0),
```

When the fallback runs, `final` is undefined, so `final?.timeToFirstTokenMs` is undefined too and the function never receives a first-token timestamp. Correcting this one needs a timestamp threaded in from the caller, so it is a different change from the other two.

Happy to send the two one-liners, and the `finalLLMResult` plumbing separately, if that split works for you.

[...]

PR #661, opened 2026-08-11
commons owned metrics

sanchitmonga22:
## Summary

Move cross-platform model-output calculations and policy into `runanywhere-commons`, then delete the SDK and example implementations that used to recompute or guess the same values.

This PR also bumps the plugin ABI to v9 so streaming engines can report real token deltas and terminal token totals. SDKs now preserve typed commons values. If commons does not provide a measurement, the SDK leaves it unknown instead of fabricating a fallback.

### C++ commons

- Fixed streaming metric ownership and arithmetic:
  - real prompt/completion counts through `tokens_in_delta` and `get_stream_token_counts`
  - separate TTFT, first-content latency, prefill, decode time, and decode throughput
  - explicit `batch_buffered` and `counts_estimated` provenance
- Added shared audio primitives for PCM16/float conversion, WAV, RMS/dB level, resampling, and duration.
- Moved VAD onset/offset, debounce, hangover, and pre-roll into the commons stream state machine.
- Added shared embedding norm/cosine functions and authoritative rerank ordering.
- Added device facts, model compatibility, download progress, and registry timestamp ownership.
- Added JSON Schema to GBNF compilation, structured-output repair, finish-reason propagation, segmentation coverage, and LoRA scale resolution.
- Added QHexRT reasoning-prefill detection so models whose templates pre-open `<think>` stream correctly.

### Plugin ABI and engines

- Bumped `RAC_PLUGIN_API_VERSION` to `9u`.
- Widened LLM stream callbacks with `tokens_in_delta`.
- Added `get_stream_token_counts` to `rac_llm_service_ops_t`.
- Updated the four LLM publishers: llama.cpp, QHexRT, MLX, and NeuRT.
- Kept cloud, ONNX, and Sherpa as non-LLM/null publishers.

### SDKs and examples

- Added thin JNI, Swift C ABI, Dart FFI, Nitro, N-API/RPC, WebAssembly, and Python bindings for the new commons APIs.
- Deleted local metric synthesis, chars-per-token estimates, thinking parsers, schema compilers, vector math, audio DSP, VAD endpointing, device tiers, model-fit rules, registry clocks, finish-reason heuristics, and LoRA default coercion.
- Removed compatibility shims such as `aggregateStream`, `streamWithMetrics`, and `synthesizeStreamResult` instead of keeping duplicate behavior.
- Example apps now render typed SDK results and keep only UI formatting, platform I/O, and explicit test-input generation.
[...]

PR #662, opened 2026-08-11
fix(web): measure decodeTokensPerSecond over decode time, not the whole span

ayaangazali:
Part of #635.

## What is wrong

`aggregateStream`'s throughput fallback divides the token count by the whole request span:

```ts
const generationTimeMs = result.generationTimeMs || totalLatencyMs;
const decodeMs = generationTimeMs - (result.promptEvalTimeMs ?? 0);
...
decodeTokensPerSecond: (result.usage?.decodeTokensPerSecond ?? 0)
  || (totalLatencyMs > 0 ? tokenCount / (totalLatencyMs / 1000) : 0),
```

`totalLatencyMs` starts at the request, so prefill sits inside the denominator of a field named `decodeTokensPerSecond`.

## Why that is wrong

commons excludes prefill deliberately and says why in `llm_module.cpp`:

```cpp
```

So the same run reports a lower rate on this path than commons reports natively. With 100 tokens, 1s prefill and a 5s span that is 20 tok/s here against 25 from commons, and the gap widens with prompt length. It is silent, because both write the same field.

The value needed is already there. `decodeMs` is computed two lines above and is trusted enough to be reported as `decodeTimeMs` in the same object, and the sibling path earlier in this file already does it correctly:

```ts
decodeTokensPerSecond: final?.usage?.decodeTokensPerSecond
  ?? (decodeMs > 0 ? (outputTokens / decodeMs) * 1000 : 0),
```

Only this one fallback still uses the full span.

## What this does

Uses `decodeMs` for the rate, matching the sibling path and commons. One line, no new state, and the guard shape is unchanged so a zero or negative window still yields 0 rather than a divide-by-zero.

Kotlin had the same defect and was fixed in #634.

## Scope

#635 lists three sites. This is the one that is a clean one-liner with the correct value already in scope. The second Web site it mentions has since been fixed independently, and the Swift one has been rewritten around a batch-buffered heuristic that deserves its own discussion rather than being folded in here.

## Testing

I could not run the web unit suite on this machine: `vitest` fails to load with `Cannot find module '@bufbuild/protobuf/wire'` after a clean `npm install`, and it fails identically with my diff stashed, so it is pre-existing rather than from this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streaming text-generation performance reporting by calculating fallback decode throughput from decode time rather than total request latency.

### DIFF
--- a/sdk/runanywhere-web/packages/core/src/Public/Extensions/RunAnywhere+TextGeneration.ts
+++ b/sdk/runanywhere-web/packages/core/src/Public/Extensions/RunAnywhere+TextGeneration.ts
@@ -429,7 +429,7 @@ export async function aggregateStream(
       outputTokens,
       totalTokens: (result.usage?.totalTokens ?? 0) || inputTokens + outputTokens,
       decodeTokensPerSecond: (result.usage?.decodeTokensPerSecond ?? 0)
-        || (totalLatencyMs > 0 ? tokenCount / (totalLatencyMs / 1000) : 0),
+        || (decodeMs > 0 ? (tokenCount / decodeMs) * 1000 : 0),
       prefillMs: result.usage?.prefillMs ?? 0,
       ttftMs: result.usage?.ttftMs ?? (ttftMs ?? 0),
     },


sanchitmonga22:
Same story as #663, and same apology — overtaken by about 75 minutes.

The analysis was right: measuring `decodeTokensPerSecond` across the whole span rather than decode time understates throughput. #661 (`7cc9212af`) deleted `aggregateStream` (156 lines) from the Web SDK and moved the computation into C++ commons so all bindings share one implementation. There's no longer a call site for this patch to land on.

Your earlier Kotlin PR #634 **did** land and is real work — thank you for that one. (Fair warning: #661 later deleted the code #634 touched too, for the same reason. That's a process failure on our side, not a reflection on the contribution.)

The root cause is issue #635, which stayed open pointing at code that had already moved. I'm closing it so this stops happening to people.

If you'd like the equivalent fix in its new home, `core/src/features/llm/streaming_metrics.cpp` is where the derivation lives now — one fix there covers every binding.

PR #663, opened 2026-08-11
fix(swift): rate decodeTokensPerSecond over the decode window

ayaangazali:
Part of #635. Companion to #662 (Web) and #634 (Kotlin, merged).

## What is wrong

`applyTimingMetrics` computes the decode window, then never uses it as a denominator:

```swift
let wallTps = generationTimeMs > 0 && outputTokens > 0
    ? Double(outputTokens) / (generationTimeMs / 1000.0) : 0
let decodeWindow = reportedTtft.map { generationTimeMs - Double($0) } ?? generationTimeMs
let batchBuffered =
    reportedTtft != nil && decodeWindow < max(50.0, generationTimeMs * 0.05)
if batchBuffered {
    ...
} else {
    result.usage.decodeTokensPerSecond = reportedTps > 0 ? reportedTps : wallTps
```

`decodeWindow` exists only to detect the batch-buffered case. In the ordinary branch, where TTFT is trustworthy and the window is real, the fallback is still `wallTps`, which is measured across the whole span.

## Why that is wrong

`generationTimeMs` starts at the request, so prefill is inside the denominator of a field named `decodeTokensPerSecond`. commons excludes it deliberately:

```cpp
```

So the same run reports a lower rate on Swift than commons reports natively, silently, and the gap widens with prompt length. Kotlin had exactly this and was fixed in #634; the Web equivalent is #662.

## What this does

Rates the fallback over `decodeWindow`, the value the function already computes.

Deliberately narrow:

- The `batchBuffered` branch is untouched. Its `wallTps` is correct, and the comment above the function explains why: for Maple/Bonsai the tokens arrive in one dump, `decode = total - ttft` collapses to a few ms, and rating over it would produce an absurd tok/s and a fake TTFT.
- When TTFT is unknown, `decodeWindow` is already `generationTimeMs`, so the new expression reduces to exactly the old `wallTps`. That path is unchanged by construction.
- `reportedTps` still wins whenever the backend supplies a rate.

## Testing

`swiftc -parse` on the file is clean.

SwiftLint is not runnable on this machine, which has Command Line Tools but no full Xcode, so `swiftlint` aborts loading `sourcekitdInProc.framework`. I am relying on the `swift-spm` CI job for lint rather than implying I ran it.

Worth flagging separately: `swift-spm` is currently failing on `main` itself, at link time with `Undefined symbols for architecture arm64` for `rac::desktop::install_device_manager_provider()` and the three `rac_desktop_*` symbols. That is unrelated to this diff, which touches one Swift expression.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reported text-generation throughput when backend metrics are unavailable.
  - Decode performance is now calculated over the appropriate generation window, including a fallback when time-to-first-token data is unavailable.

### DIFF
--- a/sdk/runanywhere-swift/Sources/RunAnywhere/Public/Extensions/LLM/RunAnywhere+TextGeneration.swift
+++ b/sdk/runanywhere-swift/Sources/RunAnywhere/Public/Extensions/LLM/RunAnywhere+TextGeneration.swift
@@ -285,6 +285,11 @@ private extension RunAnywhere {
         let wallTps = generationTimeMs > 0 && outputTokens > 0
             ? Double(outputTokens) / (generationTimeMs / 1000.0) : 0
         let decodeWindow = reportedTtft.map { generationTimeMs - Double($0) } ?? generationTimeMs
+        // Rate over the decode window, matching commons' llm_module.cpp. Falls
+        // back to the full span when TTFT is unknown, in which case
+        // `decodeWindow` already IS the full span, so this is wallTps.
+        let decodeTps = decodeWindow > 0 && outputTokens > 0
+            ? Double(outputTokens) / (decodeWindow / 1000.0) : wallTps
         let batchBuffered =
             reportedTtft != nil && decodeWindow < max(50.0, generationTimeMs * 0.05)
         if batchBuffered {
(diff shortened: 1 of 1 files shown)


sanchitmonga22:
Thanks for this — and I'm sorry, this one got overtaken by about 74 minutes.

Your diagnosis was **correct**. `decodeTokensPerSecond` really was being rated over the whole span including TTFT. It was independently confirmed by the commons migration that landed in #661 (`7cc9212af`, merged 18:07Z), which deleted `applyTimingMetrics` and 167 lines from `RunAnywhere+TextGeneration.swift` wholesale and moved the calculation into C++ so every binding shares it. A rebase can't produce a meaningful patch here — every context line your diff anchors on is gone.

That's on us, not you: issue #635 was still pointing contributors at this code after it had moved. I'm closing that issue too.

**If you want the same fix where it now counts:** `core/src/features/llm/streaming_metrics.cpp:228-235` still derives the rate as `output_tokens / (latency_ms - ttft_ms)` and doesn't route through `compute_stream_timing_from_totals`. Same bug, one layer down — and fixing it there helps Swift, Kotlin, Web, Flutter, RN and Electron at once. I'd merge that happily.

Also: your unrelated `swift-spm` link-failure report was accurate, and it's fixed in `Package.swift:491-503`.

PR #668, opened 2026-08-12
fix(commons): answer loaded-state from the lifecycle store before the host

ayaangazali:
Follows up #615, which you closed on the premise. You were right that the resolved-id fallback was unreachable, and you pointed at the real bug underneath it:

> `is_model_loaded` is not wired to any lifecycle store — the JNI stub above literally lies to commons. Wiring that up for real, and aligning `storage_analyzer.cpp:1209` (the site that actually *deletes files* based on loaded state), is a genuinely valuable fix.

This is the read side of that.

## What is wrong

`known_loaded_state` can only answer through the host callback:

```cpp
if (!handle || !model_id || !out_is_loaded || !handle->callbacks.is_model_loaded) {
    return false;
}
```

No host answers it correctly:

- JNI hardcodes the answer (`core/src/jni/runanywhere_commons_jni.cpp:4083`):
  ```cpp
  static rac_result_t jni_storage_is_model_loaded(const char* /*model_id*/, rac_bool_t* out_is_loaded, void*) {
      if (out_is_loaded != nullptr) { *out_is_loaded = RAC_FALSE; }
      return RAC_SUCCESS;
  }
  ```
- Swift and Electron leave the slot NULL, so the guard above returns "unknown".

So the delete path believes nothing is ever loaded. `unload_if_loaded` never fires, and the recursive delete runs underneath a model the engine still has open.

## Why the fix belongs here

Commons is the component that performed the load, and it already tracks it: `g_loaded` maps each component to a `LoadedModel`, which carries `model_id`. It has been able to answer this question all along and was never asked.

Fixing it here fixes Android, Swift and Electron at once rather than asking three hosts to each implement a callback they are not well placed to answer, which is also what the layering rule in AGENTS.md asks for.

## What this does

Adds `rac::lifecycle::is_model_loaded(const char*)` on the existing internal lifecycle bridge (`core/src/features/rac_nonllm_lifecycle_bridge.h`), implemented beside the other accessors, reading `g_loaded` under `g_lifecycle_mutex`, and has `known_loaded_state` consult it before the callback.

Kept off the public C ABI: the only caller is inside commons, so a new `RAC_API` export would widen the surface every platform carries for no reason. `core/include` is untouched by this PR.

(Edited after the v2 rework. The first revision added a public `RAC_API rac_model_lifecycle_is_model_loaded`; the `RAC_API export drift check (strict)` gate correctly flagged it, and the merged shape is the internal one described above.)

**Only the positive answer is treated as final.** `RAC_FALSE` means commons did not load it, not that nobody did, so that path still defers to the host callback exactly as before. The consequence worth stating plainly: this can only ever *prevent* a delete, never permit one that was previously blocked. That asymmetry is documented on the declaration and at the call site.

I left the JNI stub in place rather than deleting it. With the lifecycle consulted first its wrong answer is no longer reachable for commons-loaded models, and removing it is a separate question about whether the callback should stay in the struct at all.

## Scope

Read side only, deliberately. `unload_model` has the same hole, but actually unloading from inside the delete path means calling into the lifecycle unload, which is a heavier change with its own reentrancy questions. I would rather land the read side so the path stops lying, then do the unload wiring separately. Say the word if you would rather see them together.

## Testing

Built and run locally, macOS, `-DRAC_BUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Debug`:

- `test_storage_analyzer_proto` — exit 0, `0 test(s) failed`, including `test_delete_plan_blocks_loaded_missing_and_missing_path`, which is the loaded-state delete guard.
- `test_core --run-all` — exit 0, `18 passed, 0 failed, 18 total`.

Both are targets that exist in `core/tests/CMakeLists.txt` and I watched both run. On #615 I cited `test_storage_service_proto_abi`, which does not exist in this repo; that was my error and I have fixed how I verify target names.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of whether models are loaded across supported components.
  * Storage analysis now recognizes lifecycle-managed model loads even when platform callbacks are unavailable.
  * Model load status checks now handle invalid or missing model identifiers more reliably.

### DIFF
--- a/core/src/core/model_lifecycle_accessors.cpp
+++ b/core/src/core/model_lifecycle_accessors.cpp
@@ -467,3 +467,37 @@ void release_lifecycle_segmentation(LifecycleSegmentationRef* ref) {
 }
 
 }  // namespace rac::lifecycle
+
+// ---------------------------------------------------------------------------
+// Loaded-state query
+//
+// The storage delete path needs to know whether a model is open before it
+// removes files underneath it. Hosts answer that through
+// rac_storage_callbacks_t::is_model_loaded, but commons is the component that
+// performed the load and already tracks it in g_loaded, so it can answer
+// without asking anyone. Internal on purpose: the only caller is inside
+// commons, so this stays off the public C ABI.
+// ---------------------------------------------------------------------------
+
+namespace rac::lifecycle {
+
+bool is_model_loaded(const char* model_id) {
+#if defined(RAC_HAVE_PROTOBUF)
+    if (model_id == nullptr || *model_id == '\0') {
+        return false;
+    }
+    std::lock_guard<std::mutex> lock(rac::core::model_lifecycle::detail::g_lifecycle_mutex);
+    for (const auto& entry : rac::core::model_lifecycle::detail::g_loaded) {
+        const auto& loaded = entry.second;
+        if (loaded && loaded->model_id == model_id) {
+            return true;
+        }
+    }
+    return false;
+#else
+    (void)model_id;
+    return false;
+#endif
+}
+
+}  // namespace rac::lifecycle
--- a/core/src/features/rac_nonllm_lifecycle_bridge.h
+++ b/core/src/features/rac_nonllm_lifecycle_bridge.h
@@ -89,6 +89,19 @@ void release_lifecycle_diarization(LifecycleDiarizationRef* ref);
 rac_result_t acquire_lifecycle_segmentation(LifecycleSegmentationRef* out_ref);
 void release_lifecycle_segmentation(LifecycleSegmentationRef* ref);
 
+
+/**
+ * @brief Whether commons currently holds @p model_id loaded on any component.
+ *
+ * Answers from the lifecycle store, so it is true wherever commons performed
+ * the load regardless of what a host reports through
+ * rac_storage_callbacks_t::is_model_loaded. Callers that must not act while a
+ * model is open should treat `true` as authoritative; `false` only means
+ * commons did not load it, and a host that loads outside commons may still
+ * have it open.
+ */
+bool is_model_loaded(const char* model_id);
+
 }  // namespace rac::lifecycle
 
 #endif  // RAC_FEATURES_RAC_NONLLM_LIFECYCLE_BRIDGE_H
--- a/core/src/infrastructure/storage/storage_analyzer.cpp
+++ b/core/src/infrastructure/storage/storage_analyzer.cpp
@@ -32,6 +32,7 @@
 #include <unordered_set>
 #include <vector>
 
+#include "features/rac_nonllm_lifecycle_bridge.h"
 #include "rac/core/rac_logger.h"
 #include "rac/foundation/rac_proto_adapters.h"
 #include "rac/infrastructure/download/rac_download_orchestrator.h"
@@ -161,7 +162,23 @@ int64_t model_metric_size(rac_storage_analyzer_handle_t handle, const rac_model_
 
 bool known_loaded_state(rac_storage_analyzer_handle_t handle, const char* model_id,
                         bool* out_is_loaded) {
-    if (!handle || !model_id || !out_is_loaded || !handle->callbacks.is_model_loaded) {
+    if (!handle || !model_id || !out_is_loaded) {
+        return false;
+    }
+    // Ask commons before asking the host. The lifecycle store is what actually
+    // performed the load, whereas the callback is unanswerable for several
+    // hosts: the JNI one reports RAC_FALSE unconditionally, and Swift and
+    // Electron leave the slot NULL, which used to make every model look
+    // unloaded and let the delete path run underneath an open model.
+    //
+    // Only the positive answer is taken as final. A false answer means commons
+    // did not load it, not that nobody did, so that case still defers to the
+    // host. This can therefore only prevent a delete, never permit a new one.
+    if (rac::lifecycle::is_model_loaded(model_id)) {
+        *out_is_loaded = true;
+        return true;
+    }
+    if (!handle->callbacks.is_model_loaded) {
         return false;
     }
     rac_bool_t is_loaded = RAC_FALSE;


ayaangazali:
Reworked after CI caught a real problem with the first revision, and rebased onto current `main`.

The `validation` job failed on `RAC_API export drift check (strict)`, which was correct: I added `rac_model_lifecycle_is_model_loaded` as a new `RAC_API` symbol without adding it to `core/exports/RACommons.exports`.

Adding it to the export list would have made the check pass, but the check was pointing at a design mistake rather than a missing line. The only caller is `storage_analyzer.cpp`, which lives inside commons, so nothing needed a public C ABI symbol. The first revision widened the exported surface that every platform has to carry, and picked up the wasm and Windows symbol gates on the way.

So the second revision keeps it internal: `rac::lifecycle::is_model_loaded` declared on the existing `features/rac_nonllm_lifecycle_bridge.h` seam, which `model_lifecycle_accessors.cpp` already implements against and other feature TUs already consume. `core/include` is now untouched by this PR.

Behaviour is unchanged from the description above, including the asymmetry: only a positive answer is authoritative, so this can only prevent a delete, never permit one.

Re-verified after the rework and the rebase, on macOS with `-DRAC_BUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Debug`:

- `test_storage_analyzer_proto` — exit 0, `0 test(s) failed`

sanchitmonga22:
Thanks for this one, and apologies for how it was handled: this PR went in through an admin merge without the normal review pass. That was our process failure, not yours. This comment is the review it should have had.

Credit first, because it is deserved. You found the real bug under #615 and fixed it at the right layer. `known_loaded_state` could only answer through `rac_storage_callbacks_t::is_model_loaded`, and no host answers it: the JNI callback is a hardcoded `RAC_FALSE` (`core/src/jni/runanywhere_commons_jni.cpp:4083`), Swift sets the slot to nil deliberately (`bindings/swift/Sources/RunAnywhere/Foundation/Bridge/Extensions/CppBridge+Storage.swift:60`), and the Electron addon leaves it NULL (`bindings/electron/native/download_bridge.cpp:265`). I verified all three. So the delete path genuinely believed nothing was ever loaded, everywhere. Reading `g_loaded` under `g_lifecycle_mutex` is the correct answer from the correct component, and treating only the positive answer as authoritative while still deferring to the host on false is the right asymmetry to pick.

The v2 rework was also the right call. The `RAC_API` export drift gate was pointing at a design mistake rather than a missing line, and moving the function onto the existing `rac::lifecycle` seam instead of widening the public C ABI is what should have happened. `core/include` untouched is the correct end state.

Two things are still outstanding on main today.

**1. CodeRabbit's `COMPONENT_LIFECYCLE_STATE_READY` comment is right, and is still unfixed.** A failed load does not leave the slot empty: `install_failed_entry_preserving` (`core/src/core/model_lifecycle.cpp:445`) installs a sentinel into `g_loaded[component]` with `state = COMPONENT_LIFECYCLE_STATE_ERROR` and `model_id = request.model_id()`, and nothing erases it until a later load or an explicit unload. `is_model_loaded` matches on `model_id` alone, so it answers true for a model that has no backend at all. Effect: a model whose load failed becomes undeletable. The plan drops it with "Model is loaded and cannot be safely deleted", and the execute path fails it with `RAC_ERROR_MODEL_NOT_LOADED`. That is exactly the model a user wants to delete after a bad download or an unsupported quant. The fix is one clause, `loaded->state == COMPONENT_LIFECYCLE_STATE_READY`, which is also what `acquire_lifecycle_llm` a few hundred lines up already does. We are fixing that forward in a follow-up PR. Nothing for you to do.

**2. Landing the read side alone shifts behaviour on iOS, worth naming.** `RunAnywhere.performDelete` sets `unloadIfLoaded = true`, but Swift leaves `unload_model` nil, so now that commons answers "loaded" the delete takes the "Model is loaded but no unload callback is set" branch and `models.delete(id:)` throws where it previously deleted. Failing loudly beats deleting under an open model, but it is a hard failure on a path that used to work. Android is the mirror image: `jni_storage_unload_model` returns `RAC_SUCCESS` without unloading anything, so the delete still runs underneath the resident model and the new protection there is only apparent. Electron happens to be safe because `assets.ts` releases a resident model before calling `deleteModel`. Your instinct to split read from unload was reasonable, and by the same layering argument you made in the description, the unload should come from commons too rather than from three host callbacks. That raises the priority of the second half rather than changing your reasoning.

CodeRabbit's other comment, the atomicity one, is a real race but not one you introduced. The same window existed with the host callback, and the reservation scheme it proposes is a much larger change than this PR should have carried. Fine to leave open.

One small note for next time: the description says the new symbol is `rac_model_lifecycle_is_model_loaded`, which was true of the first revision but not of what merged. Worth editing the body when a rework changes the shape, since the description is what people read first.

Careful, well argued work, and thank you for correcting the wrong test target name from #615 explicitly instead of quietly. That kind of honesty in a PR body is worth more than the diff.

ayaangazali:
Thank you for writing the review the PR should have had. Verifying all three host callbacks yourself, including the Swift slot being deliberately nil, is more than I could show from my side.

Taking the correction on the description first: you are right, and I have edited it. It described the v1 public `RAC_API` symbol, not the internal `rac::lifecycle::is_model_loaded` that actually merged, with a note saying why it changed. That is a fair hit, and the reason it matters is exactly what you said, the body is what people read first.

On the `COMPONENT_LIFECYCLE_STATE_ERROR` sentinel: that one is a genuine hole in my change and I did not see it. I checked presence in `g_loaded` and stopped there, without asking what could be in the map that is not actually loaded. `install_failed_entry_preserving` keeping a `model_id`-bearing sentinel means a failed load becomes undeletable, which is precisely the model a user is trying to delete. Understood that you are fixing it forward, so I will not touch it. The general lesson I am taking is the same one from #615: presence in a store is not the invariant, the state is.

On the unload half, I would like to take it if it is still free. Your point about iOS is the one that decides it. Landing the read side alone converts a silent wrong delete into a hard failure on a path that used to work, and Android's `RAC_SUCCESS`-without-unloading means the protection there is only apparent, so the second half is what makes the first half honest rather than merely louder. By the same layering argument, the unload belongs in commons rather than in three host callbacks.

Shape I would propose, so you can redirect before I write it: on the delete path, when the model is loaded and `unload_if_loaded` is set, call the commons lifecycle unload for the component holding it rather than the host callback, and keep the callback as the fallback for a host that loaded outside commons. The reentrancy question I want to get right is the delete path calling into lifecycle unload while holding nothing, versus `g_lifecycle_mutex` being taken inside, so I would want that ordering to be explicit rather than incidental.

Happy to wait until your `COMPONENT_LIFECYCLE_STATE_READY` fix lands first, since the unload path should sit on top of a correct loaded-state answer rather than race it. Say the word either way.

PR #669, opened 2026-08-12
test(qhexrt): drive catalog arch expectations off the policy table

ayaangazali:
The follow-up you asked for when closing #631:

> the arch expectations are three hand-copied `unordered_set` literals, which is why they drift. Deriving them from the policy table instead would make this class of mismatch impossible. Related — the `all.size() == model_count()` guard is structurally blind to a mask widening.

## What is wrong

`test_native_catalog_owns_arch_and_auth_policy` restated the whole catalog by hand:

```cpp
const std::unordered_set<std::string> v75 = { "lfm2_5_230m", ... };
const std::unordered_set<std::string> v79 = { ... };
const std::unordered_set<std::string> v81 = { ... };
```

Those sets are a second copy of policy that already lives in `kModelPolicies`, so they drift. `lfm2_5_2_6b` and the three `cosmos3_edge_*` rows were each granted archs the sets still denied, which is a correct catalog failing its own test.

The union guard could not catch it either:

```cpp
ASSERT_EQ(all.size(), rac_qhexrt_catalog_model_count());
```

A set union compared against a row count is blind to any change that keeps the size equal, an add plus a remove being the obvious one, and it says nothing about which archs a row claims.

## What this does

Reads the table through the existing test seam and derives expectations from it.

- `policy_row_at(index, &row)` walks real rows, so add, remove and rename all change what the loop reads. That replaces the count guard rather than keeping both.
- `policy_arch_bit(arch)` lets the loop compare `rac_qhexrt_catalog_model_supports_arch` against the mask the catalog actually stores, per arch, per row.
- Each row must claim at least one arch, since a row reachable on no arch can never register.

The auth gate is deliberately **not** derived. Putting a model behind a gated HF repo should be an intentional act, so the private-row count stays pinned at 1 and `kitten_nano_0_8_varlen` is still named explicitly. Deriving that one would have made the test agree with any future gating change silently.

Net effect is 105 lines removed against 81 added, most of the removal being the three literals.

## Is it still testing anything?

It is checking that the public lookups agree with the policy the catalog stores, which is the part that can actually break, rather than restating the policy. I verified the assertions are live rather than tautological by inverting the mask test in `rac_qhexrt_catalog_model_supports_arch`:

```
ASSERT FAILED: rac_qhexrt_catalog_model_supports_arch(row.id, arch) == expected ? RAC_TRUE : RAC_FALSE
  @ engines/qhexrt/tests/test_qhexrt_model_catalog.cpp:97
1 QHexRT catalog test(s) failed
```

and it goes back to green once the injected bug is reverted.

## Testing

Built and run locally on macOS with `-DRAC_BUILD_TESTS=ON -DRAC_BACKEND_QHEXRT=ON -DCMAKE_BUILD_TYPE=Debug`. QHexRT configures as the non-routable shell here, which is enough for a pure-policy catalog test.

`test_qhexrt_model_catalog` — exit 0, `All 9 QHexRT catalog tests passed`, both before and after the change, plus the injected-bug run above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model catalog validation across supported architectures.
  * Corrected checks for model identifiers, architecture availability, and authentication requirements.
  * Ensured catalog policies are consistently reflected in model lookup results.
  * Improved consistency when identifying models that require authentication.

* **Tests**
  * Expanded coverage to validate every catalog policy entry and its associated access requirements.
  * Added checks for architecture support and expected public model lookup behavior.

### DIFF
--- a/engines/qhexrt/qhexrt_model_catalog.cpp
+++ b/engines/qhexrt/qhexrt_model_catalog.cpp
@@ -312,6 +312,21 @@ rac_result_t pin_hf_ref_to_arch(const std::string& input, rac_qhexrt_hexagon_arc
 
 namespace rac::qhexrt::catalog {
 
+bool policy_row_at(size_t index, PolicyRow* out_row) {
+    if (out_row == nullptr || index >= std::size(kModelPolicies)) {
+        return false;
+    }
+    const ModelPolicy& policy = kModelPolicies[index];
+    // `id` is a string_view over a literal in kModelPolicies, so it is
+    // NUL-terminated and outlives the caller.
+    out_row->id = policy.id.data();
+    out_row->arch_mask = policy.arch_mask;
+    out_row->requires_hf_auth = policy.requires_hf_auth;
+    return true;
+}
+
+uint8_t policy_arch_bit(rac_qhexrt_hexagon_arch_t arch) { return arch_mask(arch); }
+
 rac_result_t register_for_arch_proto(const uint8_t* request_bytes, size_t request_size,
                                      rac_qhexrt_hexagon_arch_t detected_arch,
                                      rac_bool_t engine_available, rac_bool_t* out_registered,
--- a/engines/qhexrt/qhexrt_model_catalog_internal.h
+++ b/engines/qhexrt/qhexrt_model_catalog_internal.h
@@ -12,6 +12,26 @@ rac_result_t register_for_arch_proto(const uint8_t* request_bytes, size_t reques
                                      rac_bool_t engine_available, rac_bool_t* out_registered,
                                      rac_proto_buffer_t* out_model);
 
+/** One row of the arch/auth policy table, as the catalog actually stores it. */
+struct PolicyRow {
+    const char* id;
+    uint8_t arch_mask;
+    bool requires_hf_auth;
+};
+
+/**
+ * Test seam: read row @p index of the policy table, false when out of range.
+ *
+ * Exposed so tests can drive their expectations off the table itself. Copying
+ * the arch sets into the test by hand is what let them drift behind catalog
+ * rows (lfm2_5_2_6b, and the three cosmos3_edge_* rows, were all granted archs
+ * the test still denied).
+ */
+bool policy_row_at(size_t index, PolicyRow* out_row);
+
+/** The bit @p arch occupies in PolicyRow::arch_mask. */
+uint8_t policy_arch_bit(rac_qhexrt_hexagon_arch_t arch);
+
 }  // namespace rac::qhexrt::catalog
 
 #endif  // RAC_QHEXRT_MODEL_CATALOG_INTERNAL_H
--- a/engines/qhexrt/tests/test_qhexrt_model_catalog.cpp
+++ b/engines/qhexrt/tests/test_qhexrt_model_catalog.cpp
@@ -60,114 +60,55 @@ int test_supported_arch_policy() {
 }
 
 int test_native_catalog_owns_arch_and_auth_policy() {
-    const std::unordered_set<std::string> v75 = {
-        "lfm2_5_230m",          "lfm2_5_350m",         "qwen3_0_6b",      "qwen3_5_0_8b",
-        "qwen3_5_2b",           "ternary_bonsai_1_7b", "internvl3_5_1b",  "qwen3_vl",
-        "nemotron_ocr",         "nemotron_ocr_v1",     "nemotron_parse",  "whisper_base",
-        "whisper_small",        "moonshine_tiny",      "moonshine_base",  "parakeet_tdt_0_6b_v2",
-        "parakeet_tdt_0_6b_v3", "parakeet_rnnt_1_1b",  "canary_1b_flash", "nemotron_asr_streaming",
-        "melotts_en",           "kokoro_en",           "kitten_nano_0_8", "kitten_nano_0_8_varlen",
-        "embeddinggemma_300m", "lfm2_5_2_6b",
-        "nv_embedqa_1b",        "nv_rerankqa_1b",      "siglip2_base",
-        "nemotron_nano_8b",     "canary_180m_flash",
-        "parakeet_ctc_1_1b",    "nemotron_3_embed_1b", "magpie_tts_357m",
+    // Expectations come from the policy table itself rather than three
+    // hand-copied sets. Those sets are what drifted: lfm2_5_2_6b and the three
+    // cosmos3_edge_* rows were each granted archs the test still denied, so the
+    // test failed on a correct catalog. Reading the table makes that class of
+    // mismatch impossible while still checking that the public lookups agree
+    // with the policy the catalog actually stores.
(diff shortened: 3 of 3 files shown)


ayaangazali:
Rebased onto current `main` to re-run CI. The two red jobs were both a mirror outage rather than anything from this diff:

```
android-arm64   Download vendored Sherpa-ONNX (Android)
                curl: (22) The requested URL returned error: 503

linux-debug     error: downloading '.../runanywhere-build-mirrors/releases/download/bzip2-1.0.8/...'
                The requested URL returned error: 503
```

Both are downloads from `github.com/Siddhesh2377/runanywhere-build-mirrors` returning 503 during dependency fetch, before anything in `engines/qhexrt` is compiled.

Re-verified locally after the rebase, macOS, `-DRAC_BUILD_TESTS=ON -DRAC_BACKEND_QHEXRT=ON -DCMAKE_BUILD_TYPE=Debug`:

`test_qhexrt_model_catalog` — exit 0, `All 9 QHexRT catalog tests passed`.

sanchitmonga22:
Post-merge review. This landed via admin merge without the normal review pass, which is on us, not on you. Here is the review it should have had before merging.

Thank you for this one, and specifically for the part that is easy to get wrong: you did not derive everything. Deleting the three `unordered_set` literals and reading `kModelPolicies` through `policy_row_at` is right for the arch masks, and keeping the HF auth gate pinned (`private_row_count == 1` plus the named `kitten_nano_0_8_varlen` assertion) is the correct call. A derived auth check would agree with any future gating change silently, and you said exactly that in the description. `policy_arch_bit` delegating to the real `arch_mask()` rather than re-declaring the bit values is also the right shape for a test seam.

## What I verified

Configured with `-DRAC_BUILD_TESTS=ON -DRAC_BACKEND_QHEXRT=ON -DCMAKE_BUILD_TYPE=Debug` on macOS and ran `test_qhexrt_model_catalog` against main today: all 9 pass. I also compiled the pre-PR version of the test file (from base `554ea76`) against the same catalog and linked it: that also passes, all 9. The catalog table has not changed since your merge base, and `554ea76`'s test already listed `lfm2_5_2_6b` and the three `cosmos3_edge_*` rows in the arch sets.

So the "a correct catalog failing its own test" framing was true when you first hit it, but someone had already hand-patched the sets by the time this PR was opened. That does not weaken the change at all, it is the drift argument proving itself, but the description reads as if the test was red at merge time and it was not.

## One real coverage regression

The old `ASSERT_EQ(all.size(), rac_qhexrt_catalog_model_count())` was blind to a mask widening, as you said. It did catch one thing the new loop does not: a duplicated row id, since a set of distinct ids could not equal an inflated row count.

I appended a second `{"qwen3_0_6b", kV75V81, false}` to `kModelPolicies` and rebuilt:

- main today: `PASS native_catalog_owns_arch_and_auth_policy`, all 9 pass
- pre-PR test: `ASSERT FAILED: all.size() == rac_qhexrt_catalog_model_count()`

That matters because `find_model_policy` stops at the first match, so a duplicate id makes the second row dead policy that nothing reads. Fix is one line in the loop, an `unordered_set<string> seen_ids` and `ASSERT_TRUE(seen_ids.insert(row.id).second)`. I built and ran that: green on the real table, fails on the injected duplicate. Being fixed forward in a follow-up PR, not something for you to chase.

## CodeRabbit's comment is still outstanding, and still right

Its "require one known architecture bit" point is valid but under-specified, and I would take a stronger form. Instead of an OR-accumulated `supports_known_arch` flag, precompute the union of the bits `arches` covers and assert each row is a subset:

```cpp
ASSERT_TRUE(row.arch_mask != 0);
ASSERT_EQ(static_cast<uint8_t>(row.arch_mask & ~known_arch_bits), static_cast<uint8_t>(0));
```

That catches CodeRabbit's case and two it misses: a mixed mask like `kV81 | (1U << 3)`, and a new arch bit added to the table without being added to the test's `arches` array. Verified by setting `maple_preview` to `1U << 3`: main passes, the subset assert fails. I would grade it a latent robustness gap rather than Major, since no row today carries a bit outside `kV75|kV79|kV81`, but it is worth having.

## Nits

- `#include <unordered_set>` in the test is now unused. It earns its place back with the duplicate check above.
- `qhexrt_model_catalog.cpp` still says the private set is "pinned by test_qhexrt_model_catalog's private_ids" in the `requires_hf_auth` comment. This PR deleted `private_ids`, so that comment points at a symbol that no longer exists. Should say `private_row_count`.
- `policy_row_at` returning `policy.id.data()` is correct because every `id` is a `string_view` over a literal, and your comment says so. Making `ModelPolicy::id` a `const char*`, or returning a length, would remove the invariant instead of documenting it. Not worth a change on its own.

## On our side

This test does not run in CI. `RAC_BACKEND_QHEXRT` defaults OFF and no workflow turns it on, so `test_qhexrt_model_catalog` only ever executes when someone builds the backend locally, which is why the original drift sat there. That is our gap to close, and it is the reason your change is worth more than a test refactor usually is: the version that can drift silently is the one nobody runs.

Thanks again. Genuinely good instinct on which expectations should read from the table and which should stay pinned.

ayaangazali:
Taking the framing correction: you are right and I overstated it. The description reads as if the test was red at merge time, and it was not. I hit the mismatch when I opened #631, someone hand-patched the sets before this PR existed, and I carried the original framing forward without re-checking that it still described the tree I was merging into. The drift argument stands on its own without that claim, so it did not need dressing up. I will re-verify the premise against the actual merge base next time rather than the moment I first saw it.

The duplicate-id regression is a fair catch and a real loss. I removed the count guard because it could not see a mask widening and did not ask what it *could* see that the loop cannot, which is the same mistake in the opposite direction: I checked what the assertion failed to catch and not what it uniquely caught. `seen_ids.insert(row.id).second` is the right shape, and your point about `find_model_policy` stopping at the first match is what makes it matter rather than being tidiness. Noted that you are fixing it forward, so I will keep off that file.

Your stronger form of the arch-bit assertion is better than CodeRabbit's and better than what I would have written. Asserting `row.arch_mask & ~known_arch_bits == 0` catches the new-arch-bit-added-without-updating-the-test case, which is the one that would actually bite, and an OR-accumulated flag cannot. Leaving that to the same follow-up.

Two of the nits are mine and I should have caught them:

- The `requires_hf_auth` comment in `qhexrt_model_catalog.cpp` still points at `private_ids`, a symbol my own diff deleted. Leaving a dangling reference to something I removed is careless. Happy to send the one-word fix to `private_row_count`, or fold it into your follow-up if that is cleaner, whichever avoids the collision.
- `#include <unordered_set>` being orphaned is on me too. Agreed it earns its place back with the duplicate check, so it is not worth a churn PR ahead of yours.

The CI point is the one I will remember: `RAC_BACKEND_QHEXRT` defaults OFF and no workflow enables it, so the file only runs when someone builds the backend by hand. I only ran it because I checked whether I *could* gate it locally after being burned citing a test target that did not exist. Worth saying that a test nobody runs drifting silently is a stronger argument for the change than anything in my description.

PR #671, opened 2026-08-12
release: 0.20.18 — ship the Electron sherpa carrier fix

sanchitmonga22:
## Why this release exists

`@runanywhere/electron-sherpa@0.20.17` shipped a **non-functional 33 KB stub**. Its
`librunanywhere_sherpa.dylib` referenced none of `g_sherpa_{stt,tts,vad}_ops`, so
`sherpa_capability_check()` returned `RAC_ERROR_BACKEND_UNAVAILABLE` unconditionally and the
registry refused it with `-811` every time. Because one failing backend used to brick the whole
SDK, every Electron verb threw.

#670 fixed both halves — the registration isolation, and the CMake cause (`rac_add_engine_plugin`
applies backend `COMPILE_DEFINITIONS` **PRIVATE**, so the thin carrier that actually ships never
received them). **This release delivers that fix to users.**

## Proof the fix is real

The natives gate `scripts/validation/gates/check_plugin_natives.py` rejects the old artifact and
accepts the new one. Same script, same machine:

| artifact | sherpa carrier sha256 | ops refs | gate |
|---|---|---|---|
| published `electron-sherpa@0.20.17` | `61fab0fa…4245c` | **0 / 3** | **exit 1 — FAIL** |
| rebuilt `0.20.18` | `c2a05513…f849` | **3 / 3** | exit 0 — ok |

`nm -gU` cannot tell these apart: a healthy thin carrier is *also* ~33 KB and *also* exports zero
ops symbols, because those are **undefined** symbols resolved from the sibling
`librac_backend_sherpa.dylib`. The gate looks for the names whether defined or undefined, which is
the only signal that separates them.

Runtime, against the real addon:

```
# published 0.20.17 stub
loadPlugin -> REJECTED: rac_registry_load_plugin failed: -811
listPlugins():             []
listUnavailablePlugins():  [{ name: "sherpa", status: -811 }]

# rebuilt 0.20.18
LOADED sherpa / llamacpp / onnx
listPlugins():             ["llamacpp","onnx","sherpa"]
listUnavailablePlugins():  []
```

The `-811` line also shows the isolation half of #670 working —
`backend 'sherpa' is unavailable (-811) — other backends keep serving`.

## Ordering: natives rebuilt AFTER the bump

0.20.17 built natives *before* the bump, so every Electron package shipped binaries stamped
0.20.16 while `package.json` said 0.20.17. This release rebuilds after, and packaging verifies it
via `--expect-version`: every bundled `librac_commons.dylib` embeds `0.20.18`.

## Contents

- `scripts/release/sync-versions.sh 0.20.18` across all manifests
- three lockfiles refreshed; `## [0.20.18]` added to all five Flutter CHANGELOGs
- `check_release_version_coherence.sh` exits 0

## Not built on this machine

- **Windows natives** — not possible on macOS. Left to `release.yml`.
- **Android `.so` / WASM** — left to `release.yml`, which is also what produces the
  `RACommons-android-*-v0.20.18.zip` assets the Flutter Android side downloads.
- **`electron-qhexrt`** — deliberately **not** published. It has no prebuilds on this platform, and
  packaging now skips it rather than shipping an empty NPU package (which is how 0.20.17 shipped a
  4 KB JS-only tarball).

## Summary by CodeRabbit

- **New Release**
  - Released SDK version 0.20.18 across supported platforms and integrations.
  - Updated package dependencies and installation guidance for the new release.

- **Bug Fixes**
  - Restored Sherpa backend registration and availability.
  - Backend initialization now continues when an individual engine is unavailable.
  - Rebuilt Electron native backend components.

- **Documentation**
  - Updated architecture, installation, version references, and release notes to reflect version 0.20.18.


sanchitmonga22:
Thanks — both findings were right and are fixed in 338edda1d.

**Flutter per-package CHANGELOGs.** The entries described an Electron sherpa
carrier rebuild, which is not a change to any Flutter package. Only the commons
half of this release reaches Flutter (backend registration isolation), so the
entries now say that, and the Electron work is labelled as suite context rather
than as a change to the package the file documents. Applied to all five.

**`bindings/flutter/AGENTS.md` date.** Correct, and it is a recurring defect
rather than a one-off: `scripts/release/sync-versions.sh` rewrites the version
inside "Verified state: 2026-07-12 against the 0.20.17 package architecture"
but never the date, so *every* release makes that line assert the new version
was verified on an older date. Reworded to state what is actually true — the
architecture was last verified on that date and is unchanged as of 0.20.18 —
which stays accurate under future mechanical bumps.

Not adopted: the suggestion to date the line 2026-08-12. I did not re-verify
the Flutter package architecture in this release, so stamping today's date
would have made the line less true, not more.

PR #677, opened 2026-08-13
fix(react-native): ship rac_defaults_generated.h in the Android header set

sanchitmonga22:
## The break

`@runanywhere/core@0.20.18` **cannot compile for React Native Android**:

```
rac_llm_types.h:27:10: fatal error: 'rac/rac_defaults_generated.h' file not found
```

The published tarball carries that header in all three iOS xcframework `Headers/` directories but **not** in `android/src/main/jniLibs/include/rac/` — while five headers in that same Android bundle (`rac_{llm,stt,tts,vad,vlm}_types.h`) `#include` it.

## Cause

The header is gitignored codegen output (`.gitignore:543`), written by `generate_cpp_defaults.py`. `release.yml`'s `sdk_react_native` job generated only `ts`, so the C++ side never ran — and `package-sdk.sh` stages headers with a bare `cp -R` that copies whatever happens to exist.

An incomplete header set is not an error there. So the package built, validated, published, and failed only later inside a consumer's NDK compile.

## Fix

1. **`sdk_react_native` generates `ts,cpp`** so the header exists before staging. Sibling jobs audited: only this one stages `core/include/rac`.
2. **`package-sdk.sh` now fails closed** — after staging it resolves every intra-tree `#include "rac/..."` against the copied tree and aborts naming any header referenced but absent.

## The guard is verified, not assumed

Run against the actual broken 0.20.18 payload, it fails and names the culprit:

```
GUARD WOULD FAIL (correct) — missing:
   rac/rac_defaults_generated.h
```

Without it, the next missing codegen output ships the same silent way.

## Scope

npm is append-only, so `@runanywhere/core@0.20.18` stays broken for RN Android and needs a follow-up publish. **Only that package is affected** — `llamacpp`, `onnx` and `mlx` ship no Android headers at all (verified: 0 entries under `android/**/include/rac/`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved React Native SDK packaging for Android by including required generated native bindings.
  * Added validation to detect incomplete native header packages before release, preventing SDKs with missing dependencies from being distributed.
  * Packaging now stops with an actionable error when required generated files are unavailable.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -920,7 +920,18 @@ jobs:
       # PATH. RN packaging runs ensure_generated.sh --only ts; without ts-proto it exits 127.
       - uses: ./.github/actions/generate-idl
         with:
-          languages: ts
+          # ts for the proto-ts bindings the packaging script regenerates, AND
+          # cpp because package-sdk.sh copies core/include/rac wholesale into
+          # packages/core/android/src/main/jniLibs/include. One of those public
+          # headers, rac/rac_defaults_generated.h, is gitignored codegen output
+          # (.gitignore:543). Generating only ts left it absent while the five
+          # rac_{llm,stt,tts,vad,vlm}_types.h headers beside it still #include
+          # it, so @runanywhere/core@0.20.18 cannot compile for RN Android:
+          #   rac_llm_types.h:27:10: fatal error:
+          #   'rac/rac_defaults_generated.h' file not found
+          # The copy is an unguarded `cp -R`, so a missing generated header does
+          # not fail the build, it just ships an unusable package.
+          languages: ts,cpp
       # The RN tree pins `packageManager: yarn@3.6.1` (Yarn Berry). Activate it
       # via Corepack at the workflow level so package-sdk.sh (and any later
       # `yarn` invocation against the RN workspace) doesn't silently fall back
--- a/bindings/react-native/scripts/package-sdk.sh
+++ b/bindings/react-native/scripts/package-sdk.sh
@@ -376,6 +376,30 @@ rm -rf "$CORE_RAC_HEADERS"
 mkdir -p "$(dirname "$CORE_RAC_HEADERS")"
 cp -R "$PUBLIC_RAC_HEADERS" "$CORE_RAC_HEADERS"
 
+# FAIL CLOSED: the copy above is an unguarded `cp -R`, so it happily ships an
+# INCOMPLETE header set. Some public headers are gitignored codegen output --
+# rac/rac_defaults_generated.h is written by generate_cpp_defaults.py and is
+# #included by all five rac_{llm,stt,tts,vad,vlm}_types.h. If codegen did not
+# run for cpp, those five ship while their dependency does not, and the package
+# installs fine and then fails at NDK compile time in the consumer's app with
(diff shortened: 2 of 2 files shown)


sanchitmonga22:
## Post-merge review

This was merged with admin rights and never got the normal review pass. That is our process failure, not a reflection on the change. Here is the review it should have had, done against the real published artifact rather than the description.

### Verified

I pulled `@runanywhere/core@0.20.18` off npm and inspected the payload:

- `rac_defaults_generated.h` is present in all three xcframework `Headers/rac/` dirs and absent from `android/src/main/jniLibs/include/rac/` (162 headers ship there, zero of them that one).
- `features/llm/rac_llm_types.h:27` in that same Android tree is `#include "rac/rac_defaults_generated.h"`. The break is exactly as described.
- I ran the guard's loop verbatim against that extracted tree: exit 1, naming `rac/rac_defaults_generated.h`. Against an intact tree: pass. The guard does what it claims.
- `languages: ts,cpp` is a valid selector for `.github/actions/generate-idl`, and `cpp` codegen on a macos-26 runner is already proven by `native_rcli_macos` in the same workflow, so the added step is low risk.
- The audit claim holds. `bindings/react-native/scripts/package-sdk.sh` is the only packaging script that references `core/include/rac`.

Diagnosis is right and the instinct is right: guarding the artifact instead of only correcting the workflow input is the correct call, because the workflow input is the kind of thing that regresses quietly.

### Three things I would do differently

**1. The script still cannot generate what it validates.** `ensure_generated.sh --only ts` sits at line 497, roughly 120 lines below the header staging at line 376, and it never asks for `cpp`. So the actual fix lives entirely in `release.yml`, and any invocation outside that one job (local packaging, a fresh clone) now hard-fails at the new guard instead of self-healing. The repo convention is that each packaging script calls `ensure_generated.sh` before it packs. A `--only cpp` call above the `cp -R` would make the workflow change belt and braces rather than load bearing.

**2. CodeRabbit's blank-record comment is correct and I reproduced it.** When grep matches nothing, the here-doc still supplies one empty line, `_missing_headers` becomes a single space, and packaging aborts while printing an empty list. Not reachable today (63 quoted `rac/` includes in the tree), but it converts a future include-convention change into a mystery release failure with a useless message. One line fixes it: `[ -n "$_inc" ] || continue`. The array nitpick beside it is worth folding in while touching the block, though it is style rather than a bug, since C header paths here never contain spaces or globs.

**3. The guard only resolves `"rac/..."`-prefixed includes.** Three headers we ship in that Android tree reference protobuf headers that are not in it at all: `rac/infrastructure/storage/storage_event_publisher.h` needs `storage_types.pb.h`, and `rac/solutions/pipeline_executor.hpp` plus `rac/solutions/solution_runner.hpp` need `pipeline.pb.h` and `solutions.pb.h`. Zero `.pb.h` files ship in the 0.20.18 Android tree. The Nitro bridge does not include those headers today so nothing is broken, but it is the same failure class the guard exists to catch and the guard cannot see it.

### Still outstanding, and not code

npm `latest` for `@runanywhere/core` is still `0.20.18`, so RN Android stays broken for consumers until a republish. The PR body says this; flagging it here so it does not get lost.

Items 1, 2 and 3 go into a follow-up PR. Nothing here warrants a revert.

PR #678, opened 2026-08-13
fix(web): make --webgpu imply RAG the way --onnx-webgpu already does

ayaangazali:
Complements #673 rather than redoing it. That PR fixed the two call sites; this fixes the flag they were working around, so the trap cannot be stepped in again.

## What is wrong

`--onnx-webgpu` turns RAG on for itself. `--webgpu` does not:

```sh
--webgpu)
    WEBGPU="ON"
    LLAMACPP="ON"  # WebGPU accelerates llama.cpp
    shift
    ;;
--onnx-webgpu)
    ONNX="ON"
    ONNX_WEBGPU="ON"
    ONNX_REQUESTED="ON"
    RAG="ON"
    shift
    ;;
```

`--webgpu` enables neither ONNX nor `RAC_WASM_RAG_STANDALONE`, which is exactly the condition under which `core/CMakeLists.txt` force-sets RAG off:

```cmake
set(RAC_BACKEND_RAG OFF CACHE BOOL "RAG requires ONNX embeddings on Emscripten; force off when RAC_BACKEND_ONNX is not enabled" FORCE)
```

So a bare `--webgpu` build silently drops the eight `rac_rag_*_proto` exports its CPU twin has, which is the failure #673 describes.

## Why the call-site fix is not the whole fix

Three places still document the command that produces the broken bundle:

- `AGENTS.md:327` — `npm run build:wasm -- --webgpu   # WebGPU variant`
- `bindings/web/AGENTS.md:113`
- `bindings/web/docs/DEVELOPMENT.md:18`

Anyone following those gets the RAG-less bundle, and the next call site added anywhere gets it too. Fixing the flag makes all three correct without editing them.

## Verification

No Emscripten on this machine, so I could not build the bundle and compare exports. What I could do is prove the flag resolution deterministically, by printing the resolved variables just before the script's `emcc` check.

Before, on `main`:

```
$ bash bindings/web/wasm/scripts/build.sh --webgpu
PROBE RAG=OFF LLAMACPP=ON WEBGPU=ON ONNX=OFF
$ bash bindings/web/wasm/scripts/build.sh --onnx-webgpu
PROBE RAG=ON LLAMACPP=OFF WEBGPU=OFF ONNX=ON
$ bash bindings/web/wasm/scripts/build.sh --webgpu --rag     # what #673 pinned
PROBE RAG=ON LLAMACPP=ON WEBGPU=ON ONNX=OFF
```

`RAG=OFF` with `ONNX=OFF` is precisely the state that trips the force-off above.

After this change:

```
$ bash bindings/web/wasm/scripts/build.sh --webgpu
PROBE RAG=ON LLAMACPP=ON WEBGPU=ON ONNX=OFF
```

`bash -n` on the script is clean. The probe line was temporary and is not part of the diff.

Being explicit about the gap: I have shown the flag now resolves the way #673 wants, not that the emitted `.wasm` exports match. That last step needs an emsdk build, so it is worth a CI run confirming the WebGPU artifact still carries the `rac_rag_*_proto` symbols.

## Summary by CodeRabbit

* **New Features**
  * WebGPU builds now include RAG support alongside WebGPU and llama.cpp functionality.
  * RAG protocol symbols are available in WebGPU bundles.

### DIFF
--- a/bindings/web/wasm/scripts/build.sh
+++ b/bindings/web/wasm/scripts/build.sh
@@ -99,6 +99,13 @@ while [[ $# -gt 0 ]]; do
         --webgpu)
             WEBGPU="ON"
             LLAMACPP="ON"  # WebGPU accelerates llama.cpp
+            # Same reason --onnx-webgpu sets it below. core/CMakeLists.txt
+            # force-sets RAC_BACKEND_RAG=OFF on Emscripten unless ONNX or
+            # RAC_WASM_RAG_STANDALONE is on, and this target enables neither, so
+            # without it the WebGPU bundle silently ships without the eight
+            # rac_rag_*_proto exports its CPU twin has (that twin only keeps
+            # them because it is built in the same invocation as --onnx).
+            RAG="ON"
             shift
             ;;
         --onnx-webgpu)


sanchitmonga22:
Thanks for this, @ayaangazali. Reviewing it properly now.

Transparency first: this was merged with admin rights and skipped the normal review pass. That is on us, not on you. Copilot was quota-blocked and CodeRabbit returned "no actionable comments", so in practice nothing looked at this before it landed. This comment is the review it should have had.

## The change is correct

I verified it against main rather than taking the description on faith:

- Re-ran your probe on main's `build.sh`. `--webgpu` now resolves `RAG=ON`.
- `RAG` feeds both `-DRAC_BACKEND_RAG` and `-DRAC_WASM_RAG_STANDALONE` (`build.sh:416-417`), and `RAC_WASM_RAG_STANDALONE=ON` is what short-circuits the force-off at `core/CMakeLists.txt:77-79`. The mechanism you described is the one that actually fires.
- The WebGPU target shares `_llamacpp_exports`, and `RAC_EXPORTED_FUNCTIONS_RAG` is appended to that list under `if(RAC_BACKEND_RAG)` at `bindings/web/wasm/CMakeLists.txt:1469-1471`, before the WebGPU target consumes it. The symbols do land on `racommons-llamacpp-webgpu`.
- Your count of 8 is right (`bindings/web/wasm/CMakeLists.txt:1141-1150`). The comment five lines above that `if` says "6 rac_rag_*_proto symbols" and has been stale for a while. You matched the code, not the comment.
- Safe on the pthreads-off Asyncify WebGPU target: nothing under `core/src/features/rag/` touches `std::thread` or pthreads, and USearch is configured with SIMSIMD and FP16LIB off.
- Zero risk to release artifacts, which is the nice property. `build:wasm:all` and `release.yml` already pass `--rag`, so the published bundle's configuration is unchanged. This only repairs the hand-run and documented invocations, exactly as you claimed.

What I especially liked: you were precise about the boundary of your verification. "I have shown the flag now resolves the way #673 wants, not that the emitted `.wasm` exports match" is the right thing to write, and it matters more than you knew. PR CI's `wasm` job runs `cmake --preset wasm` and never invokes `build.sh`, so nothing in the pre-merge checks exercised this file at all. Your probe was the only evidence that existed. Leaving the now-redundant `--rag` at both call sites alone was also the right call.

## One gap, and it is ours to fix forward

The identical trap survives one flag over:

```
[--webgpu]    -> RAG=ON     (your fix)
[--llamacpp]  -> RAG=OFF
[]            -> RAG=OFF
```

The CPU llamacpp bundle is the artifact `core/CMakeLists.txt:66-70` names as the one the Docs/RAG tab auto-loads. Release CI hides it because `--core --llamacpp --onnx` is a single invocation and `--onnx` sets `RAG=ON`, so the shipped bundle keeps its exports. But anyone running the documented `npm run build:wasm -- --llamacpp` (`AGENTS.md:326`, `bindings/web/AGENTS.md:112`, `bindings/web/docs/DEVELOPMENT.md:17`) gets a bundle that differs from the released one by exactly those 8 exports, plus a `RAC_ENABLE_SOLUTIONS=ON with RAC_BACKEND_RAG=OFF` configure warning and a Solutions RetrieveNode that returns `RAC_ERROR_FEATURE_NOT_AVAILABLE`. Your own note in the diff ("that twin only keeps them because it is built in the same invocation as --onnx") predicted this precisely.

A follow-up PR will carry:

1. `RAG="ON"` in the `--llamacpp` case and in the no-target default block, same shape as yours.
2. The `--rag` help and header text, which still says "requires --onnx". That was already wrong before this PR; standalone `--rag` is the entire point of `RAC_WASM_RAG_STANDALONE`.
3. The export assertion you asked for. `build.sh` already `strings`-scans the `.wasm` for a "gzip -d" regression, so the same block can assert `_rac_rag_session_create_proto` is in the emitted glue whenever RAG is on. That converts this bug class from silent to loud.

Structural nit, not worth changing now: `RAG` is a script global applied to every target in an invocation, so `--core --webgpu` also configures the core target with RAG on. Harmless here (core exports no RAG symbols) and pre-existing with `--onnx`, but deriving it per target inside `build_target` is the cleaner shape.

No reviewer comments outstanding. Thanks again, good catch, and the writeup made it fast to verify.

ayaangazali:
Thank you for tracing the mechanism rather than taking my word for it, and for the two details I got wrong on the way to being right.

The one that matters: I said `--rag` sets `RAC_BACKEND_RAG`, and the reason the force-off is actually skipped is `RAC_WASM_RAG_STANDALONE=ON` short-circuiting it at `core/CMakeLists.txt:77-79`. I read the force-off condition and matched the flag, without following which of the two knobs `build.sh:416-417` uses to defeat it. Same outcome, but I described a mechanism I had not verified end to end, which is the kind of thing that is only harmlessly wrong until it is not.

Also noted that the "6 rac_rag_*_proto symbols" comment above the export list is stale and I matched the code rather than the comment. That was luck rather than method: I counted from `bindings/web/wasm/CMakeLists.txt:1141-1150` because I wanted a number I could stand behind, and did not notice the comment disagreed.

`--llamacpp` and the bare default carrying the identical trap is a better catch than mine. I scoped to `--webgpu` because that was the reported class and I did not want to make a behavioural claim about a target someone might deliberately want lean, but your reasoning settles it: the CPU llamacpp bundle is the artifact the Docs/RAG tab auto-loads, and the documented `--llamacpp` invocation produces something that differs from the released bundle by exactly those exports plus a Solutions RetrieveNode that fails at runtime. That is not a lean build anyone asked for. Leaving all three items to your follow-up.

The export assertion is the part I would most like to see land. `build.sh` already `strings`-scans the emitted glue for the gzip regression, so asserting `_rac_rag_session_create_proto` when RAG is on turns this whole class from silent to loud, and it would have caught the original bug without anyone reasoning about CMake force-offs.

Your structural nit is right too: `RAG` being a script global rather than derived per target inside `build_target` is why `--core --webgpu` configures core with RAG on. Harmless today, but it is the same shape of latent coupling that produced this bug.

Useful to know that PR CI's `wasm` job runs `cmake --preset wasm` and never invokes `build.sh`. That explains why the probe was the only evidence available, and it is worth me remembering that "CI is green" says nothing about a file CI does not execute.

PR #679, opened 2026-08-13
fix(react-native): export the two types RunAnywhere.solutions.run needs

ayaangazali:
Same gap #672 closed, two names it missed.

## What is wrong

`RunAnywhere.solutions.run` is public:

```ts
export const solutions = {
  run,
};
```

and `solutions` is on the `RunAnywhere` object (`Public/RunAnywhere.ts:575`). Its signature is:

```ts
async function run(args: SolutionRunArgs): Promise<SolutionHandle>
```

Neither `SolutionRunArgs` nor `SolutionHandle` is re-exported from `packages/core/src/index.ts`, so a consumer cannot annotate the argument they build or the handle they get back without importing from an internal path.

## Why #672 missed them

Both sit outside the shape that sweep covered:

- `SolutionRunArgs` is declared in `Public/Extensions/Solutions/RunAnywhere+Solutions.ts` rather than `Public/Api/Types.ts`, which is where the other 17 came from.
- `SolutionHandle` is a `class`, so a scan for exported `interface` and `type` declarations does not see it at all. That is the one I nearly missed too.

I checked the rest of the surface rather than assuming these were the only two. Every type declared in `Public/Api/Types.ts` is already exported, so that module is clean. Of the other unexported declarations under `src/Public`, I ruled out:

- `RegisterModelInput`, `RegisterArchiveModelInput`, `RegisterMultiFileModelInput` — internal. `Public/Api/Models.ts` builds those object literals itself; the public `models.register*` signatures take already-exported types.
- `StreamController` — only appears in `pushStream`, which is not exported from the entry point.
- `EventBus`, `SDKEventHandler`, `ModelLifecycleChange`, `EventBusCancellable` — no `events` accessor on the public `RunAnywhere` object, so consumers cannot reach them.
- `RegisteredTool`, `STTPartialResultWithOutput` — not in any public signature.

So this is two names, not a sweep.

## What this does

Adds a type-only re-export. `SolutionHandle` is constructed by the SDK from a native handle and its constructor rejects a null one, so callers only ever need to name it, never build it. Keeping it type-only avoids adding a runtime export for a class nobody should instantiate.

## Testing

I could not run the repo's gate on this machine. `yarn` is not installed and `npx yarn@3.6.1` fails here, so `yarn typecheck` was unavailable, and I am not going to claim a gate I did not watch.

What I did verify statically, which is the substance of what `tsc` checks for a re-export: the module path resolves (`Public/Extensions/Solutions/RunAnywhere+Solutions.ts` exists relative to `index.ts`), both names are exported from it (`export class SolutionHandle` at :42, `export type SolutionRunArgs` at :123), and neither name was already exported from `index.ts`, so there is no duplicate-export collision. Relying on the `rn-typecheck` CI job for the real check.

## Summary by CodeRabbit

* **New Features**
  * Added public type definitions for solution handles and solution execution arguments.
  * Improved type support for integrating with the Solutions extension.

### DIFF
--- a/bindings/react-native/packages/core/src/index.ts
+++ b/bindings/react-native/packages/core/src/index.ts
@@ -159,6 +159,15 @@ export { formatFramework } from './Public/Helpers/formatFramework';
 // Explicit tool-calling run loop — mirrors Swift/Kotlin generateWithTools for
 // hosts that need full control over the tool-loop budget (max calls, thinking,
 // parallel execution) beyond what llm.generate's inline tools expose.
+// `RunAnywhere.solutions.run(args)` takes SolutionRunArgs and resolves to a
+// SolutionHandle, so a consumer cannot annotate either side of that call
+// without these. Type-only: SolutionHandle is constructed by the SDK from a
+// native handle, never by callers.
+export type {
+  SolutionHandle,
+  SolutionRunArgs,
+} from './Public/Extensions/Solutions/RunAnywhere+Solutions';
+
 export { generateWithTools } from './Public/Extensions/LLM/RunAnywhere+ToolCalling';
 export type {
   GenerateWithToolsOptions,


sanchitmonga22:
Reviewing this after the fact. It was merged with admin rights and skipped the normal review pass. That was our process failure, not yours. This comment is the review it should have had before landing.

Thank you, and specifically for the second name. `SolutionRunArgs` sitting in `Public/Extensions/Solutions/RunAnywhere+Solutions.ts` instead of `Public/Api/Types.ts` is an easy miss, but `SolutionHandle` being a `class` is the genuinely interesting one: any audit that scans for exported `interface` and `type` declarations walks straight past it, and `RunAnywhere.solutions.run` is the only public verb in the RN surface that returns a class. You also went and checked the rest of `src/Public` and gave the reasoning for each name you ruled out instead of shipping a blind sweep. That writeup is why this took me ten minutes to verify rather than an hour.

I verified the claim against main rather than trusting the description:

- `solutions` really is on the public `RunAnywhere` object (`Public/RunAnywhere.ts:575`), so the call is reachable by consumers.
- Neither name appeared in `index.ts` or `internal.ts` before your change, so the only prior route was a deep path import. The gap was real.
- `tsc --noEmit` over `packages/core` passes on the merged state (92 source files, exit 0). You said you could not run the gate and did not claim you had, which I appreciate. It passes.
- A consumer importing both names from `@runanywhere/core` and calling `start` / `feed` / `closeInput` / `destroy` / `isAlive` typechecks. All three `SolutionRunArgs` variants annotate, and the union still rejects two input kinds at once.

Two judgment calls I think you got right. Type-only for `SolutionHandle`: the constructor is `@internal` and rejects a null handle, so a runtime export would only buy `instanceof`, and `run` already hands you the type. And leaving `SolutionConfig` alone: it comes from `@runanywhere/proto-ts/solutions`, and the convention here is that consumers depend on `proto-ts` directly (the example app does exactly that in ~20 files). `SDKEnvironment` is the exception, not the rule.

One thing I would do differently, being fixed forward on our side: the block landed between the `generateWithTools` comment and its export, so `generateWithTools` lost its doc comment and the Solutions export now sits under a paragraph that opens "Explicit tool-calling run loop". Cosmetic, but it will mislead the next reader of that file. A blank line plus moving the tool-calling paragraph back down its own export fixes it.

No reviewer comment is outstanding. CodeRabbit produced no actionable findings, and Copilot never ran because the requester was out of quota.

Separately, and not yours to fix: web has the same gap. `RunAnywhere.solutions` is public there (`bindings/web/packages/core/src/Public/API/RunAnywhere.ts:374`) but `bindings/web/packages/core/src/index.ts` never re-exports `SolutionHandle` or `SolutionRunInput`. Your PR is what made that visible. It goes into the same follow-up.

ayaangazali:
Thank you for verifying it against main rather than the description, and for confirming `tsc --noEmit` passes. I could not run that gate and said so; good to know the answer.

The comment-placement problem is mine and I should have caught it. I inserted the block between the `generateWithTools` doc comment and its export, which orphaned that paragraph and left the Solutions export sitting under prose about the tool-calling run loop. I was looking at the diff rather than the resulting file, which is exactly how that goes unnoticed. Noted you are fixing it forward, so I will not send a competing patch.

Good to have the two judgment calls confirmed, particularly `SolutionConfig`. I left it because it comes from `@runanywhere/proto-ts/solutions` and I could not tell whether re-exporting proto types was the convention or the exception here. Your point that the example app imports from `proto-ts` directly in about twenty files, and that `SDKEnvironment` is the exception rather than the rule, is the answer I was missing.

On web having the same gap: that one is genuinely worth flagging as a pattern rather than two incidents. `RunAnywhere.solutions` is public in both bindings, and in both the handle type is a class that a sweep for exported `interface` and `type` declarations walks straight past. If it is useful, the check that would catch the whole class is asserting that every type named in a public verb's signature is reachable from the package entry point, rather than auditing declarations. That is a bigger piece of work than either fix and I have not thought through what it costs, so I mention it as an idea rather than a proposal.

Happy to leave the web half to your follow-up since you are already carrying it.

PR #680, opened 2026-08-13
fix(commons): preserve model download path when allocation fails

shubhamsinnh:
## Description
Fix `rac_model_registry_update_download_status()` so replacing a model's download path is transactional when allocation fails.

Previously, the registry freed the existing `local_path` before duplicating the replacement. If `rac_strdup()` failed, the previous path was lost, `updated_at` was still mutated, and the function could persist the partial state and return success.

The replacement path is now allocated before registry mutation. Allocation failure returns `RAC_ERROR_OUT_OF_MEMORY` while preserving the existing path and timestamp. Passing `NULL` still explicitly clears the path without requiring an allocation.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring

## Testing
- [ ] Lint passes locally

Validation performed:
- `git diff --check origin/main...HEAD` — passed.
- Full diff, diff stat, and commit range inspected; only `core/src/infrastructure/model_management/model_registry.cpp` changed in one commit.
- `bash core/scripts/lint-cpp.sh` — not run because this Windows host has no installed WSL distribution.
- Direct `clang-format --dry-run --Werror` — not run because `clang-format` is unavailable.
- `cmake --preset windows-debug` — configuration attempted but could not run because Visual Studio 2022 is not installed.
- No tests were added because this focused change did not request new tests.

**Swift SDK / iOS Sample:**
- [ ] Tested on iPhone (Simulator or Device)
- [ ] Tested on iPad / Tablet
- [ ] Tested on Mac (macOS target)

- [ ] Tested on Android Phone (Emulator or Device)
- [ ] Tested on Android Tablet

- [ ] Tested on iOS
- [ ] Tested on Android

- [ ] Tested on iOS
- [ ] Tested on Android

**Web SDK / Web Sample:**
- [ ] Tested in Chrome (Desktop)
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] WASM backends load (LlamaCpp + ONNX)
- [ ] OPFS storage persistence verified (survives page refresh)
- [ ] Settings persistence verified (localStorage)

## Labels

**SDKs:**

**Sample Apps:**

## Checklist
- [x] Self-review completed

## Screenshots
Not applicable; this is a commons registry-state fix with no UI changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when updating model download locations.
  * Prevented existing paths from being lost if a new path cannot be stored.
  * Safely clears the download location when no local path is provided.

### DIFF
--- a/core/src/infrastructure/model_management/model_registry.cpp
+++ b/core/src/infrastructure/model_management/model_registry.cpp
@@ -687,13 +687,16 @@ rac_result_t rac_model_registry_update_download_status(rac_model_registry_handle
 
     rac_model_info_t* model = it->second;
 
-    // Free old local path
-    if (model->local_path) {
-        free(model->local_path);
+    char* replacement_path = nullptr;
+    if (local_path) {
+        replacement_path = rac_strdup(local_path);
+        if (!replacement_path) {
+            return RAC_ERROR_OUT_OF_MEMORY;
+        }
     }
 
-    // Set new local path
-    model->local_path = rac_strdup(local_path);
+    free(model->local_path);
+    model->local_path = replacement_path;
     // Proto ModelInfo.updated_at_unix_ms — milliseconds, not seconds.
     model->updated_at = rac_get_current_time_ms();
 


sanchitmonga22:
## Post-merge review (the one this PR should have had before it landed)

Transparency first: this was merged with admin rights about 36 minutes after it opened, with zero approving reviews and no inline comments. CodeRabbit's run came back `Review failed: The pull request is closed`, so no automated review landed either. That was a process failure on our side, not yours. This comment is the review it should have had.

Thank you @shubhamsinnh, and specifically: you found a real ordering bug in `rac_model_registry_update_download_status()`, not a theoretical one. The old code freed `model->local_path` and only then called `rac_strdup()`. On allocation failure the registry entry silently lost its download path, `updated_at` was still bumped, and the function fell through to `store_proto_snapshot_locked(..., overwrite_registry_state=true)`, which rewrote the cached proto snapshot in `handle->model_proto_bytes` with the missing path and then returned success. That snapshot is the read-back surface every SDK sees, and `rac_model_info_is_downloaded()` is derived purely from `local_path` (`core/src/infrastructure/model_management/model_types.cpp:672`), so the practical result was an already-downloaded model quietly reverting to "not downloaded" and the app offering to re-download gigabytes it already had. Allocate-then-swap is exactly the right shape for that.

### What I verified on main today

- `core/src/infrastructure/model_management/model_registry.cpp:690-699` reads as intended: allocate first, `return RAC_ERROR_OUT_OF_MEMORY` before any mutation, so both `local_path` and `updated_at` survive the failure.
- `rac_strdup()` returns `nullptr` for a `NULL` input (`core/src/core/rac_memory.cpp:39`), so the documented "NULL clears the path" contract in the header is preserved bit for bit. Your version makes that branch explicit instead of relying on the implicit `strdup(NULL)` behavior, which reads better.
- Dropping the `if (model->local_path)` guard before `free()` is fine, `free(NULL)` is a no-op, and the pointer is reassigned on the next line.
- There is no second field to keep in sync. `is_downloaded` is computed, not stored.
- All five callers already branch on the return code and degrade gracefully, so the new failure return is safe: `download_orchestrator.cpp:1733`, `model_lifecycle.cpp:696`, `model_registry_refresh.cpp:262`, `storage_analyzer.cpp:1326`, `runanywhere_commons_jni.cpp:2323`.
- `clang++ -std=c++20 -fsyntax-only -DRAC_HAVE_PROTOBUF` passes on the merged file.

Not verified: I did not run the native test suite or a device build against this change. PR CI was green on every native job at merge time except `swift-spm`, which was still pending when the merge happened.

### Outstanding reviewer comments

None. There were no human reviews and no inline comments, and CodeRabbit never produced substantive findings because the PR was already closed when it ran.

### One gap, and it is ours to close

The identical pattern is still live about 330 lines above, in `save_model_info_impl()` (`core/src/infrastructure/model_management/model_registry.cpp:362-366`):

```cpp
if (should_preserve_path) {
    if (copy->local_path)
        free(copy->local_path);
    copy->local_path = rac_strdup(existing_local_path);
}
```

If that `rac_strdup()` fails, `copy->local_path` becomes `NULL`, the previous entry is freed on the next line, `copy` replaces it in the map, and the function still returns success. To be precise about blast radius: `try_reconcile_model_local_path_locked()` runs a few lines later (`:395`) and will relink the path for models sitting in the canonical `{base_dir}/RunAnywhere/Models/{framework}/{id}/` folder, so many cases self-heal. The loss sticks for externally imported paths outside that folder and for folders `std::filesystem` cannot see, such as un-hydrated OPFS on Web. Same failure mode you fixed, in the branch whose entire job is not to lose the path. That is not a defect you introduced and not something we expect you to revisit on a merged change. We are fixing it forward in a follow-up PR.

Small note for next time, purely optional: a one-line "sibling site at :362 left alone deliberately" in the PR body would have made the scope call obvious to a reviewer.

Thanks again for a clean, well-scoped commons fix. This one benefits all five SDKs at once, which is exactly where we want fixes to land.

shubhamsinnh:
Glad I could contribute. 
Thanks for the detailed post-merge review.

PR #681, opened 2026-08-13
fix(commons): handle streaming result allocation failures

shubhamsinnh:
## Description
Handle allocation failures while constructing an owned streaming metrics result.

`rac_streaming_metrics_get_result()` previously returned `RAC_SUCCESS` even when duplicating the generated text or model ID failed. A model-ID allocation failure also left the already-allocated text in a partially populated result.

The result is now initialized to its canonical default, each owned-string allocation is checked, and partial output is released before returning `RAC_ERROR_OUT_OF_MEMORY`.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring

## Testing
- [ ] Lint passes locally

Validation performed:

- `git diff --check` — passed
- Complete diff and commit-range inspection — passed; one intended file and one focused commit
- `bash core/scripts/lint-cpp.sh` — not run because this Windows host has no installed WSL distribution and no native `clang-format`
- `cmake --preset windows-debug` — configuration unavailable because Visual Studio 17 2022 is not installed
- No tests were added or run, per repository guidance for this focused change

**Swift SDK / iOS Sample:**
- [ ] Tested on iPhone (Simulator or Device)
- [ ] Tested on iPad / Tablet
- [ ] Tested on Mac (macOS target)

- [ ] Tested on Android Phone (Emulator or Device)
- [ ] Tested on Android Tablet

- [ ] Tested on iOS
- [ ] Tested on Android

- [ ] Tested on iOS
- [ ] Tested on Android

**Web SDK / Web Sample:**
- [ ] Tested in Chrome (Desktop)
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] WASM backends load (LlamaCpp + ONNX)
- [ ] OPFS storage persistence verified (survives page refresh)
- [ ] Settings persistence verified (localStorage)

## Labels

**SDKs:**

**Sample Apps:**

## Checklist
- [x] Self-review completed

Documentation is unchanged because this does not alter the public API.

## Screenshots
Not applicable; this change has no UI impact.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streaming metrics result handling when memory allocation fails.
  * Added safer cleanup of partially allocated result data.
  * Ensured result structures are initialized consistently.

### DIFF
--- a/core/src/features/llm/streaming_metrics.cpp
+++ b/core/src/features/llm/streaming_metrics.cpp
@@ -235,11 +235,20 @@ rac_result_t rac_streaming_metrics_get_result(rac_streaming_metrics_handle_t han
     }
 
     // Populate result
+    *out_result = RAC_STREAMING_RESULT_DEFAULT;
     out_result->text = rac_strdup(handle->full_text.c_str());
-    out_result->thinking_content = nullptr;
+    if (!out_result->text) {
+        return RAC_ERROR_OUT_OF_MEMORY;
+    }
+
+    out_result->model_id = rac_strdup(handle->model_id.c_str());
+    if (!out_result->model_id) {
+        rac_streaming_result_free(out_result);
+        return RAC_ERROR_OUT_OF_MEMORY;
+    }
+
     out_result->input_tokens = input_tokens;
     out_result->output_tokens = output_tokens;
-    out_result->model_id = rac_strdup(handle->model_id.c_str());
     out_result->latency_ms = latency_ms;
     out_result->tokens_per_second = tokens_per_second;
     out_result->ttft_ms = ttft_ms;


sanchitmonga22:
Thanks @shubhamsinnh, and an apology up front: this was merged with admin rights before it got a review pass, with `swift-spm` still running. That was our process failure, not yours. This comment is the review it should have had.

### What your change actually fixed

`rac_streaming_metrics_get_result` called `rac_strdup` twice and checked neither. On a failed allocation it still returned `RAC_SUCCESS`, so a caller received a "successful" `rac_streaming_result_t` with a NULL `text` or a NULL `model_id`. The struct documents both as owned strings the caller must free, so nothing signals that a null-check is needed on the success path. An OOM was being reported as a completed generation. Converting that to `RAC_ERROR_OUT_OF_MEMORY` is the right call.

### The part that shows you read the code

Putting `*out_result = RAC_STREAMING_RESULT_DEFAULT;` first is load bearing, not cosmetic. `rac_streaming_result_free` unconditionally reads `result->thinking_content`, and under the old ordering that field was not written until after both `rac_strdup` calls. Without the default assignment, the cleanup on the `model_id` failure path would have called `free()` on whatever the caller happened to have in an uninitialized stack struct. Getting that ordering right is the difference between a fix and a fresh crash.

It also lines the function up with `rac_streaming_metrics_get_text` further down the same file, which already did `return *out_text ? RAC_SUCCESS : RAC_ERROR_OUT_OF_MEMORY;`. The file disagreed with itself and now it does not.

### Verified against main today

Read the merged diff against `39dc8f2c5` (current main HEAD), plus `rac_streaming_result_free`, `rac_strdup`, and the full `rac_streaming_result_t` definition:

- `rac_strdup` returns NULL only on a NULL input or a `malloc` failure, so an empty `full_text` or `model_id` still yields a valid 1-byte allocation. No false OOM.
- `rac_streaming_result_free` frees the three owned members with `free()` (matching `rac_strdup`'s `malloc`) and nulls them, and never frees the struct itself. Calling it on a caller-owned out param is correct, and the struct handed back on the OOM path is fully zeroed.
- `clang++ -std=c++20 -fsyntax-only` on the file passes.
- CI was green on `macos-debug`, `macos-release`, `linux-debug`, `linux-asan`, `android-arm64`, `ios-device`, and `wasm`. `swift-spm` never finished, but this diff is one `.cpp` with no header, ABI, or export-list change, so it cannot move that job.

One scoping note for anyone reading this later: `rac_streaming_metrics_get_result` has no in-tree caller today. It is exported public C ABI, so the blast radius is external consumers of `librac_commons` and whoever wires it up in-tree next. That does not make the fix less right, it makes it cheap insurance correctly taken.

### Outstanding reviewer comments: none

CodeRabbit posted its walkthrough, but the actual review never ran. Its comment reads "Review failed, the pull request is closed" because we merged out from under it. Zero inline comments and zero submitted reviews on this PR, so there is nothing unaddressed.

### One nit, and it is ours

The doc block for `rac_streaming_metrics_get_result` in `core/include/rac/features/llm/rac_llm_metrics.h` still says `@return RAC_SUCCESS or error code`. It should name `RAC_ERROR_OUT_OF_MEMORY` and state that `*out_result` is now unconditionally overwritten, since a caller reusing a struct that still owns strings would leak them. Two lines of Doxygen. We will fold it into a follow-up.

No defect found in the change itself. Tight, correctly scoped, and it got the hard part right. Thanks again.

shubhamsinnh:
Thanks @sanchitmonga22 
 I really appreciate the detailed review and the context you shared, especially the point about the default initialization preventing a second crash & yes, no worries about the merge timing.
 
 I appreciate you being so open about it, glad the fix held up. 
 Looking forward to contributing more and learning from reviews like this!

PR #682, opened 2026-08-13
docs(swift): point external consumers at runanywhere-swift, not this monorepo

sanchitmonga22:
`Package.swift` told external users to depend on:

```swift
.package(url: "https://github.com/RunanywhereAI/runanywhere-sdks", from: "0.20.18")
```

**That does not work.** The generated protobuf Swift sources under `bindings/swift/Sources/RunAnywhere/Generated/` are codegen output and no longer committed, so tag `v0.20.18` carries **1** file there where `v0.20.17` carried **42**. Resolving this URL by version fails with ~347 errors of the form `cannot find type 'RAModelInfo' in scope`.

Nothing warned about it: the de-commit landed between the two tags and the manifest kept advertising the old path, so a consumer following our own documentation gets a package that cannot build.

`RunanywhereAI/runanywhere-swift` is generated **with** those sources and is the supported path. It declares the same remote binaryTargets with the same checksums against the same release assets here, so no XCFramework is duplicated, and it is ~3 MB against this repo's ~340 MB. Verified: a scratch package with `from: "0.20.18"` against the distribution repo resolves cleanly (exit 0, all 6 binary targets checksum-validated).

Comment-only change. No target, dependency or checksum is touched.

## Summary by CodeRabbit

* **Documentation**
  * Updated external-consumer guidance to reference the dedicated Swift distribution repository.
  * Added clarification about generated sources, repository size, and matching binary targets and checksums.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -9,9 +9,29 @@ import Foundation
 // This is the SINGLE Package.swift for both local development and SPM consumption.
 //
 // FOR EXTERNAL USERS (consuming via GitHub):
-//   .package(url: "https://github.com/RunanywhereAI/runanywhere-sdks", from: "0.20.18")
-//   No environment override is needed. SPM downloads the checksum-verified
-//   XCFramework archives from the GitHub release by default.
+//   .package(url: "https://github.com/RunanywhereAI/runanywhere-swift.git", from: "0.20.18")
+//
+//   Consume the SWIFT DISTRIBUTION REPO, never this monorepo. Two reasons, and
+//   the first one is fatal:
+//
+//   1. This repo's tags DO NOT COMPILE as a Swift package. The generated
+//      protobuf Swift sources under bindings/swift/Sources/RunAnywhere/Generated/
+//      are no longer committed (they are codegen output), so tag v0.20.18
+//      carries 1 file there where v0.20.17 carried 42. Resolving this URL by
+//      version yields ~347 errors of the form
+//      "cannot find type 'RAModelInfo' in scope". The distribution repo is
+//      generated WITH those sources, so it builds.
+//   2. This monorepo is ~340 MB of C++, Kotlin, WASM and Dart. The distribution
+//      repo is ~3 MB of Swift.
+//
+//   Both manifests declare the SAME remote binaryTargets with the SAME
+//   checksums, pointing at the SAME release assets on runanywhere-sdks, so the
+//   XCFrameworks are never duplicated or re-uploaded. No environment override is
+//   needed; SPM downloads the checksum-verified archives from the GitHub release.
+//
+//   bindings/swift/scripts/sync-dist-repo.sh cuts that repo, and
+//   scripts/validation/gates/check_swift_dist_repo_sync.sh fails every PR here
+//   until it carries the matching tag.
 //
 // FOR LOCAL DEVELOPMENT:
 //   1. Build native XCFrameworks from the repo root:


sanchitmonga22:
## Post-merge review (this PR was admin-merged without one)

This landed via admin merge with `reviewDecision: REVIEW_REQUIRED`, no human review, and CodeRabbit's only comment was the "processing new changes" placeholder posted 16 minutes before the merge. It never came back with an actual review. That was our process failure, not a comment on the change. Here is the review it should have had.

### The claims check out

I verified each one against main today rather than trusting the description:

- **42 to 1 file.** `git ls-tree v0.20.17 -- bindings/swift/Sources/RunAnywhere/Generated/` gives 42 files; `v0.20.18` gives exactly one, `Versions.swift`. `.gitignore:494` is what changed underneath.
- **The tag genuinely does not compile.** `struct RAModelInfo` is defined only in `Generated/model_types.pb.swift`, and 14 tracked source files reference it. So this is not a theoretical break.
- **The distribution repo is a real destination.** `RunanywhereAI/runanywhere-swift` is 3082 KB, carries bare-semver tags `0.20.17` and `0.20.18`, so `from: "0.20.18"` resolves.
- **Same checksums, no duplicated XCFrameworks.** All six `checksum:` lines in the dist manifest at tag `0.20.18` are byte-identical to the six in root `Package.swift`.

One thing I did **not** verify: the "~347 errors" count. Reproducing it needs a full SwiftPM resolve of the 340 MB clone plus binary downloads, and I did not run that. The mechanism is proven; treat the number as approximate.

Good instinct putting the *reason* in the manifest instead of just swapping a URL. The next person who wonders why we do not just publish the monorepo now has the answer at the point of use.

### Three things this missed, being fixed forward

**1. It broke a CI gate, and main is red right now.**

`scripts/validation/gates/check_release_version_coherence.sh:199` hard-asserts the exact literal this PR deleted:

```bash
expect_literal "Package.swift" ".package(url: \"https://github.com/RunanywhereAI/runanywhere-sdks\", from: \"${VERSION}\")"
```

The gate exits 1 locally, and the `centralization` job fails at "Release-train version coherence" on main head `39dc8f2c`. It was already failing on this PR before merge, which is exactly what a review pass would have caught. The fix is the URL swap in both branches of that conditional (the `SPM_TEMP_PIN` one too), not a revert.

**2. The next release bump will abort.**

`scripts/release/sync-versions.sh:221` rewrites that same comment line with a regex anchored on `runanywhere-sdks`. It no longer matches, `bump_line` returns 1, and the script runs under `set -euo pipefail`, so `sync-versions.sh 0.20.19` dies at the Swift step *after* already writing `core/VERSION` and `core/VERSIONS`. Half-bumped tree, loud failure. Same one-line URL fix.

**3. The redirect is incomplete in the place that matters most.**

`bindings/swift/Sources/LlamaCPPRuntime/README.md:13` and `Sources/ONNXRuntime/README.md:13` still say:

```swift
.package(url: "https://github.com/RunanywhereAI/runanywhere-sdks", exact: "0.20.18"),
.product(name: "RunAnywhere", package: "runanywhere-sdks"),
```

`sync-dist-repo.sh` copies `git ls-files bindings/swift/Sources` verbatim, so these ship *inside* runanywhere-swift. I confirmed it live: those files exist at tag `0.20.18` in the dist repo with the broken URL. A consumer who follows the newly-correct `Package.swift`, resolves the package, then opens the backend README, gets sent right back to the path this PR calls fatal.

Also `bindings/swift/README.md:46` is still the monorepo URL at `0.20.11`, seven releases stale. It drifted because the version-coherence gate lists the two `Sources/*/README.md` files but not this one. Worth adding while we are in there.

### Net

The diagnosis was right and the direction was right. What was missing is that the string being edited was load-bearing in two scripts, and that the same guidance lives in three other files. A follow-up PR covers all four fixes. None of this is a reason to revert.

PR #691, opened 2026-08-14
release: 0.20.19 — unblock React Native Android, fix the QHexRT catalog

sanchitmonga22:
## Why this release exists

**`@runanywhere/core@0.20.18` cannot compile for React Native Android.** It shipped without `rac/rac_defaults_generated.h` while five headers beside it `#include` that file:

```
rac_llm_types.h:27:10: fatal error: 'rac/rac_defaults_generated.h' file not found
```

There is no consumer-side workaround. That is the only hard blocker in 0.20.18 and the reason 0.20.19 exists.

## What rides along

| fix | PR |
|---|---|
| Android header set completeness, plus a fail-closed `#include`-closure guard | #677 |
| QHexRT `lfm2_5_1_2b_thinking` policy row (`-259` on a real v81 device) | #688 |
| QHexRT `lfm2_5_vl_3b` no longer HF-auth gated (its repo is public) | #684 |
| Commons lifecycle / download-path / streaming-result allocation fixes | #668, #680, #681 |
| WebGPU RAG export parity | #673, #678 |
| Fail-closed pre-publish gate | #689 |
| Electron platform honesty | #690 |

## Two release-tooling bugs, found by cutting the release

Both were self-inflicted and are fixed here:

1. `bindings/swift/README.md` was in the **coherence gate's** checked list (added by #686) but **not** in `sync-versions`' bump list, so the gate failed on a file the bump never touched.
2. That loop used `bump_line`, whose `sed` has no `/g`. A line carrying the version several times kept stale copies — README line 35 does exactly that (`Select version **X** (or from: "X") ... if X binary assets`). Added `bump_all_on_line` for the docs loop; `bump_line` is unchanged for the ~30 anchored call sites where one-per-line is correct.

## Verification

- Coherence gate: **0.20.19, zero mismatches**
- No stale `0.20.18` in `core/VERSION`, `Package.swift`, `gradle.properties`, or the Swift README
- Lockfiles regenerated (web `package-lock.json`, root and RN `yarn.lock`)
- Flutter changelogs written per package, describing only what actually reaches each one

## Context

0.20.18 was audited first: 32 published artifacts verified with `nm -a`, five apps built and launched on real hardware (Web, Electron, iOS Simulator, macOS, Android on an SM8850 v81). **22/22 backend cells live, nothing hollow.** Those five apps are shippable on 0.20.18 today; this release is for React Native Android and the QHexRT catalog.

## Summary by CodeRabbit

* **New Release**
  * Updated the SDK and platform packages to version **0.20.19** across Swift, Kotlin, Flutter, React Native, Web, Electron, and Python.
  * Aligned package installation requirements and dependency versions across supported integrations.

* **Bug Fixes**
  * Model-loaded status now follows the model lifecycle.
  * Existing download paths are preserved when replacement allocation fails.
  * Streaming string allocation failures now report out-of-memory errors.
  * QHexRT no longer requires an unnecessary Hugging Face token.

* **Documentation**
  * Updated installation guides, architecture references, and changelogs for version 0.20.19.


sanchitmonga22:
All four CodeRabbit findings addressed in e4ddd1b0.

**1. RN MLX README installs a stale version (Minor).** Correct, and worse than it looks. The line said `npm install @runanywhere/core@0.20.19 @runanywhere/mlx@0.20.11` — a user following it would pair a current core with a nine-release-old MLX.

Fixed, but the interesting part is *why no bump caught it*: `sync-versions` rewrites the **previous** version only, so a literal that drifts once is invisible to every subsequent release forever. `0.20.11` has never matched any bump pattern since `0.20.11`. I swept the other seven docs in the release-doc loop for the same shape; this was the only one.

**2, 3, 4. `normalize-swift-modhash.py` (1 Critical, 2 Major).** All three are real, and the right fix is that **the file should not have been in this PR at all.** It has been removed.

It is an inert helper an agent wrote while investigating `RunAnywhereMLXRuntime`'s byte-reproducibility. It was never executed against a real archive, nothing references it, and it was deliberately not wired into the build because doing so would change archive bytes and invalidate committed checksums mid-release. It reached this commit through a careless `git add -A` on my part, not by intent.

Shipping an unreviewed, never-run archive-rewriting script inside a release commit is exactly the kind of thing that should not slip through, so thank you for catching it. Your three findings stand on their merits and should be addressed if and when it is wired in for real:

- raw member size vs. BSD extended name in the alignment math
- load-command extent validation against the member boundary
- preserving successful no-op behavior for an archive with no `__swift_modhash` section

I have kept the script outside the repo rather than landing a half-verified fix for a tool nothing calls yet.

Also added the `release:patch` label the `centralization` gate correctly required for a version-bumping PR.

PR #695, opened 2026-08-14
release: 0.20.20 — republish Electron with telemetry baked in

sanchitmonga22:
## Why

The 0.20.19 Electron packages were built from a local tree with `STAGING_BASE_URL` unset, so CMake substituted the `YOUR_STAGING_BASE_URL` placeholder into `development_config.cpp`. `rac_dev_config` rejects it, the SDK cannot keyless-resolve a development backend, and **those packages report no metrics at all**.

Every CI-built artifact was fine — verified directly in iOS ×3 slices, Android ×3 ABIs, RN, Maven, and all 5 WASM binaries. Only the hand-built Electron ones were affected, because `release.yml` fails closed on a missing secret and a local build silently does not.

## Why 0.20.20 and not a patch to 0.20.19

npm is append-only and `unpublish` is 2FA-gated. An Electron-only manifest bump is not viable:
- `package-sdk.sh:50-51` derives the proto-ts tarball name from the electron manifest version
- `check_plugin_natives.py --expect-version` substring-checks `librac_commons.dylib` against `core/VERSION`

A prerelease is worse: `0.20.19-hotfix.1` sorts **below** 0.20.19 and satisfies neither `^0.20.19` nor the `>=0.1.0 <1` that electron-llamacpp/onnx/sherpa declare for `@runanywhere/electron` — the backends would resolve back to the broken build and the fix would reach nobody.

## Scope

Only the **5 Electron packages** are published at 0.20.20. The other 18 stay at 0.20.19.

**No `v0.20.20` tag is pushed.** The SwiftPM manifests are deliberately held at 0.20.19 (there are no v0.20.20 release assets), which `check_release_version_coherence.sh` explicitly supports and reports as `[OK] SwiftPM manifests temporarily pin 0.20.19`. The next full release train is **0.20.21**.

## Verification

- `prepublish_check.py` gained a telemetry gate (d113d2eb1) that fails exactly this bug — proven FAIL on the bad tarball, PASS on the fixed one
- all 5 tarballs pass the gate at 0.20.20
- `@runanywhere/electron-sherpa@0.20.20` was pulled back **from the registry** and re-verified: `placeholder=0 staging=1`
- `check_release_version_coherence.sh`: `[OK] release version coherence: 0.20.20`

The 4 bad 0.20.19 versions are deprecated on npm pointing at 0.20.20.

## Summary by CodeRabbit

* **New Features**
  * Released version 0.20.20 across the SDK ecosystem, including Electron, Flutter, Kotlin, Python, React Native, Swift, Web, and supporting packages.
  * Updated package dependencies and installation examples to reference version 0.20.20.
  * Improved Flutter example plugin registration for Android and iOS file-picker support.

* **Documentation**
  * Updated version references, architecture notes, installation instructions, and changelogs to reflect 0.20.20.


sanchitmonga22:
Merging with admin override and **deliberately no `release:*` label**.

`check_release_version_coherence.sh:71` fails a version change without a release label, but that label makes `auto-tag.yml` push `v0.20.20` and dispatch a full release run. That must not happen: no v0.20.20 release assets exist, so the tag would break `check_swift_dist_repo_sync.sh` on every subsequent PR until `runanywhere-swift` were also cut at 0.20.20.

`auto-tag.yml` is an explicit no-op when no release label is present (see its header comment), so merging unlabelled lands `core/VERSION=0.20.20` without creating a tag — which is exactly the intent.

Merging is still necessary: the 5 Electron packages are already published at 0.20.20 on npm and that is irreversible. If `core/VERSION` stayed at 0.20.19, the next release train would target 0.20.20 and collide with the live packages. Recording the bump here makes the next train **0.20.21**.

The coherence gate passes on main after this lands, because base and VERSION will both be 0.20.20 and the label rule only applies to a version *change*.

PR #696, opened 2026-08-14
fix(commons): rate VLM decode_tokens_per_second over the decode window

ayaangazali:
This is the same metric the Kotlin fix (#634) corrected, found one layer down in commons itself. Both VLM paths were affected; see the comment below for the non-streaming half added after opening.

## What is wrong

`populate_result_from_stream` rates the throughput over the whole request span:

```cpp
out->set_total_time_ms(elapsed_ms);
if (elapsed_ms > 0) {
    // TokenUsage has no tokens_per_second field; decode_tokens_per_second
    // is the decode-phase-only throughput this maps onto.
    out->mutable_usage()->set_decode_tokens_per_second(
        static_cast<double>(ctx.token_count) / (static_cast<double>(elapsed_ms) / 1000.0));
}
```

`elapsed_ms` is `now_ms() - ctx.started_ms`, measured from before `process_stream`, so prefill is inside the denominator of a field named `decode_tokens_per_second`. The comment above it concedes the mapping rather than the arithmetic, which is what makes it easy to miss.

## Why that is wrong

`llm_module.cpp` excludes prefill from this metric deliberately and states why:

```cpp
```

So the two modules in the same library disagree about what the same field means, and the LLM one is the definition every SDK was pointed at. For a VLM the prefill is the image encode rather than a text prompt, so the understatement is larger here than the text case that motivated the original rule.

There is a second, quieter half. `GeneratedStreamCtx` never recorded a first-token time, so `usage.ttft_ms` stayed 0 on this path, while the telemetry a few lines below reads `result.usage().ttft_ms()` (`vlm_module.cpp:1321`, `:1477`) and therefore reported 0 for every streamed VLM generation.

## What this does

- Records `first_token_ms` for the first token that survives the sentinel filter. The trampoline already distinguishes that token to emit `GENERATION_EVENT_KIND_FIRST_TOKEN_GENERATED`, so this reuses the existing point rather than adding a second notion of "first".
- Publishes it as `usage.ttft_ms`.
- Rates the throughput over `elapsed_ms - ttft_ms`.

Falls back to the full span when the first-token time is unknown or is not strictly inside the span, so a cancelled stream that produced no token reports exactly what it reported before. Both call sites, the cancelled and the completed one, go through the same helper.

## Scope

Only the streaming path. The non-streaming path already tracks `first_token_time` separately (`vlm_module.cpp:527-528`, `:562-564`) and sets `time_to_first_token_ms` from it, so it was already honest about the distinction; this brings the streaming path in line rather than changing that one.

## Testing

Built and run locally on macOS with `-DRAC_BUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Debug`:

- `test_vlm_options_defaults` — exit 0, `vlm option default tests passed`
- `test_llm_stream_proto` — exit 0, `0 test(s) failed`

Being explicit about what those do and do not cover: they exercise the surrounding proto and options surface, not the streaming rate arithmetic itself, which has no test today. I did not add one, because reaching `populate_result_from_stream` means driving `ref.ops->process_stream` through a fake VLM engine and controlling the clock, which is a fair amount of scaffolding. Happy to add it if you would rather the arithmetic were pinned.

## Summary by CodeRabbit

* **New Features**
  * Streaming results now include time-to-first-token metrics.
  * Added decode throughput calculations based on the interval after the first token.
  * Added accurate metrics for streams with delayed prefill.

* **Bug Fixes**
  * Timing and usage data now appear for completed and cancelled streams.
  * Added fallback throughput calculations when first-token timing is unavailable.
  * Corrected token and throughput reporting for streams that produce no tokens.

### DIFF
--- a/core/src/features/vlm/vlm_module.cpp
+++ b/core/src/features/vlm/vlm_module.cpp
@@ -688,10 +688,21 @@ extern "C" rac_result_t rac_vlm_component_process_stream(
         final_result.time_to_first_token_ms = ttft_duration.count();
     }
 
-    // Calculate tokens per second
-    if (final_result.total_time_ms > 0) {
+    // Tokens per second over the decode window, not the whole span. This value
+    // is mapped onto TokenUsage.decode_tokens_per_second by
+    // rac_proto_adapters.cpp, and llm_module.cpp states the rule: prefill in
+    // the denominator systematically understates generation speed. TTFT is
+    // computed immediately above, so the window is already known here.
+    // Falls back to the full span when no token was seen, or when TTFT is not
+    // strictly inside the span.
+    const int64_t vlm_decode_ms =
+        (final_result.time_to_first_token_ms > 0 &&
+         final_result.time_to_first_token_ms < final_result.total_time_ms)
+            ? final_result.total_time_ms - final_result.time_to_first_token_ms
+            : final_result.total_time_ms;
+    if (vlm_decode_ms > 0) {
         final_result.tokens_per_second = static_cast<float>(final_result.completion_tokens) /
-                                         (static_cast<float>(final_result.total_time_ms) / 1000.0f);
+                                         (static_cast<float>(vlm_decode_ms) / 1000.0f);
     }
 
     if (complete_callback) {
@@ -1029,6 +1040,7 @@ rac_result_t check_lifecycle_model(const runanywhere::v1::VLMGenerationRequest& 
 struct StreamCtx {
     std::string text;
     int32_t token_count{0};
+    int64_t ttft_ms{0};
 };
 
 void populate_result_from_stream(const StreamCtx& ctx, int64_t elapsed_ms,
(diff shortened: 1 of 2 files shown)


ayaangazali:
Expanded this before anyone reviewed it: the first revision fixed only the streaming path and left the identical bug in the non-streaming one, in the same file.

```cpp
// Calculate tokens per second
if (final_result.total_time_ms > 0) {
    final_result.tokens_per_second = static_cast<float>(final_result.completion_tokens) /
                                     (static_cast<float>(final_result.total_time_ms) / 1000.0f);
}
```

`time_to_first_token_ms` is computed two lines above that and simply not used, and `final_result.tokens_per_second` reaches the same `decode_tokens_per_second` field through `rac_proto_adapters.cpp:517-519`. So shipping only the streaming half would have fixed the metric on one path and left it understated on the other, which is exactly the incomplete-sweep problem I have been flagging in other people's fixes.

Both paths now rate over the decode window with the same fallback. Description updated.

Re-verified after the change: `test_core --run-all` exit 0, `18 passed, 0 failed, 18 total`; `test_vlm_options_defaults` exit 0.

Siddhesh2377:
Added the coverage you offered to add, in `test_advanced_modality_proto_abi.cpp` — the mock VLM and lifecycle-load harness `test_vlm_process_stream_events` already builds was enough scaffolding, so it only needed a stream op that sleeps.

Two cases. `test_vlm_stream_decode_rate_excludes_prefill` runs 400ms of prefill against a ~40ms decode and asserts `usage.ttft_ms` lands at or past the prefill, stays strictly inside the span, and that `decode_tokens_per_second` equals `tokens / ((total - ttft) / 1000)` exactly — both numbers come off the same result, so it pins the formula rather than the machine. A second assertion checks the rate is well clear of the whole-span rate, since the exact-formula one passes on the old tree where ttft is 0 and the window collapses to the span. `test_vlm_stream_without_tokens_falls_back` covers the guard: no token, so ttft stays 0, the window falls back to the full span, and the rate reports 0 as before.

Reverting `vlm_module.cpp` to origin/main turns both halves red (`publishes a real ttft_ms` and `well clear of the whole-span rate`). Restored: 227 checks, 0 failures, and stable across repeated runs. `test_core --run-all` is 18/18.

One note on the timing — a tight decode loop puts the first and last token in the same millisecond, which trips the strictly-inside-the-span guard and falls back for a reason unrelated to the arithmetic. Hence the 10ms gap between tokens; the comment in the mock says so.

Siddhesh2377:
Fixed the assertion that went red on `macos-debug`. My `decode_tokens_per_second > whole_span_rate * 3.0` check encoded how idle my laptop was: on a loaded runner the 10ms decode gaps stretch further than the 400ms prefill does, which collapses the ratio below 3. It now compares the two rates directly. That holds for any prefill a real runner produces, because the rates differ by exactly the ttft the assertion above already pins, and it still fails on the old code where ttft is 0 and the window collapses to the span. Reverting `vlm_module.cpp` still gives 2 failures; restored, 227 checks 0 failures across repeated runs.

The `rcli-windows` failure on the previous run was unrelated to this PR. It was a link-step race: `vcpkg z-applocal` on `test_model_id_from_url.exe` hit "The process cannot access the file because it is being used by another process". No compile error anywhere in the log. Rerunning it.

Siddhesh2377:
Thanks Ayaan - clear diagnosis, and the two modules now agree on what the field means.

PR #700, opened 2026-08-14
fix(react-native): point the podspec source at a tag that exists

ayaangazali:
Follows #697, which fixed this for Flutter. The RN podspecs have the same defect in a different shape and were not covered.

## What is wrong

The four RN podspecs resolve their source from a git tag built from `s.version`:

```ruby
s.version      = package["version"]
s.source       = { :git => "https://github.com/RunanywhereAI/runanywhere-sdks.git", :tag => "v#{s.version}" }
```

`package.json` is at `0.20.20`. The newest tag upstream is `v0.20.19`:

```
$ gh api repos/RunanywhereAI/runanywhere-sdks/tags --jq '.[].name' | head -3
yaprun-v0.1-mac
voice-assistant-v0.1.0
v0.20.19
```

So `pod install` resolves `v0.20.20`, which was never pushed, and fails. Unlike the Flutter case this does not degrade quietly, it stops the install.

## Why it happens

Exactly the reason #697 gives for Flutter. `s.version` tracks the repo version, which leads the last published release whenever a release republishes only some packages, and 0.20.20 did precisely that (it republished Electron). #697 handled the Flutter side by decoupling `asset_version` from `s.version`, and its own comment says to bump back "once a matching release exists". The RN side kept deriving from `s.version`.

## What this does

Pins the source tag to the release the SwiftPM manifests already pin, using the same shape as Flutter's `asset_version`, and extends `check_release_version_coherence.sh` to cover the RN podspecs.

The gate is the part that matters. It already asserted `asset_version == SPM_VERSION` for the four Flutter podspecs, and RN was simply not in the loop, which is why this side drifted while the Flutter side was being fixed.

`s.version` is untouched, so the published CocoaPods version still tracks `package.json`. Only the source reference moves.

## Verification

The gate fails on the unfixed state and passes on the fixed one. Reverting one podspec to `origin/main` and re-running:

```
[FAIL] bindings/react-native/packages/core/RunAnywhereCore.podspec: expected 'source_tag_version = '0.20.19''
[FAIL] bindings/react-native/packages/core/RunAnywhereCore.podspec: source tag interpolates s.version; it must use source_tag_version so it points at a tag that exists
[FAIL] release version coherence: 2 mismatch(es)
```

With all four fixed:

```
[OK] SwiftPM manifests temporarily pin 0.20.19 until v0.20.20 assets are published
[OK] release version coherence: 0.20.20
exit 0
```

`bash -n` on the gate is clean.

I did not run `pod install`, so I have shown the tag does not exist and that the gate now enforces the pin, not that CocoaPods resolves end to end. That needs macOS with CocoaPods and a published pod.

One follow-up I did not fold in: when v0.20.20 is tagged, both `asset_version` and this `source_tag_version` need bumping together. `sync-versions.sh` deliberately leaves `asset_version` alone, so the same applies here, and the gate will fail loudly if only one moves.

## Summary by CodeRabbit

- **Chores**
  - Improved React Native package release consistency by pinning package sources to the intended release tag.
  - Added automated checks to detect mismatched or incorrectly derived release tags before publication.
  - Added validation coverage to ensure all React Native packages are included in release-coherence checks.
  - Reduced the risk of installation or integration issues caused by packages referencing inconsistent release versions.

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -70,6 +70,11 @@ jobs:
             git fetch --no-tags --depth=1 origin "${PR_BASE_SHA}"
           fi
           bash scripts/validation/gates/check_release_version_coherence.sh
+      - name: Release-train version coherence gate tests
+        # The step above only proves the tree is coherent today. These prove the
+        # gate still rejects the podspec source-tag shapes it was added for, so
+        # it cannot quietly stop catching them.
+        run: python3 scripts/validation/gates/test_release_version_coherence.py
       - name: Swift distribution repo (runanywhere-swift) is cut at this release
         # RunanywhereAI/runanywhere-swift is the lightweight SPM distribution of
         # bindings/swift. Its tags MUST keep pace with this repo's releases or
--- a/bindings/react-native/packages/core/RunAnywhereCore.podspec
+++ b/bindings/react-native/packages/core/RunAnywhereCore.podspec
@@ -13,7 +13,15 @@ Pod::Spec.new do |s|
 
   s.platforms    = { :ios => "17.5" }
   s.swift_version = "6.2"
-  s.source       = { :git => "https://github.com/RunanywhereAI/runanywhere-sdks.git", :tag => "v#{s.version}" }
+  # TEMP: the source tag is pinned to the last release that was actually
+  # tagged, not to s.version. s.version tracks package.json and leads the
+  # published release whenever a release republishes only some packages, and
+  # a tag built from s.version then resolves to one that was never pushed, so
+  # `pod install` fails outright. Mirrors asset_version in the Flutter
+  # podspecs. Bump back in step with the SwiftPM pin once a matching release
+  # exists; check_release_version_coherence.sh enforces that they agree.
+  source_tag_version = '0.20.19'
+  s.source       = { :git => "https://github.com/RunanywhereAI/runanywhere-sdks.git", :tag => "v#{source_tag_version}" }
 
   # =============================================================================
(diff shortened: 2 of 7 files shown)


ayaangazali:
Rebased onto current `main` after 0.20.21 landed. Re-ran the gate on the new base: exit 0, `[OK] release version coherence: 0.20.21`.

The pin is unchanged and still correct. `Package.swift` still holds `sdkVersion = 0.20.19` and the newest tag upstream is still `v0.20.19`, so `source_tag_version = '0.20.19'` continues to name the last release that exists.

The gap this fixes is now two releases wide rather than one: `package.json` reads `0.20.21`, so an unfixed podspec resolves `:tag => "v0.20.21"`, which was never pushed.

Siddhesh2377:
Added `scripts/validation/gates/test_release_version_coherence.py` and wired it into the `validate` job next to the other script-contract tests.

Four cases: the unmodified tree passes (so the negatives can't be blamed on one of the gate's other forty assertions), each of the four podspecs reverted to `:tag => "v#{s.version}"` is rejected with the interpolation message, a pin to the wrong release is rejected against `Package.swift`'s `sdkVersion`, and the gate's podspec list is asserted to equal what's actually on disk — that last one is the failure you described, RN drifting because it was never in the loop.

It runs the real gate against a symlink mirror of the repo with only the mutated podspec copied for real, so nothing writes to the working tree. Reverting `RunAnywhereMLX.podspec` to origin/main turns it red on both messages; restored, 4 pass.

Siddhesh2377:
Thanks Ayaan - extending the gate to cover RN is the part that stops this drifting again.

PR #702, opened 2026-08-15
fix(commons): stop dropping real_time_factor from the STT completed event

ayaangazali:
## What is wrong

`emit_stt_transcription_completed` accepts `real_time_factor` and assigns every other argument to the event, but never that one:

```cpp
void emit_stt_transcription_completed(..., int32_t word_count, double real_time_factor,
                                      const char* language, int32_t sample_rate,
                                      rac_inference_framework_t framework) {
    ...
    voice.set_processing_duration_ms(static_cast<int64_t>(duration_ms));
    voice.set_input_audio_duration_ms(static_cast<int64_t>(audio_length_ms));
    voice.set_input_audio_bytes(audio_size_bytes);
    voice.set_word_count(word_count);
    if (language)
        voice.set_language(language);
    voice.set_sample_rate(sample_rate);
    voice.set_framework(...);
```

The field is not missing. `VoiceLifecycleEvent` declares it:

```proto
double              real_time_factor         = 18;   // STT
```

So the argument is taken and discarded, and the field is 0 for every subscriber no matter what a caller measured.

## Why it went unnoticed

`set_real_time_factor` appears exactly once in the tree, in `core/tests/test_telemetry_extraction.cpp:138`:

```cpp
vo->set_real_time_factor(0.5);
```

The field is readable and covered by a test, it is just never written by production code, so nothing failed.

There is a consumer that noticed the absence and worked around it, in `bindings/flutter/example/.../stt_benchmark_provider.dart:68`:

```dart
// Public Transcription has no commons real_time_factor; leave unset
```

## Scope, and being straight about impact

One line, the assignment the function was already shaped to make.

I checked how live this is rather than overselling it. The JNI bridge forwards a `realTimeFactor` argument (`runanywhere_commons_jni.cpp:3727`, `RunAnywhereBridge.kt:1050`), but I could not find anything on the Kotlin side that computes a non-zero value today, so in practice the observable change right now is that a real measurement would survive instead of being dropped. I would rather state that than imply subscribers immediately start seeing new numbers.

## One related thing I deliberately did not fix here

`emit_tts_synthesis_completed` drops `characters_per_second` the same way, and I found it with the same sweep. That one is different: `VoiceLifecycleEvent` has no `characters_per_second` field at all, so the argument cannot be delivered without an IDL change and regenerated bindings. Wiring one and quietly leaving the other would have been worse than saying so. Happy to take it as its own PR if you want the field added, or to delete the dead parameter instead.

## Testing

Built and run locally on macOS with `-DRAC_BUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Debug`:

- `test_telemetry_extraction` — exit 0, `39 checks, 0 failures`

Neither asserts the new assignment, since no test drives `emit_stt_transcription_completed` and observes a subscriber. The verifiable claim is the grep above: before this change the only `set_real_time_factor` in the repo was in a test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Voice lifecycle events now include the correct real-time factor when speech-to-text transcription is completed.
  * Text-to-speech completion events now report characters-per-second throughput.
  * Telemetry now captures and delivers both speech-to-text real-time factor and text-to-speech characters-per-second metrics for more complete performance reporting.


Siddhesh2377:
Thanks Ayaan - the dropped-argument sweep was the right instinct; the TTS half and the telemetry read are wired now too.

Siddhesh2377:
Thanks Ayaan - the dropped-argument sweep was the right instinct; the TTS half and the telemetry read are wired now too.

PR #703, opened 2026-08-15
fix(electron): tell the truth about Windows, and unstick bundle:native

sanchitmonga22:
`0.20.21` published `win32-x64` (llamacpp / ONNX / Sherpa) and `win32-arm64` (QHexRT on the Hexagon NPU) natives, but the tree still described a darwin-only release. Nothing here changes behaviour for an existing consumer — it makes the repo say what the published packages already contain, and removes a coupling that blocks the machine producing those natives.

## 1. `resolveAddon()` denied a prebuild it ships

`PREBUILT_PLATFORMS` was `['darwin-arm64']`, so a Windows user whose addon failed to load for any reason got:

> No prebuild ships for `win32-x64`. This release carries `darwin-arm64` only.

That message is the first thing anyone reads when a load fails, and it pointed away from the truth — the payload was sitting right there. Now `['darwin-arm64', 'win32-x64', 'win32-arm64']`, which is exactly what `bundle-native` stages and what the six published tarballs contain.

## 2. The package README was the npm page for "not on npm"

`README.md` opened with *"Unpublished preview (`private: true`, version `0.1.0`), not on npm"* and sent readers to build from source. It is listed in `files[]`, so it ships **inside** the tarball and has been the package page on npm for several releases. It now documents installing from the registry, states which platform gets which engines, and keeps build-from-source where it is still the answer (Linux, or working against local native changes).

## 3. `bundle:native` needed an unrelated module to compile

`bundle:native` ran `build:scripts`, which compiles every `scripts/**/*.ts` — including `manual-resolve.ts` and its `import { resolveModel } from '../dist/download'`. So staging natives required the SDK's TypeScript to be built first, which is not true on a box whose only job is producing prebuilds. Reproduced on this branch by parking `dist/`:

```
scripts/manual-resolve.ts(12,46): error TS2307: Cannot find module '../dist/download'
scripts/manual-resolve.ts(13,39): error TS2307: Cannot find module '../dist/download'
```

`bundle-native.ts` imports only `fs` and `path`. It now compiles through its own `tsconfig.bundle.json`, which succeeds on that same `dist`-less tree. `build:scripts` and `typecheck` keep their current coverage, so CI is unaffected and `manual-resolve.ts` is still type-checked.

This is the concrete blocker hit while building the `0.20.21` Windows natives: the operator had to hand-compile `bundle-native.ts` to stage a release.

## Verification

`npm run typecheck` clean, 230 unit tests pass, and the `dist`-less compile above was run both ways on this branch.

## Summary by CodeRabbit

- **New Features**
  - Added Electron native prebuilt support for macOS ARM64, Windows x64, and Windows ARM64.
  - Added Windows ARM64 support using the QHexRT engine with Hexagon NPU acceleration.
  - Improved platform-specific backend selection during installation.

- **Documentation**
  - Updated Electron SDK installation and source-build guidance.
  - Clarified supported platforms, available engines, and Windows ARM64 limitations.
  - Documented configuration options for local native builds.

### DIFF
--- a/bindings/electron/AGENTS.md
+++ b/bindings/electron/AGENTS.md
@@ -94,27 +94,20 @@ Adapted from `thoughts/shared/plans/BEST_PRACTISES.md` for this package:
 
 - Document what is true today. Do not claim encryption or remote Phase-2 auth unless
   packaging and runtime are wired end-to-end.
-- **Windows QHexRT/NPU is NOT wired, and neither is Windows.** As of 0.20.18 this SDK
-  ships prebuilds for **`darwin-arm64` only** — the addon at `prebuilds/darwin-arm64/`
-  plus one directory each in the llamacpp / onnx / sherpa packages. There is nothing
-  for linux, win32, or x64. `package.json` keeps `os`/`cpu` permissive on purpose,
+- **Windows ships at 0.20.21.** Prebuilds: `darwin-arm64` (llamacpp / ONNX / Sherpa),
+  `win32-x64` (llamacpp / ONNX / Sherpa), `win32-arm64` (QHexRT / Hexagon NPU only).
+  Linux still has no prebuild. `package.json` keeps `os`/`cpu` permissive on purpose,
   because the TypeScript facade genuinely is cross-platform and narrowing them would
   also block TS-only consumers and our own Linux CI; `resolveAddon()` in
-  `src/bridge.ts` instead names the gap explicitly against `PREBUILT_PLATFORMS`.
-- **`@runanywhere/electron-qhexrt` ships no native plugin on any platform.** It is
-  pinned at 0.20.17 and npm-deprecated saying so. `register()` records a path with no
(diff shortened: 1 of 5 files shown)


sanchitmonga22:
@coderabbitai thanks — one finding, declining it.

**`PREBUILT_PLATFORMS` as typed `{platform, arch}` objects:** not taking this. The array exists to answer one question, `is this host's prebuild in the release`, and the value it is compared against is `${process.platform}-${process.arch}` — already the compound form. Splitting it into fields means composing them back at both use sites (the `includes()` check and the `join(', ')` in the error message) to recover the string we started with. That is more code and one more place for the two spellings to drift, for no behaviour change.

The compound string is also the wire format everywhere else this concept appears: the `prebuilds/<platform>-<arch>/` directory name, `bundle-native.ts`'s `platformArch`, and the resolver path in this same file. Keeping one spelling is what stops the resolver and the staging script disagreeing about what a platform is called.

PR #704, opened 2026-08-15
fix(electron): stage each backend vendor runtime beside its own plugin

sanchitmonga22:
Found by the Windows agent testing the **published** `0.20.21` packages on `win32-x64`. This is a live defect in what is on npm right now, not a hypothetical.

## What breaks

Against pristine published natives, the inference gate gives **2 passed, 3 failed**:

```
ok  the engines this platform ships are actually loaded
ok  llamacpp generates text        "Colour"  15.9 tok/s
x   onnx embeds text               Error: Backend not ready
x   sherpa synthesises speech      Error: inference host exited unexpectedly (code 3221225477)
x   sherpa transcribes speech      Error: inference host exited unexpectedly (code 3221225477)
```

`3221225477` is `0xC0000005` — an access violation that takes the utility host down.

## Why

`rac_backend_onnx.dll` imports `onnxruntime.dll` and `rac_backend_sherpa.dll` imports `sherpa-onnx-c-api.dll`, but `bundle-native.ts` staged those vendor runtimes into **`@runanywhere/electron` only**. Commons loads a plugin with `LOAD_WITH_ALTERED_SEARCH_PATH`, which searches the *plugin's* own directory — the core package's copy is not on that path. Same story on the other platforms via `@loader_path` / `$ORIGIN`.

Copying the two DLLs beside each plugin and changing nothing else: **5 passed, 1 skipped**, embedding dims `[384, 384]`, tts `47564` samples, stt `" the quick brown fox."`.

## Why nobody caught it

The app's own `scripts/use-sdk.mjs` overlay already stages vendor runtimes **per package** (`if (id === 'onnx') files.push(ort, ortProviders)`). So the dev overlay was more correct than the shipping bundler, and byte-identical natives passed from a local build while failing from the tarball. Every Windows validation to date ran through that overlay.

macOS is unaffected: the vendor libraries are statically linked into `librac_backend_*.dylib` there, which is why the darwin packages ship three files and no `libonnxruntime`.

## The change

Staged file counts move from `[core:5, llamacpp:3, onnx:3, sherpa:3]` to `[core:5, llamacpp:3, onnx:5, sherpa:6]`. Core keeps its copies — the fat/dev path still resolves from beside the addon.

## Note for the release

The natives themselves are correct and unchanged; only staging was wrong. Reaching users needs a republish, since npm is immutable — see the release discussion on the tracking issue.

Reported and fixed by the Windows agent on a Snapdragon X2 Elite; PR opened from the Mac because `gh` is not installed on that box.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved desktop application packaging for ONNX and sherpa-based features.
  * Required runtime components are now included more reliably across supported platforms, reducing missing-library errors.
  * Builds remain resilient when optional components are unavailable on Linux and macOS.
  * Windows packaging now reports missing required runtime components early, helping prevent incomplete application bundles.

### DIFF
--- a/bindings/electron/scripts/bundle-native.ts
+++ b/bindings/electron/scripts/bundle-native.ts
@@ -203,20 +203,34 @@ function backendCandidates(id: BundlePackageId): readonly string[] {
 }
 
 /**
- * Vendor runtimes a backend links directly, by platform-correct file name.
- * ONNX Runtime is shared by both engines; sherpa additionally needs its C API.
+ * Vendor runtimes a backend links DYNAMICALLY, by platform-correct file name.
+ *
+ * macOS stages none: the vendor libraries are linked statically into
+ * `librac_backend_<id>.dylib`, which is why the darwin packages ship three files
+ * and no `libonnxruntime`. Windows and Linux link them dynamically.
  */
 function vendorRuntimeNames(id: BundlePackageId): readonly string[] {
+  if (mac) return [];
   const ort = win
     ? ['onnxruntime.dll', 'onnxruntime_providers_shared.dll']
-    : mac
-      ? ['libonnxruntime.dylib']
-      : ['libonnxruntime.so.1', 'libonnxruntime.so'];
+    : ['libonnxruntime.so.1', 'libonnxruntime.so'];
   if (id === BundlePackageId.Sherpa) {
-    return [...ort, win ? 'sherpa-onnx-c-api.dll' : mac ? 'libsherpa-onnx-c-api.dylib' : 'libsherpa-onnx-c-api.so'];
+    return [...ort, win ? 'sherpa-onnx-c-api.dll' : 'libsherpa-onnx-c-api.so'];
   }
   return ort;
 }
+
+/**
+ * Whether a missing vendor runtime should fail the bundle rather than warn.
+ *
+ * Windows is a proven contract: `rac_backend_onnx.dll` and
+ * `rac_backend_sherpa.dll` name these in their PE import tables, so a package
+ * staged without them is guaranteed broken — 0.20.21 shipped exactly that.
+ * Linux stays best-effort because no Linux prebuild is published and its file
+ * set is unverified; failing there would break local staging on a guess rather
+ * than on evidence.
+ */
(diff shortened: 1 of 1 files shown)


sanchitmonga22:
@coderabbitai two findings — took the first, declined the second. Pushed 2f0ebdf.

**1. Silently omitting a required vendor sidecar — agreed, and this is the important one.** Best-effort staging is precisely how `0.20.21` shipped broken: the file was absent, the bundler printed nothing, and the defect reached a user's machine as `Backend not ready` and an access violation naming no DLL. A dependency the binary cannot start without is the wrong thing to skip quietly.

Missing vendor runtimes now stop the bundle and print the backend, the platform, the file and every candidate path searched. Scope, deliberately:

- **Windows — required.** Proven, not assumed: `rac_backend_onnx.dll` and `rac_backend_sherpa.dll` name `onnxruntime.dll` and `sherpa-onnx-c-api.dll` in their PE import tables, so a package staged without them cannot load.
- **macOS — stages none.** The vendor libraries are statically linked into `librac_backend_<id>.dylib`, which is why the darwin packages ship three files and no `libonnxruntime`. The previous revision listed `libonnxruntime.dylib` for macOS, so making that required would have broken darwin staging outright. Verified unchanged: `[core:2, llamacpp:3, onnx:3, sherpa:3]`.
- **Linux — best-effort.** No Linux prebuild is published and its vendor file set is unverified. Failing there would be acting on a guess.

**2. A per-backend ONNX Runtime for Sherpa — declining, on evidence.** The theory is reasonable (sherpa-onnx vendors its own ORT build), but the single shared copy is what was measured working on the device: with one `onnxruntime.dll` staged into both packages, the win32-x64 inference gate goes 2 passed / 3 failed → **5 passed, 1 skipped**, with real embeddings `[384, 384]`, 47564 TTS samples and a correct transcription. The app's own dev overlay has also resolved one shared ORT for both engines for as long as it has existed.

Giving sherpa a *different* ORT than the one that produced that result would replace a verified configuration with an unverified one, in the same change that fixes a shipping defect. If the two builds ever genuinely diverge, the symptom is a load failure at staging time and this fail-closed path now reports it with the exact paths searched.

PR #705, opened 2026-08-15
fix(electron): resolve backend plugins out of app.asar.unpacked

sanchitmonga22:
The SDK-side root cause of the packaged-app failure the Windows agent found. [runanywhere-electron#6](https://github.com/RunanywhereAI/runanywhere-electron/pull/6) works around it in our app to unblock shipping; this fixes it for **every** packaged Electron consumer.

## What breaks

```
backends: []
backend:qhexrt — the plugin library could not be loaded
                 (missing file, wrong architecture, or unresolved symbols)
```

Each backend package calls `resolvePluginArtifactPath()` with a `packageRoot` taken from its own `__dirname`. Packaged, that is a path **into `app.asar`** — virtual, with no file behind it — so electron-builder's `asarUnpack` writes the real bytes to `app.asar.unpacked`.

## Why it was invisible

Electron's `fs` shim reads both transparently. So `pluginArtifactExists()` returns **true** for the virtual path, the path survives the existence filter into `RUNANYWHERE_PLUGIN_PATHS`, and then the OS loader — which has no asar shim — cannot open it. The addon itself resolved fine, so the SDK came up healthy and only the engines vanished, and nothing in the failure text names the archive.

## The change

`resolvePluginArtifactPath()` rewrites `app.asar` → `app.asar.unpacked`, and only when the unpacked file is actually present. An app that unpacks nothing, and an ordinary unpackaged tree, behave exactly as before.

This belongs in the SDK, not in each app: the path is computed here from a `packageRoot` the app never supplies, every packaged consumer hits it, and the failure gives them no route to this cause.

## Verification

Verified on a packaged Windows ARM64 build (Snapdragon X2 Elite, Hexagon v81) against the published `0.20.21` packages — `backends` went from `[]` to `["QHEXRT"]`, generating at 55.34 tok/s.

Four new unit tests cover the rewrite, the nothing-unpacked case, and a package root that merely contains the text `asar`. 234 tests pass, typecheck clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved plugin artifact loading in packaged desktop applications.
  - Uses unpacked plugin files when available while preserving valid archive paths when they are not.
  - Prevents incorrect path rewrites for unrelated filenames that merely contain similar text.

- **Tests**
  - Added coverage for packaged paths, unpacked files, fallback behavior, and path-matching edge cases.

### DIFF
--- a/bindings/electron/src/backend/plugin-registry.ts
+++ b/bindings/electron/src/backend/plugin-registry.ts
@@ -98,13 +98,40 @@ export function pluginPrebuildDir(locator: PluginArtifactLocator): string {
 }
 
 /**
+ * Rewrite a path that lands inside `app.asar` to the unpacked copy beside it.
+ *
+ * A packaged Electron app resolves this package's root from `__dirname`, which
+ * is a path INTO the asar archive — virtual, with no file behind it. Native
+ * code cannot load from there, so electron-builder's `asarUnpack` writes the
+ * real bytes to `app.asar.unpacked/...`. Electron's `fs` shim reads both
+ * transparently, so the existence check below passes on the virtual path and
+ * hands the OS loader a file it cannot open: the addon comes up, every backend
+ * then reports "the plugin library could not be loaded", and nothing names the
+ * archive. Every packaged consumer hits this, so the rewrite belongs here
+ * rather than in each app.
+ *
+ * Only rewritten when the unpacked file is actually there, so an app that
+ * unpacks nothing, or a plain unpackaged tree, is left exactly as it was.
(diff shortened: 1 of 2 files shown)


sanchitmonga22:
@coderabbitai agreed — fixed in adaa5cd.

`stageAsarLayout` created a temp tree per run and left it in `os.tmpdir()`. It now returns the root and both asar tests remove it in a `finally`, so a failing assertion cannot leak one either. 234 tests pass and no `ra-asar-*` directories remain afterwards.

PR #707, opened 2026-08-15
chore: release 0.20.22 — fix the Windows staging that 0.20.21 shipped broken

sanchitmonga22:
Suite version bump only, so the 5 Electron packages can be republished. The other 18 stay where they are.

## Why a new version at all

`0.20.21` is the first release carrying Windows natives, and it went out with two defects — both found on the device, both already fixed on `main`:

| defect | effect | fix |
|---|---|---|
| ONNX and Sherpa staged **without** the vendor runtimes their backends link | on `win32-x64`, onnx reported `Backend not ready` and sherpa took the utility host down with `0xC0000005` | [#704](https://github.com/RunanywhereAI/runanywhere-sdks/pull/704) |
| packaged Electron apps resolved plugins to a path inside `app.asar` | **zero** backends loaded, on every platform | [#705](https://github.com/RunanywhereAI/runanywhere-sdks/pull/705) |

npm is immutable, so neither fix can reach anyone as `0.20.21`.

## Why the natives must be rebuilt even though they are correct

The binaries are fine — only *which files got copied where* was wrong. But `check_plugin_natives.py --expect-version` substring-checks the shipped `rac_commons` against the package version, so a `0.20.22` package needs a `rac_commons` that embeds `0.20.22`. Both Windows lanes and macOS get rebuilt against this tree before the corrected staging can ship.

Decoupling the npm version from the native version would avoid that, and I considered it: the natives genuinely are 0.20.21 and rebuilding them to satisfy a string check verifies a tautology. Rejected because it needs three changes to the release path (the proto-ts tarball name derives from the electron manifest, the `--expect-version` source, and `rewrite_npm_package.py`'s exact-version rewrite) and the very next step after those changes is an irreversible publish. The rebuild also lets the **fixed** bundler produce the payload rather than anyone hand-copying DLLs, which is the whole point of [#704](https://github.com/RunanywhereAI/runanywhere-sdks/pull/704).

## Asset versions deliberately do not move

Same as `0.20.21`, and `sync-versions.sh` does not know it — it bumped all three, and this reverts them to `0.20.19`, the last release that actually has GitHub release assets:

- `Package.swift` `sdkVersion` and the documented consumer pin
- `bindings/kotlin/gradle.properties` `runanywhere.nativeLibVersion`
- the 4 Flutter podspec `asset_version` literals (already correct)

No `v0.20.20`, `v0.20.21` or `v0.20.22` tag exists and none has release assets, so each of those would otherwise resolve a download URL that 404s. The unpublished-assets pin exists for exactly this state; the `TEMP` marker in `Package.swift` now names `v0.20.22` as the release being waited on.

## Gate

```
[OK] SwiftPM manifests temporarily pin 0.20.19 until v0.20.22 assets are published
[OK] bindings/kotlin/gradle.properties pins nativeLibVersion 0.20.19 until v0.20.22 assets are published
[OK] release version coherence: 0.20.22
```

Nothing is published by this PR. The Windows rebuild happens against this tree, then macOS stages, gates and publishes.

## Summary by CodeRabbit

- **Release**
  - Updated the SDK and platform packages to version **0.20.22** across supported integrations.
  - Aligned package dependency requirements and installation examples with the new release.
  - Republished Electron packages with corrected Windows native staging, including required ONNX and Sherpa runtime assets.

- **Documentation**
  - Updated version references, architecture notes, installation instructions, and changelogs to reflect 0.20.22.


sanchitmonga22:
**On the `centralization` failure:** `[FAIL] PR changes version 0.20.21 -> 0.20.22 without a release:* label`.

Merging without the label, deliberately, the same way [#701](https://github.com/RunanywhereAI/runanywhere-sdks/pull/701) and the `0.20.20` republish were merged. `release:patch` **auto-tags `v0.20.22` on merge**, and a tag here would be a claim we cannot back:

- `v0.20.20` and `v0.20.21` do not exist as tags either — both were Electron-only npm republishes with no GitHub release assets. `v0.20.19` is the last tag, which is exactly why `Package.swift`, `nativeLibVersion` and the podspec `asset_version` literals all stay pinned there in this PR.
- Creating `v0.20.22` with no assets would make `check_swift_dist_repo_sync.sh` fail **every** subsequent PR until someone cuts the Swift distribution repo for a release that ships nothing for Swift.

This bump exists only so five npm packages can be republished with corrected Windows staging. When a real release train tags again, the asset pins move together and the label applies then.

PR #708, opened 2026-08-15
fix(electron): resolve the sidecar dirs out of app.asar.unpacked too

ayaangazali:
Follows #705. That PR introduced `asarUnpacked()` and applied it to the plugin library path; the two sibling paths in `bridge.ts` still hand archive paths to the OS loader.

## What is wrong

`resolveCommonsLibrary` accepts a path on `fs.existsSync` and returns it:

```ts
export function resolveCommonsLibrary(addonDir: string): string | undefined {
  const candidate = path.join(addonDir, commonsLibraryFileName());
  return fs.existsSync(candidate) ? candidate : undefined;
}
```

That is the exact trap your own comment on `asarUnpacked` describes: inside a packaged app the candidate can point into `app.asar`, `fs.existsSync` succeeds through Electron's shim, and the file behind it is virtual. The result does not stay in JS, it goes to `addSidecarDirToDllSearch`, which is an OS-level call:

```ts
function prepareNativeLoad(addonPath: string): void {
  const addonDir = path.dirname(addonPath);
  addSidecarDirToDllSearch(addonDir);
  const commons = resolveCommonsLibrary(addonDir);
  if (commons) addSidecarDirToDllSearch(path.dirname(commons));
```

`addonPath` is still an archive path at that point. `resolveAddon` builds the packaged candidate from `__dirname`, and `require()` succeeds on it because Electron patches `process.dlopen` to extract a `.node` out of asar. The addon therefore loads, and every directory registered for the loader afterwards points inside the archive, so `rac_commons` and the vendor runtimes staged next to it by #704 are not on the search path.

## Why the plugin fix did not cover it

The distinction is which loader sees the path. The plugin library is `dlopen`ed from C++, so it never passes through Electron's patched `process.dlopen` and #705 had to rewrite it. The addon does pass through that patch, which is why it loads from the archive and hides the problem, while the sidecar directories derived from it silently do not work.

## What this does

Sends both paths through the existing `asarUnpacked`, exported for the purpose rather than duplicated. It already returns the input unchanged when the path is not inside `app.asar` and when the unpacked file is not present, so unpackaged trees and apps that unpack nothing are untouched.

## Testing, and what I could not verify

`npm install` in `bindings/electron` succeeds, but `tsc -p tsconfig.json --noEmit` cannot resolve `@runanywhere/proto-ts/*` on this machine: that package's `exports` map to `./dist/*`, and building it needs `@bufbuild/protobuf/wire`, which `npm install` inside `bindings/proto-ts` will not install here because it resolves against the monorepo root and hits an ERESOLVE conflict.

So I checked the diff does not add errors rather than claiming a green typecheck:

- clean tree: 118 `error TS` lines, all `TS2307` proto-ts resolution
- with this diff: 118, identical
- errors in `bridge.ts` or `backend/plugin-registry.ts`: none in either run

I also have not run this in a packaged Windows app, which is where it matters. The reasoning above is from the code and from the mechanism your #705 comment documents, so it is worth a packaged smoke test before trusting it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved native component loading in packaged desktop applications.
  * Ensured required libraries and add-ons are correctly located when running from packaged installations.
  * Fixed path handling for components stored outside the application archive.
  * Improved compatibility across packaged and unpackaged installations, including cases where optional libraries are unavailable.
  * Added coverage for packaged path resolution and native library loading scenarios.

### DIFF
--- a/bindings/electron/src/backend/plugin-registry.ts
+++ b/bindings/electron/src/backend/plugin-registry.ts
@@ -113,7 +113,7 @@ export function pluginPrebuildDir(locator: PluginArtifactLocator): string {
  * Only rewritten when the unpacked file is actually there, so an app that
  * unpacks nothing, or a plain unpackaged tree, is left exactly as it was.
  */
-function asarUnpacked(filePath: string): string {
+export function asarUnpacked(filePath: string): string {
   if (!filePath.includes(`app.asar${path.sep}`)) return filePath;
   const unpacked = filePath.replace(
     `app.asar${path.sep}`,
--- a/bindings/electron/src/bridge.ts
+++ b/bindings/electron/src/bridge.ts
@@ -6,6 +6,7 @@
 import * as fs from 'fs';
 import * as path from 'path';
 
+import { asarUnpacked } from './backend/plugin-registry';
 import { asSDKException } from './errors';
 
 // Re-exported for existing importers (RunAnywhere.ts imports it from here); the
@@ -485,7 +486,12 @@ function commonsLibraryFileName(platform: NodeJS.Platform = process.platform): s
 
 /** Absolute path to shared commons beside a thin addon, or `undefined`. */
 export function resolveCommonsLibrary(addonDir: string): string | undefined {
-  const candidate = path.join(addonDir, commonsLibraryFileName());
+  // Same asar trap resolvePluginArtifactPath guards against: inside a packaged
+  // app this path can point into app.asar, where fs.existsSync succeeds through
+  // Electron's shim but the OS loader sees nothing. The result feeds
+  // addSidecarDirToDllSearch, which is an OS-level call, so it has to be the
+  // real file on disk.
+  const candidate = asarUnpacked(path.join(addonDir, commonsLibraryFileName()));
   return fs.existsSync(candidate) ? candidate : undefined;
 }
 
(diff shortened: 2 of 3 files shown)


Siddhesh2377:
Added `test/unit/asar.test.ts` covering both paths this touches: `asarUnpacked` (outside the archive, rewritten when the unpacked file exists, left alone when nothing was unpacked, and the directory case `prepareNativeLoad` relies on) and `resolveCommonsLibrary` staged in `app.asar.unpacked`, absent, and in a plain unpackaged tree.

They run under plain Node without Electron's fs shim, which is the useful half of the packaged shape — the archive path genuinely has nothing behind it, same as what dlopen sees. Reverting the `bridge.ts` line drops `resolveCommonsLibrary finds commons staged in the unpacked tree` to red and the rest stay green, so it discriminates.

Full suite: 241 pass, 0 fail. On the proto-ts blocker you hit — `npm install --package-lock=false --workspaces=false` inside `bindings/proto-ts` then `npm run build` resolved it here, and `tsc -p tsconfig.json` is clean afterwards. Still unverified in a packaged Windows app, as you said.

Siddhesh2377:
Thanks Ayaan - good catch that the addon loads from the archive and hides the problem for the sidecars.

PR #711, opened 2026-08-16
fix(commons): give rac_model_storage_metrics_t a free function

### LINKED ISSUE #636: rac_model_storage_metrics_t has no documented free function for the strings get_model_metrics allocates
`rac_storage_analyzer_get_model_metrics` fills the out struct with three heap copies:

```cpp
out_metrics->model_id = model->id ? strdup(model->id) : nullptr;
out_metrics->model_name = model->name ? strdup(model->name) : nullptr;
...
out_metrics->local_path = strdup(model->local_path);
```

The header (`include/rac/infrastructure/storage/rac_storage_analyzer.h`) documents no free function for `rac_model_storage_metrics_t`, and `rac_storage_info_free` covers the array-returning call rather than this one. So a caller of the single-model function has no documented way to release those three strings.

Every SDK reaches this through the C ABI, so if there is no intended owner the leak is per call on all of them.

Three ways this could go, and I do not know which you intend:

1. A `rac_model_storage_metrics_free` exists and I missed it, in which case the header should reference it from `get_model_metrics`.
2. The caller is meant to free the fields directly, in which case documenting that on the function is enough.
3. It is a genuine gap and wants a new free function, which is ABI surface and your call.

Raising rather than guessing, since adding a public function is not something to slip into a bug fix. Noticed while working on #625, which fixes an unrelated error-path issue in the same function.

Siddhesh2377:
Closes #636.

## The gap

`rac_storage_analyzer_get_model_metrics` fills the out struct with three `strdup` copies:

```cpp
out_metrics->model_id = model->id ? strdup(model->id) : nullptr;
out_metrics->model_name = model->name ? strdup(model->name) : nullptr;
out_metrics->local_path = strdup(model->local_path);
```

The header documented no owner for them. `rac_storage_info_free` covers the array returned by `get_info`, not this call.

Of the three outcomes the issue offered, it is the third: a genuine gap wanting a new function.

## It is not hypothetical

`StorageBridge::getModelStorageMetrics` in the React Native core package calls it, copies the three strings into `std::string`, and returns. Nothing frees the originals, so every call from the RN SDK leaks all three. That is fixed here as well.

## What this adds

`rac_model_storage_metrics_free(rac_model_storage_metrics_t*)`. Null-safe, zeroes the struct so a second call is safe, and mirrors what `rac_storage_info_free` already does per element. The doc comment on `get_model_metrics` now names it and says explicitly that `rac_storage_info_free` is for the other call.

`bindings/swift/.../CRACommons/include/rac_storage_analyzer.h` is a hand-maintained flattened copy of the same header, so it carries the declaration too. The RN copy under `jniLibs/include/` is staged from `core/include` at build time and needs no edit.

## Not changed

`core/exports/RACommons.exports` is untouched. `rac_storage_analyzer_get_model_metrics` is not on that curated list either, and the list is only applied for `RAC_BUILD_SHARED` non-macOS builds, which every Apple preset is not. Adding the free function without the allocating function would be asymmetric; if the pair should be exported, that is a separate call.

## Verification

- `cmake --build` clean, `nm` finds `_rac_model_storage_metrics_free` in `librac_commons.a`
- `test_core --run-all`: 18 passed, 0 failed
- `clang-format` reports no new violations on the touched core files; the pre-existing ones in `storage_analyzer.cpp` are at lines 878, 1236, 1323, 1324 and 1355, none of them mine

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved storage metrics cleanup to prevent memory leaks.
  * Added safe handling for repeated or null cleanup requests.

* **Documentation**
  * Clarified ownership and cleanup requirements for returned storage metrics.

### DIFF
--- a/bindings/react-native/packages/core/cpp/bridges/StorageBridge.cpp
+++ b/bindings/react-native/packages/core/cpp/bridges/StorageBridge.cpp
@@ -213,15 +213,19 @@ std::optional<ModelStorageMetrics> StorageBridge::getModelStorageMetrics(
         return std::nullopt;
     }
 
+    // The three strings are heap copies owned by us from here on. The
+    // std::string assignments below allocate and can throw, so the release has
+    // to survive unwinding rather than sit at the end of the happy path.
+    struct MetricsOwner {
+        rac_model_storage_metrics_t* metrics;
+        ~MetricsOwner() { rac_model_storage_metrics_free(metrics); }
+    } owner{&cMetrics};
+
     ModelStorageMetrics metrics;
     metrics.modelId = cMetrics.model_id ? cMetrics.model_id : "";
     metrics.modelName = cMetrics.model_name ? cMetrics.model_name : "";
     metrics.localPath = cMetrics.local_path ? cMetrics.local_path : "";
     metrics.sizeOnDisk = cMetrics.size_on_disk;
-
-    // The three strings above are heap copies owned by the caller; the
-    // std::strings have taken their own copies by now.
-    rac_model_storage_metrics_free(&cMetrics);
 
     return metrics;
 }
--- a/bindings/react-native/packages/core/cpp/bridges/StorageBridge.cpp
+++ b/bindings/react-native/packages/core/cpp/bridges/StorageBridge.cpp
@@ -219,6 +219,10 @@ std::optional<ModelStorageMetrics> StorageBridge::getModelStorageMetrics(
     metrics.localPath = cMetrics.local_path ? cMetrics.local_path : "";
     metrics.sizeOnDisk = cMetrics.size_on_disk;
 
+    // The three strings above are heap copies owned by the caller; the
+    // std::strings have taken their own copies by now.
+    rac_model_storage_metrics_free(&cMetrics);
+
     return metrics;
 }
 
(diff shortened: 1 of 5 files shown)


Siddhesh2377:
Corrected. I had reasoned that the curated Apple export list should stay untouched because `rac_storage_analyzer_get_model_metrics` looked absent from it, and the `validation` job showed me that was wrong twice over: the allocating function *is* on the list at line 722, and `check_rac_api_exports.sh --strict` requires every `RAC_API` decl to be there.

`_rac_model_storage_metrics_free` is now listed next to its siblings. Both gates pass locally: exports reports "all 886 RAC_API-decorated decls are present", and `check_deprecated_surfaces.sh` reports no new regressions.

Siddhesh2377:
@coderabbitai good catch, taking it. The three `std::string` assignments allocate and can throw, and the free sat after them on the happy path only, so unwinding would have leaked exactly what this PR set out to stop leaking.

Replaced the trailing call with a scope-local owner declared right after the success check, so the release runs on the normal return and on unwinding:

```cpp
struct MetricsOwner {
    rac_model_storage_metrics_t* metrics;
    ~MetricsOwner() { rac_model_storage_metrics_free(metrics); }
} owner{&cMetrics};
```

String-copy behaviour is unchanged.

PR #712, opened 2026-08-16
feat(kotlin,flutter): put registry refresh on the models namespace

### LINKED ISSUE #639: models.refresh() is on the Swift namespace but not Kotlin's after #605
#605 aligned the public API across the SDKs around the namespace surface. One method looks like it did not make the crossing, and I want to check whether that is deliberate before anyone writes a patch.

Swift exposes registry refresh on the namespace:

```swift
// Public/API/Namespaces/ModelsNamespace.swift:347
public func refresh(
    rescanLocal: Bool = true,
    includeRemoteCatalog: Bool = false,
    pruneOrphans: Bool = false
) async
```

and that is the only public route there, since `performRegistryRefresh` is `internal` (`Public/Extensions/Models/RunAnywhere+ModelRegistry.swift:43`).

Kotlin's `ModelsNamespace` has `list`, `get`, `register`, `unregister`, `download`, `delete`, `load`, `unload` and `state`, but no `refresh`. The capability is not missing from the SDK: it is public on the older extension surface with the same three parameters.

```kotlin
// public/extensions/Models/RunAnywhereModelRegistry.kt
suspend fun RunAnywhere.refreshModelRegistry(
    rescanLocal: ..., includeRemoteCatalog: ..., pruneOrphans: ...
)
```

So a Kotlin caller can refresh; they just reach it by a different name on a different surface than a Swift caller. Flutter's namespace forwards to `RunAnywhereModels.shared.refreshModelRegistry()`, which looks like the shape Kotlin would take.

Which of these is intended?

1. The namespace migration is staged and Kotlin's `models.refresh()` is simply still to come.
2. Kotlin keeps `refreshModelRegistry` as the public spelling on purpose and the namespaces are not meant to match here.
3. It is an oversight in #605.

If it is 1 or 3 I am happy to send the forwarding method, matching Swift's parameter defaults. Not opening a PR first, since adding a public method is API surface and I would rather have the answer than guess.

Siddhesh2377:
Closes #639.

Takes option 1 from the issue: the namespace is the intended surface, so Kotlin and Flutter get `models.refresh()` and the old spelling is deprecated the way #605 deprecated its four siblings.

## What was asymmetric

Swift has `ModelsNamespace.refresh(rescanLocal:includeRemoteCatalog:pruneOrphans:)` and deprecated `RunAnywhere.refreshModelRegistry` with `renamed:` pointing at it.

Kotlin did neither. `ModelsNamespace` had no `refresh` in any spelling, and `refreshModelRegistry` was the only one of five extensions in `RunAnywhereModelRegistry.kt` left without a `ReplaceWith` deprecation. Flutter had the same gap: `RunAnywhereModels.shared.refreshModelRegistry()` is called from inside `list()` as a lazy dirty-flag rescan, not exposed as a namespace verb.

Credit to @josuediazflores for both corrections in the thread; they are what made the intended shape obvious.

## What this does

Kotlin gains `RunAnywhere.models.refresh(rescanLocal, includeRemoteCatalog, pruneOrphans)` with Swift's defaults (`true`, `false`, `false`), forwarding through a `legacyRefreshModelRegistry` helper in `LegacyBridge.kt` like every other namespace verb does rather than reaching for `RunAnywhere` directly. `refreshModelRegistry` picks up `@Deprecated(..., ReplaceWith(...))`, matching its four siblings.

Flutter's `ModelsApi` gains the same verb with the same defaults. It clears `_registryDirty` first so the explicit call and `list()`'s lazy rescan cannot double-refresh.

No behaviour change for existing callers: the deprecated spelling still works, and the namespace method forwards to exactly the same code path.

## Verification

- Kotlin: `:compileDebugKotlin`, `ktlintCheck` and `detekt` all pass. Detekt runs `maxIssues: 0` with warnings as errors. The first compile caught a missing import in `LegacyBridge.kt`, which is fixed here.
- Flutter: `dart analyze lib` on `packages/runanywhere` reports no issues.

## Summary by CodeRabbit

* **New Features**
  * Added model registry refresh APIs for Flutter and Kotlin.
  * Refresh the registry by rescanning local models, optionally including the remote catalog, and pruning orphaned entries.
  * Configurable options are available with sensible defaults for common refresh workflows.

* **Deprecations**
  * The legacy model registry refresh method now recommends using the newer models refresh API.

### DIFF
--- a/bindings/flutter/packages/runanywhere/lib/public/api/namespaces/models.dart
+++ b/bindings/flutter/packages/runanywhere/lib/public/api/namespaces/models.dart
@@ -68,6 +68,25 @@ class ModelsApi {
     return List<ModelInfo>.unmodifiable(result.models.models);
   }
 
+  /// Re-read the registry: rescan local storage, optionally pull the remote
+  /// catalog, optionally drop entries whose files are gone.
+  ///
+  /// Matches `models.refresh` on the Swift and Kotlin namespaces, including
+  /// its defaults. [list] does a lazy rescan of its own when the registry is
+  /// dirty; this is the explicit verb for the other two.
+  Future<void> refresh({
+    bool rescanLocal = true,
+    bool includeRemoteCatalog = false,
+    bool pruneOrphans = false,
+  }) async {
+    _registryDirty = false;
+    await RunAnywhereModels.shared.refreshModelRegistry(
+      rescanLocal: rescanLocal,
+      includeRemoteCatalog: includeRemoteCatalog,
+      pruneOrphans: pruneOrphans,
+    );
+  }
+
   /// The registry entry for [id], or null when it is unknown.
   Future<ModelInfo?> get(String id) async {
     final result = await RunAnywhereModels.shared.getModel(
--- a/bindings/flutter/packages/runanywhere/lib/public/api/namespaces/models.dart
+++ b/bindings/flutter/packages/runanywhere/lib/public/api/namespaces/models.dart
@@ -74,12 +74,18 @@ class ModelsApi {
   /// Matches `models.refresh` on the Swift and Kotlin namespaces, including
   /// its defaults. [list] does a lazy rescan of its own when the registry is
   /// dirty; this is the explicit verb for the other two.
+  ///
+  /// Deliberately leaves `_registryDirty` alone. `refreshModelRegistry` is a
+  /// no-op before the SDK is initialized and logs a native failure rather than
+  /// raising, so this method cannot tell a completed refresh from a skipped
+  /// one. Clearing the flag here would let a failed refresh cancel [list]'s
+  /// pending rescan and serve stale entries; leaving it costs at most one
+  /// redundant rescan.
   Future<void> refresh({
     bool rescanLocal = true,
     bool includeRemoteCatalog = false,
     bool pruneOrphans = false,
   }) async {
-    _registryDirty = false;
     await RunAnywhereModels.shared.refreshModelRegistry(
       rescanLocal: rescanLocal,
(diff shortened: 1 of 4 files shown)


Siddhesh2377:
@coderabbitai the diagnosis is right and I have fixed it, but the prescribed fix does not work, so I went a different way. Detail, since the difference matters.

You are correct that clearing `_registryDirty` optimistically can serve stale data. `refreshModelRegistry` returns early when `!DartBridge.isInitialized`, and on a native failure it does this:

```dart
final ok = await DartBridgeModelRegistry.instance.refresh(...);
if (!ok) {
  logger.warning('rac_model_registry_refresh_proto reported failure');
}
```

It logs and returns normally. That is the bug: my line cleared the flag whether or not a refresh actually happened, so a later `list()` would skip its lazy rescan and return stale entries.

Where the suggestion does not hold: "set `_registryDirty` to `false` only after successful completion" is not implementable from this method. Because `refreshModelRegistry` returns `Future<void>` and never raises, moving the assignment below the `await` clears the flag on exactly the same failing paths. It would look safer and behave identically. "Propagate refresh failure" would mean changing that method to throw or return a bool, which is a contract change on the capability surface and outside what this PR is for.

So the fix is to not manage `list()`'s memo from here at all. `refresh()` now performs the refresh and leaves the flag alone. Worst case is one redundant rescan on the next `list()`; there is no path that reports stale data as fresh.

One thing you did not flag that is worth recording: `list()` itself has the same shape at lines 61-62, clearing the flag before awaiting the same non-raising call. That predates this PR and I have left it, since fixing it changes existing behaviour on the failure path. It should be its own change.

`dart analyze lib` is clean.

PR #720, opened 2026-08-16
feat(react-native): put registry refresh on the models namespace

ayaangazali:
## What is wrong

React Native has no public way to refresh the model registry.

`#605` aligned the public API across all eight SDKs and gave Swift `models.refresh(rescanLocal:includeRemoteCatalog:pruneOrphans:)` (`ModelsNamespace.swift:389`). Kotlin and Flutter were missed; `#712` closed those two yesterday. RN is the one left.

## Evidence

Every piece of the native path is already there on RN:

- `specs/RunAnywhereCore.nitro.ts:274` declares `refreshModelRegistry(includeRemoteCatalog, rescanLocal, pruneOrphans)`
- `cpp/HybridRunAnywhereCore+Registry.cpp:443` encodes those three flags into a `ModelRegistryRefreshRequest` (fields 1/2/3, matching `idl/model_types.proto:411`) and runs the same discovery pre-pass Swift does
- `Public/Extensions/Models/RunAnywhere+ModelRegistry.ts:1046` wraps it with the same defaults Swift uses, and its own comment says "Matches Swift"

Nothing surfaces it. `src/index.ts:13` exports only the `RunAnywhere` object, that object is namespaces plus `initialize`/`reset`/`capabilities`, and `models` (`Public/Api/Models.ts:194`) has no `refresh`. `storage` (`Public/Api/Platform.ts:25`) has no refresh either, which is where the Web SDK put its equivalent.

So `refreshModelRegistry` is reachable from inside the package and from nowhere else.

This is not a documented omission: `capabilities()` lists the things RN deliberately does not ship (`agents`, `wakeword`, `realtime`) with a reason attached to each. Registry refresh is not among them.

The practical effect is that `models.list()` reads the registry as it stands and never rescans, so an artifact that changed on disk out from under the SDK stays wrong for the life of the process.

## What this does

Adds `models.refresh(options?)`, forwarding to the existing extension function. Forwarding rather than reimplementing keeps the non-throwing best-effort behaviour and the three defaults in one place, and matches how the rest of the namespace delegates.

Placed after `state()` to match Swift's ordering.

## Verification

Ran on this branch:

- `tsc --noEmit` for `@runanywhere/core`: clean
- `tsc -p tsconfig.test.json --noEmit`: clean
- `eslint src/Public/Api/Models.ts --max-warnings 0`: clean
- `jest` for `@runanywhere/core`: 75 passed, 15 suites, 0 failed

The two typechecks needed the full generated proto tree, so I ran `idl/codegen/generate_ts.sh`, `generate_ts_convenience.py`, `generate_defaults_pool.py` and `generate_streams.sh` first. None of that output is committed here; `git status` shows only the one file.

Also compiled a scratch caller against the public entry point to confirm the verb is actually reachable and the option shapes are right:

```ts
import { RunAnywhere } from '../src';
await RunAnywhere.models.refresh();
await RunAnywhere.models.refresh({ rescanLocal: false, pruneOrphans: true });
```

No test added: this is a one-line delegation to a function that already has its behaviour defined elsewhere, and a test here would assert that a forwarder forwards.

I checked the C++ flag ordering against the proto field numbers while I was in there, in case the two adjacent booleans had been swapped across the bridge. They line up correctly.

## Summary by CodeRabbit

- **New Features**
  - Added a model registry refresh capability for React Native applications.
  - Supports optional local model rescanning, remote catalog inclusion, and removal of orphaned entries.
  - Refresh operations return a promise so applications can track completion.
  - If a refresh fails, the existing model registry remains unchanged.

### DIFF
--- a/bindings/react-native/packages/core/src/Public/Api/Models.ts
+++ b/bindings/react-native/packages/core/src/Public/Api/Models.ts
@@ -24,6 +24,7 @@ import {
   getModel as getModelProto,
   listModels,
   queryModels,
+  refreshModelRegistry,
   registerArchiveModel,
   registerModel as registerUrlModel,
   registerMultiFileModel,
@@ -458,6 +459,25 @@ export const models = {
       storageUsedBytes: Number(storage?.app?.totalBytes ?? 0),
       storageFreeBytes: Number(storage?.device?.freeBytes ?? 0),
     };
+  },
+
+  /**
+   * Rescan managed model directories and reconcile downloaded state.
+   *
+   * Matches `models.refresh` on the Swift, Kotlin, and Flutter namespaces,
+   * including its defaults. Non-throwing like the others: a failed refresh
+   * leaves the current registry contents in place.
+   *
+   * `list()` reads the registry as it stands and never rescans, so this is
+   * the only way to pick up artifacts that changed on disk out from under
+   * the SDK.
+   */
+  refresh(options?: {
+    rescanLocal?: boolean;
+    includeRemoteCatalog?: boolean;
+    pruneOrphans?: boolean;
+  }): Promise<void> {
+    return refreshModelRegistry(options);
   },
 };
 


ayaangazali:
Correction to one line in the description. I wrote that React Native is the only SDK still missing this; that is not quite right, and I would rather say so than have you find it.

Checking the remaining bindings after opening this: Electron has it as `assets.refresh` (`src/api/assets.ts:563`), Web has it as `storage.refresh`, and Python has it nowhere. Python is a different situation from RN though, which is why it is not in this PR: RN already had the whole native path and simply never surfaced it, whereas `bindings/python/native/module.cpp` has no binding for `rac_model_registry_refresh_proto` at all, so Python needs native work rather than a forwarder. Happy to open that separately if you want it.

The claim this PR actually rests on is unchanged: RN is the only SDK where the capability is fully wired natively and unreachable from the public API.

ayaangazali:

Still reproduces on `main`: `bindings/react-native/packages/core/src/Public/Api/Models.ts` has no `refresh` member at all, so `RunAnywhere.models.refresh(...)` is a runtime `undefined` on React Native while Swift and Kotlin both expose it on the same namespace.

One nuance I confirmed since opening this, in case it looks like an inconsistency: Web deliberately does not have `models.refresh` either, because it puts the equivalent on `storage.refresh()` and documents that namespace as "Not part of the cross-SDK spec" (re-reading OPFS is a browser persistence concern there). So this is React Native being the odd one out among the SDKs that do model this as a registry operation, not a four-way gap.

Happy to rework it if you would rather this verb not spread further.

ayaangazali:
Still reproduces on `45e91276e` (0.20.36), and I re-checked the parity claim across all five SDKs rather than restating it.

React Native's `models` namespace exposes `list`, `register`, `download`, `delete`, `load`, `unload`, `unloadAll`, `unregister` and no `refresh`. Everywhere else it exists:

```
swift   bindings/swift/.../Namespaces/ModelsNamespace.swift:389   public func refresh(
kotlin  bindings/kotlin/.../public/api/ModelsNamespace.kt:292     public suspend fun refresh(
web     bindings/web/.../Public/Extensions/RunAnywhere+ModelRegistry.ts:51  refresh(options?: RefreshOptions)
flutter bindings/flutter/.../core/native/rac_native.dart:1685     rac_model_registry_refresh_proto bound
```

So the commons verb is `rac_model_registry_refresh_proto`, it is shipped, and four bindings reach it. This is the one that does not, which is what makes it an omission rather than a staged rollout.

Why it matters beyond symmetry: `list()` reads the registry as it stands and never rescans, so on React Native there is currently no way to pick up artifacts that changed on disk underneath the SDK. The capability is not merely unexposed, it is unreachable.

The addition mirrors the Swift signature and its defaults, and is non-throwing like the others: a failed refresh leaves the registry contents in place.

Merges clean, CI green.

sanchitmonga22:
@coderabbitai review

sanchitmonga22:
@coderabbitai full review

sanchitmonga22:
> Replying to [this comment](https://github.com/RunanywhereAI/runanywhere-sdks/pull/720#issuecomment-5308838262)

That makes sense — thanks for the correction. The narrowed claim holds: before this PR, RN's `Models.ts` had no `refresh`, while Swift (`ModelsNamespace.swift:389`), Kotlin (`ModelsNamespace.kt:292`) and Flutter (`namespaces/models.dart:84`) already exposed `models.refresh`. On Python: yes please, a separate PR wiring `rac_model_registry_refresh_proto` into `models.refresh` would be welcome. It's part of the follow-up issue on aligning refresh across Web, Python and Electron, which will tag you.

sanchitmonga22:
> Replying to [this comment](https://github.com/RunanywhereAI/runanywhere-sdks/pull/720#issuecomment-5388040740)

Thanks for checking. Keeping `refresh` is right. It's already the verb on the models namespace in Swift (`ModelsNamespace.swift:389`), Kotlin (`ModelsNamespace.kt:292`) and Flutter (`namespaces/models.dart:84`), with the same defaults, so RN now matches the existing name rather than spreading a new one. No rework needed; treating this as addressed.

sanchitmonga22 (approved):
Thanks a lot for this, @ayaangazali! React Native can now rescan the model registry from the public API: `RunAnywhere.models.refresh()` picks up model files that changed on disk outside the SDK, with the same defaults as Swift, Kotlin and Flutter. Thanks as well for tightening the cross-SDK claim in the thread.

Checked: CodeRabbit reviewed the latest commit · CI green · built and linted locally merged into `main` (react-native: typecheck, lint, core and example tests) · two independent code reviews.

One optional note, nothing needed here: the helper this exposes awaits the throwing `ensureServicesReady()` inside its `try` (`RunAnywhere+ModelRegistry.ts:1064`). If services init fails, the local rescan is skipped, while Swift and Kotlin still rescan. It predates this PR, so it's written up as a follow-up below.

Follow-ups, not blocking: #852 (refresh should still rescan when services init fails) and #853 (align refresh across Web, Python and Electron) — you're welcome to pick these up.

PR #721, opened 2026-08-16
fix(web): stop AsyncQueue.fail from ending a parked consumer normally

ayaangazali:
## What is wrong

`AsyncQueue.fail()` resolved a parked consumer with `{ done: true }` and only stored the error:

```ts
fail(error: Error): void {
  if (this.done) return;
  this.done = true;
  this.error = error;
  if (this.resolveNext) {
    const r = this.resolveNext;
    this.resolveNext = null;
    r({ value: undefined as unknown as T, done: true });   // <- normal end
  }
}
```

The stored `this.error` is only ever read by `next()`. A consumer that was parked has already been handed a normal end-of-stream, its `for await` loop has exited, and it never calls `next()` again. So whether a caller sees the failure or a clean completion depends on whether it happened to be awaiting at that instant.

During streaming it almost always is. The consumer drains tokens faster than inference produces them, so parked is the normal state, and the silent path is the common one rather than the edge case.

## Why it matters

`streamGenerate` fails both queues on a mid-stream error (`RunAnywhere+TextGeneration.ts:183-184`). With the queue swallowing it, iterating `stream` or `events` ends as if generation had completed: the caller keeps a truncated answer and is told nothing went wrong.

The error does survive on the separate `result` promise, but `LLMStreamingResult` exposes `stream`, `events` and `result` independently, and a caller driving the token stream has no obligation to await `result`.

## What this changes

Producers now only signal, and `next()` re-reads the queue state, so a parked consumer raises the error on the same check a late consumer already hit.

That is exactly how `AsyncQueue` in `@runanywhere/proto-ts/streams/push` (the shared implementation used by React Native and Electron) already behaves. This file is a hand-rolled second copy that drifted from it, so this brings the clone back in line rather than inventing a third behaviour.

Buffered values still drain before the error surfaces, and `complete()` still ends the stream normally. Net 25 insertions, 24 deletions.

## Verification

Ran in `bindings/web/packages/core`:

- `vitest run`: **57 files, 248 tests, all pass** (including the 4 new ones)
- `tsc --noEmit`: clean
- `eslint src/Foundation/AsyncQueue.ts tests/unit/Foundation/AsyncQueue.test.ts --max-warnings 0`: clean

The new test is in `tests/unit/Foundation/AsyncQueue.test.ts`. I confirmed it is a real regression test by stashing only `AsyncQueue.ts` and re-running: the parked-consumer case fails on unfixed source with

```
- Error { "message": "rejected promise" }
+ { "done": true, "value": undefined }
```

and the other three pass on unfixed source, so they pin behaviour this change had to preserve rather than describing the fix.

The suite and typecheck need the generated proto tree, so I ran `idl/codegen/generate_ts.sh`, `generate_ts_convenience.py`, `generate_defaults_pool.py` and `generate_streams.sh` first. None of that output is committed; `git status` shows only the two files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved asynchronous queue handling for waiting consumers.
  * Ensured queued values are delivered before reported failures.
  * Improved behavior for consumers after completion or failure.
  * Reduced the risk of missed notifications when new values, completion, or errors become available.
  * Preserved existing public queue interfaces while improving result delivery consistency.

* **Tests**
  * Added coverage for queue failures, buffered values, parked consumers, and normal completion.

### DIFF
--- a/bindings/web/packages/core/src/Foundation/AsyncQueue.ts
+++ b/bindings/web/packages/core/src/Foundation/AsyncQueue.ts
@@ -25,31 +25,34 @@
 /** Async queue with a single producer + single consumer (for-await). */
 export class AsyncQueue<T> implements AsyncIterable<T> {
   private buffer: T[] = [];
-  private resolveNext: ((v: IteratorResult<T>) => void) | null = null;
+  private wake: (() => void) | null = null;
   private done = false;
   private error: Error | null = null;
+
+  /**
+   * Wake a parked consumer. Producers only ever signal; `next()` re-reads the
+   * queue state itself, which is what keeps `fail()` from having to resolve a
+   * waiter it cannot reject. Mirrors `AsyncQueue` in
+   * `@runanywhere/proto-ts/streams/push`.
+   */
+  private signal(): void {
+    const wake = this.wake;
+    this.wake = null;
+    if (wake) wake();
+  }
 
   /** Producer: push the next value. Discarded if the queue is closed. */
   push(value: T): void {
     if (this.done) return;
-    if (this.resolveNext) {
-      const r = this.resolveNext;
-      this.resolveNext = null;
-      r({ value, done: false });
-    } else {
-      this.buffer.push(value);
-    }
+    this.buffer.push(value);
+    this.signal();
   }
 
   /** Producer: signal normal end-of-stream. Idempotent. */
   complete(): void {
     if (this.done) return;
     this.done = true;
-    if (this.resolveNext) {
-      const r = this.resolveNext;
-      this.resolveNext = null;
-      r({ value: undefined as unknown as T, done: true });
-    }
+    this.signal();
   }
 
   /** Producer: signal abnormal termination. Next consumer await throws. */
@@ -57,22 +60,20 @@ export class AsyncQueue<T> implements AsyncIterable<T> {
     if (this.done) return;
     this.done = true;
     this.error = error;
-    if (this.resolveNext) {
-      const r = this.resolveNext;
-      this.resolveNext = null;
-      r({ value: undefined as unknown as T, done: true });
-    }
+    this.signal();
   }
 
   [Symbol.asyncIterator](): AsyncIterator<T> {
     return {
-      next: (): Promise<IteratorResult<T>> => {
-        if (this.buffer.length > 0) {
-          return Promise.resolve({ value: this.buffer.shift()!, done: false });
+      next: async (): Promise<IteratorResult<T>> => {
+        for (;;) {
+          if (this.buffer.length > 0) {
+            return { value: this.buffer.shift()!, done: false };
+          }
+          if (this.error) throw this.error;
+          if (this.done) return { value: undefined as unknown as T, done: true };
+          await new Promise<void>((r) => { this.wake = r; });
         }
-        if (this.error) return Promise.reject(this.error);
-        if (this.done) return Promise.resolve({ value: undefined as unknown as T, done: true });
-        return new Promise((r) => { this.resolveNext = r; });
       },
     };
   }
--- a/bindings/web/packages/core/tests/unit/Foundation/AsyncQueue.test.ts
+++ b/bindings/web/packages/core/tests/unit/Foundation/AsyncQueue.test.ts
@@ -1,0 +1,53 @@
+/**
+ * AsyncQueue.test.ts
+ *
+ * `fail()` has to reach the consumer whether or not it is parked. A streaming
+ * consumer is parked on `next()` almost all the time (it drains tokens faster
+ * than inference produces them), so a failure that only surfaces on a later
+ * `next()` is a failure the caller never sees.
+ */
+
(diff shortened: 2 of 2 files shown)


ayaangazali:

Still reproduces on `main` (`bindings/web/packages/core/src/Foundation/AsyncQueue.ts:56`):

```ts
  fail(error: Error): void {
    if (this.done) return;
    this.done = true;
    this.error = error;
    if (this.resolveNext) {
      const r = this.resolveNext;
      this.resolveNext = null;
      r({ value: undefined as unknown as T, done: true });   // <- resolves, never throws
    }
  }
```

A consumer already parked in `next()` gets a clean end-of-stream instead of the error. Only a consumer that happens to call `next()` after `fail()` reaches the `if (this.error) return Promise.reject(...)` branch, and for a token queue the parked case is the normal one.

One thing worth raising rather than sitting on, since it may change what you want here: this file duplicates a class that already exists and already gets this right. `@runanywhere/proto-ts/streams/push` exports an `AsyncQueue` whose `next()` drains the buffer, then throws a retained failure, then reports done. `bindings/web/packages/core` already depends on that package and already imports `iterableFromSubscription` from that exact module in `Foundation/EventBus.ts:41`, and Electron consumes the shared class directly in `src/api/iter.ts`.

So there is a smaller change available than this one: delete the 79-line web copy and point the four importers (`stt.ts`, `vad.ts`, `voice.ts`, `RunAnywhere+TextGeneration.ts`) at the shared class. That fixes this bug by removing the code containing it, and also gives web the `return()` the local copy lacks.

I did not do that here because it is a different change from the one you have been sent, and I did not want to swap it under you. Say the word and I will convert this PR to the deletion, or keep this as the minimal fix.

ayaangazali:
Still reproduces on `45e91276e` (0.20.36), and the clearest statement of the bug is the doc comment sitting two lines above it.

```ts
/** Producer: signal abnormal termination. Next consumer await throws. */
fail(error: Error): void {
  if (this.done) return;
  this.done = true;
  this.error = error;
  if (this.resolveNext) {
    const r = this.resolveNext;
    this.resolveNext = null;
    r({ value: undefined as unknown as T, done: true });   // clean end, not a throw
  }
}
```

"Next consumer await throws" holds only for a consumer that arrives *after* `fail()`, which takes the `if (this.error) return Promise.reject(this.error)` path in `next()`. A consumer already parked on the promise when `fail()` lands is resolved with `{ done: true }` and sees a normal end of stream. The error is stored in `this.error` and then never delivered to it.

So whether a stream failure surfaces as an error or as a clean completion depends on a race: whether the consumer happened to be awaiting at the moment the producer failed. That is the worst version of this, because the common case in tests (consume after failure) works and the common case in production (consumer parked, waiting on the model) does not.

Merges clean, CI green.

sanchitmonga22:
@coderabbitai review

sanchitmonga22:
@coderabbitai full review

sanchitmonga22:
> Replying to [this comment](https://github.com/RunanywhereAI/runanywhere-sdks/pull/721#issuecomment-5388040833)

Thanks for flagging this, and sorry it sat without an answer! Let's land this PR as the minimal fix and do the deletion separately. You're right that `bindings/proto-ts/src/streams/push.ts:305` already has the correct queue, but the swap isn't drop-in. The shared class makes `[Symbol.asyncIterator]` a `readonly` own property (`push.ts:315`), while `RunAnywhere+TextGeneration.ts:191` reassigns it to start the producer lazily. It also throws a stored failure once and then reports done, where the Web queue re-throws it.

Two follow-ups, neither blocking this PR: #855 for a loose end this fix surfaces (`generateStream`'s `result` promise can go unobserved when the stream throws), and #856 for replacing the Web queue with the shared one. You're welcome to take either.

sanchitmonga22:
> Replying to [this comment](https://github.com/RunanywhereAI/runanywhere-sdks/pull/721#issuecomment-5555940567)

Thanks for the ping, and for re-checking it against the newer `main`! Your read of the race is right: a consumer already waiting on `next()` got `{ done: true }` from `fail()`, and only one that arrived afterwards saw the error. We reviewed 1851dff: CodeRabbit is clean, CI is green, and it builds and passes the Web tests merged into `main`, so it's lined up to merge. I answered your earlier question about the duplicate queue in a separate reply.

sanchitmonga22 (approved):
Thanks a lot for this, @ayaangazali! When `generateStream` fails partway through, a consumer waiting on the next token now gets the error thrown into its `for await` loop, instead of the stream ending as if generation had finished. Having producers only wake the consumer while `next()` re-reads the state is the same pattern the shared proto-ts queue uses, and the four tests pin the race exactly.

Checked: CodeRabbit reviewed the latest commit · CI green · built and linted locally merged into `main` (web: typecheck, lint, 248 unit tests) · two independent code reviews.

Follow-ups, not blocking: #855 (`generateStream`'s `result` promise can go unobserved now that the stream throws) and #856 (replacing the Web `AsyncQueue` with the shared one) — you're welcome to pick these up.

PR #722, opened 2026-08-16
fix(web): keep a stream failure that lands while nobody is parked on next()

ayaangazali:
## What is wrong

`OffscreenRuntimeBridge.getStreamIterator` and `streamCallback` (`ProtoAdapterTypes`) both fail like this:

```ts
const fail = (err: unknown): void => {
  if (finished) return;
  finished = true;
  while (waiters.length > 0) waiters.shift()!.reject(err);
};
```

Only the waiters that are parked at that instant learn about the error, and nothing keeps it afterwards. `next()` drains the queue and then returns `{ done: true }` because `finished` is set. So a failure that arrives while the consumer is between `next()` calls is discarded, and the consumer is told the stream ended normally.

## Why both windows are ordinary

**Worker bridge.** The worker posts `error` whenever it likes, including while the consumer is rendering the token it just received. This is the documented failure here: a WASM OOM in the VLM path kills the worker (see the Web VLM Worker crash-recovery note in AGENTS.md).

**streamCallback.** `start()` defers the native call by a microtask, on purpose, so the consumer's first `next()` parks before native code starts emitting. The first emit wakes it; everything emitted after that is buffered. A call that emits and then throws therefore fails with no waiter left to reject.

The consequence in both cases is the same, and it is the one that costs a user real time: inference dies, and the app renders a truncated answer as a completed one.

## What this changes

Each iterator stores the error and raises it from `next()` on a check placed between the existing queue drain and the `finished` check. Buffered events still arrive first, and only once they are exhausted does the error surface. Cancellation and normal completion are untouched.

+21 lines of source across the two files.

## Verification

Ran in `bindings/web/packages/core`:

- `vitest run`: **57 files, 246 tests, all pass**
- `tsc --noEmit`: clean
- `eslint` on the three changed files with `--max-warnings 0`: clean

The new test is `tests/unit/runtime/StreamFailureNotParked.test.ts`. Both cases are deterministic rather than timing-dependent: the fake worker only crashes when the test tells it to, and the fake module emits a second value so the buffered-events window is a fact of the test rather than a race.

I confirmed both are real regression tests by stashing only the two source files and re-running. Both fail on unfixed source with the same symptom:

```
× is raised by the worker bridge rather than read as end-of-stream
  → promise resolved "{ value: undefined, done: true }" instead of rejecting
× is raised by streamCallback rather than read as end-of-stream
  → promise resolved "{ value: undefined, done: true }" instead of rejecting
```

The suite and typecheck need the generated proto tree, so I ran `idl/codegen/generate_ts.sh`, `generate_ts_convenience.py`, `generate_defaults_pool.py` and `generate_streams.sh` first. None of that output is committed; `git status` shows only the three files.

One note on scope: this is the same failure mode as #721, but a different one of the two races and a disjoint set of files, so I kept them apart rather than growing that diff. #721 loses the error when the consumer **is** parked; this loses it when the consumer **is not**. They can land in either order.

## Summary by CodeRabbit

* **Bug Fixes**
  * Stream errors occurring between successive reads are now correctly reported.
  * Subsequent stream requests reject with the original failure instead of incorrectly signaling the end of the stream.
  * Cancelling a stream clears pending errors and completes normally.
  * Streaming consumers receive consistent failure behavior across supported runtime environments.

* **Tests**
  * Added coverage for deferred stream failures across worker-based and native streaming scenarios, including cancellation behavior.

### DIFF
--- a/bindings/web/packages/core/src/Adapters/ProtoAdapterTypes.ts
+++ b/bindings/web/packages/core/src/Adapters/ProtoAdapterTypes.ts
@@ -549,6 +549,7 @@ export function streamCallback<T>(
       let callbackPtr = 0;
       let started = false;
       let finished = false;
+      let failure: unknown = null;
       let callActive = false;
       let emitsSinceYield = 0;
 
@@ -571,6 +572,10 @@ export function streamCallback<T>(
       const fail = (error: unknown): void => {
         if (finished) return;
         finished = true;
+        // Kept for the consumer that is not parked right now (a decode that
+        // throws while earlier events are still buffered, say); without it the
+        // next `next()` would read the stream as a clean end.
+        failure = error;
         while (waiters.length > 0) {
           waiters.shift()!.reject(error);
         }
@@ -699,6 +704,11 @@ export function streamCallback<T>(
           start();
           if (queue.length > 0) {
             return Promise.resolve({ value: queue.shift()!, done: false });
+          }
+          if (failure !== null) {
+            const error = failure;
+            failure = null;
+            return Promise.reject(error);
           }
           if (finished) {
             return Promise.resolve({ value: undefined as T, done: true });
--- a/bindings/web/packages/core/src/Adapters/ProtoAdapterTypes.ts
+++ b/bindings/web/packages/core/src/Adapters/ProtoAdapterTypes.ts
@@ -721,6 +721,11 @@ export function streamCallback<T>(
           try {
             onCancel?.();
           } finally {
+            // An explicit cancel outranks an error the consumer never asked
+            // about: `finish()` is a no-op once `fail()` has set `finished`,
+            // so drop the retained failure here rather than replaying it at
+            // the next `next()`.
+            failure = null;
             finish();
           }
           return Promise.resolve({ value: undefined as T, done: true });
--- a/bindings/web/packages/core/src/Adapters/ProtoAdapterTypes.ts
+++ b/bindings/web/packages/core/src/Adapters/ProtoAdapterTypes.ts
@@ -549,7 +549,9 @@ export function streamCallback<T>(
       let callbackPtr = 0;
       let started = false;
       let finished = false;
-      let failure: unknown = null;
+      // Boxed: a bare `unknown` sentinel cannot tell "rejected with null"
+      // apart from "never failed", and that reason would be dropped.
+      let failure: { reason: unknown } | null = null;
       let callActive = false;
       let emitsSinceYield = 0;
 
@@ -575,7 +577,7 @@ export function streamCallback<T>(
         // Kept for the consumer that is not parked right now (a decode that
         // throws while earlier events are still buffered, say); without it the
         // next `next()` would read the stream as a clean end.
-        failure = error;
+        failure = { reason: error };
         while (waiters.length > 0) {
           waiters.shift()!.reject(error);
         }
@@ -706,7 +708,7 @@ export function streamCallback<T>(
             return Promise.resolve({ value: queue.shift()!, done: false });
           }
           if (failure !== null) {
-            const error = failure;
+            const error = failure.reason;
             failure = null;
             return Promise.reject(error);
           }
--- a/bindings/web/packages/core/src/runtime/BackendWorkerHost.ts
+++ b/bindings/web/packages/core/src/runtime/BackendWorkerHost.ts
@@ -73,6 +73,13 @@ interface StreamPending {
     reject(reason: unknown): void;
   }>;
   finished: boolean;
+  /**
+   * Set by whichever path failed the stream. Rejecting the parked waiters is
+   * not enough on its own: a worker that dies while the consumer is running its
+   * loop body has no waiter to reject, and `next()` would otherwise read
+   * `finished` as a clean end of stream.
+   */
+  failure?: unknown;
 }
 
 type PendingRequest = UnaryPending | StreamPending;
@@ -223,6 +230,7 @@ export class BackendWorkerHost {
     const fail = (error: unknown): void => {
       if (state.finished) return;
       state.finished = true;
+      state.failure = error;
       this.pending.delete(requestId);
       this.activeStreamIds.delete(requestId);
       while (state.waiters.length) state.waiters.shift()!.reject(error);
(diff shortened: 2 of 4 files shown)


ayaangazali:
Both review points were fair. Fixed in de80634.

**Cancellation vs a retained failure.** Correct, and it was a wart my own change introduced. `return()` routes through `finish()`, which is a no-op once `fail()` has set `finished`, so the stored error survived the cancel and the next `next()` replayed it. A consumer that walked away should not be handed an error it never asked to read, so `return()` now clears the retained failure in both iterators.

I added a third test for it. It fails without the new line, replaying `worker died` after the cancel instead of ending the stream.

**Hand-written protocol values in the fake worker.** The casts were the real problem and they are gone. The messages were already structurally valid `WorkerRequest` / `WorkerResponse` values, so narrowing on the `init` / `cancel` discriminators and passing `requestId ?? undefined` types the fake without any `as unknown as`. `tsc --noEmit` is clean, which is the part that keeps the fake honest if the protocol changes.

On the letter of the guideline: these two are hand-written TypeScript discriminated unions in `src/runtime/StreamWorker.ts` rather than generated proto enums, so there is no generated symbol to import here. The existing fake in `tests/unit/runtime/StreamWorker.test.ts` builds its messages the same way. If you would rather these discriminators became exported constants, say so and I will do it in a separate PR that converts both fakes together, since it touches that test too.

Gates after the change, in `bindings/web/packages/core`: `vitest run` 57 files / 247 tests pass, `tsc --noEmit` clean, `eslint --max-warnings 0` clean on all three files.

ayaangazali:
Rebased on main (0.20.24) and added a third site, d491964.

I went back and swept the rest of the hand-rolled iterators instead of stopping at the two I started with, and `BackendWorkerHost` has the same shape: `stream()`'s local `fail` and `failRequest()` both set `finished` and reject only the parked waiters, keeping no record, so the following `next()` returns `{ done: true }`.

That one is the worst of the three. `failRequest` is what `handleCrash` calls for every in-flight request, and a llamacpp WebGPU worker `Abort()` after a successful load is a crash this file already documents and handles. So today a worker death ends every stream whose consumer happens to be mid-body as though generation finished, which is exactly the case the crash handling exists to report.

Retention had to live on the shared `StreamPending` rather than a closure variable, since `failRequest` fails the stream from outside `stream()`. `return()` clears it, so the cancel rule you asked for holds here too.

Verified the same way: with only the two retention lines commented out, the new test fails with

```
× is raised by the backend worker host rather than read as end-of-stream
  → promise resolved "{ value: undefined, done: true }" instead of rejecting
```

Gates on the rebased branch, in `bindings/web/packages/core`: `vitest run` 57 files / 248 tests pass, `tsc --noEmit` clean, `eslint --max-warnings 0` clean.

Happy to split the `BackendWorkerHost` change out if you would rather review it separately, but it is the same defect and the same one-line fix, so I kept the sweep in one place.

ayaangazali:

ayaangazali:

Still reproduces on `main`. `fail()` in `OffscreenRuntimeBridge.ts:344` only reaches consumers who happen to be parked:

```ts
        const fail = (err: unknown): void => {
          if (finished) return;
          finished = true;
          this.pending.delete(requestId);
          while (waiters.length > 0) waiters.shift()!.reject(err);
        };
```

and `next()` reports a clean end once `finished` is set:

```ts
            if (finished) {
              return Promise.resolve({ value: undefined as T, done: true });
            }
```

So a worker that dies while the consumer is inside its loop body has no waiter to reject, `finished` flips, and the next `next()` returns `{ done: true }`. The stream looks like it ended normally and the error is gone. That is the case this PR retains the failure for.

ayaangazali:
Still reproduces on `45e91276e` (0.20.36). Companion to #721 and worth reading together, though they are separate defects in separate files.

#721 is the queue primitive: `AsyncQueue.fail()` resolves a parked consumer with `{ done: true }` instead of throwing, so a failure that lands while someone is awaiting is downgraded to a clean end.

This PR is the layer above, in `BackendWorkerHost` / `OffscreenRuntimeBridge`: a stream failure arriving while nobody is parked is dropped rather than retained, so the next consumer to attach sees a stream that simply ended.

The two together mean a worker-side failure is lost in both directions, whether or not a consumer happens to be waiting at that instant. Either fix alone still leaves the other window open, which is why they are not folded into one PR.

The four files here are `ProtoAdapterTypes.ts`, `BackendWorkerHost.ts`, `OffscreenRuntimeBridge.ts` and a regression test (`StreamFailureNotParked.test.ts`) that pins the not-parked window specifically.

Merges clean, CI green.

ayaangazali:
Good catch, fixed in the push above, and it applied more widely than the one line you flagged.

The retained failure was using the absence of a value as its own "no failure" sentinel. `BackendWorkerHost` had `failure?: unknown` tested against `undefined`; the other two iterators this PR touches had `let failure: unknown = null` tested against `null`. In each case a producer that rejects with exactly the sentinel value is indistinguishable from one that never failed, so `next()` reports a clean end-of-stream. Same defect, three places, so I boxed the value into `{ reason }` in all three rather than only where you spotted it.

Two things I want to be straight about rather than overclaim.

**I could not find a reachable production path.** In `BackendWorkerHost` the worker error path always wraps the message (`new BackendWorkerError(response.message)`), so the only way in is a worker factory or `init()` rejecting with literally `undefined`. I did not find a caller that does that. So this is a type-level hole that the code cannot currently reach, not a live bug like the one the PR itself fixes. I still think it is worth closing, because the invariant "a sentinel value can never be a real reason" is cheaper to hold than to keep re-checking, but I did not want to imply I had reproduced it.

**I could not run the web tests locally.** Building the `@runanywhere/proto-ts` workspace that `packages/core` imports pulls in the root workspace, which fails resolution on something unrelated to this change:

```
npm error code ERESOLVE
npm error While resolving: runanywhere-ai-example@0.1.0
npm error Found: react-native-worklets@0.9.3
npm error   bindings/react-native/example
```

So I did not add a regression test for the sentinel case: I will not ship a test I cannot execute. The four existing tests in `StreamFailureNotParked.test.ts` cover the reachable failure, and the Web job in CI is what will confirm the boxing typechecks. If it comes back red I will fix it rather than leave it.

sanchitmonga22:
@coderabbitai review

sanchitmonga22:
@coderabbitai full review

sanchitmonga22:
> Replying to [this comment](https://github.com/RunanywhereAI/runanywhere-sdks/pull/722#issuecomment-5310442729)

That works for us — thanks for offering! Go ahead and open that as a separate PR whenever you get a chance; converting both fakes together sounds right. Treating this as addressed.

sanchitmonga22:
> Replying to [this comment](https://github.com/RunanywhereAI/runanywhere-sdks/pull/722#issuecomment-5318244211)

That makes sense — thanks for explaining. It's the same defect and the same one-line fix, so keeping the sweep in one place made this easier to review, not harder. Treating this as addressed.

sanchitmonga22 (approved):
Thanks a lot for this, @ayaangazali! A web SDK stream that fails while the consumer is between `next()` calls now reports the error, after any buffered events, instead of silently looking like a completed stream — covering the main-thread `streamCallback`, `BackendWorkerHost` crash/dispose paths, and `OffscreenRuntimeBridge`.

Checked: CodeRabbit reviewed the latest commit · CI green · built and linted locally merged into `main` (web — bindings/web/packages/core) · two independent code reviews.

Follow-up, not blocking: #859 (stop re-delivering an already-parked failure to a later `next()`) — you're welcome to pick it up if you're interested.

PR #728, opened 2026-08-17
fix(web): keep proto-ts installable after the workspace npm ci

MohammedAlkindi:
## Description

On a clean clone, the documented Web SDK developer flow fails before any code is touched: after IDL codegen, running `npm ci` in `bindings/web` and then `npm run build:ts` (or `typecheck`) dies with 39x `TS2307: Cannot find module '@bufbuild/protobuf/wire'`. `bindings/proto-ts` is a workspace member of the `bindings/web` npm workspace but sits *outside* the workspace root, so the workspace install hoists proto-ts's dependencies into `bindings/web/node_modules` — a directory Node's resolution from `bindings/proto-ts/src` can never reach — and leaves `bindings/proto-ts` without its own `node_modules`.

This is a known trap — CI already carries the workaround: the `web-typecheck` job in `pr-build.yml` has an explicit "Install proto-ts dependencies (standalone install)" step (`npm ci --workspaces=false` in `bindings/proto-ts`) right after the workspace install, with a comment describing this exact mechanism. This PR ports that repair into the package so the documented local flow works too: a `postinstall` on the private `@runanywhere/web-root` package runs the same command, so `npm ci`/`npm install` in `bindings/web` leaves `bindings/proto-ts` able to compile itself. Today the repair exists only in workflow YAML — the root `AGENTS.md` ("Web SDK": `cd bindings/web/`, `npm run build`, `npm run typecheck`), `CONTRIBUTING.md`'s `./run sdk web build`, and `bindings/web/AGENTS.md` all describe the flow without it, so every fresh contributor checkout hits the TS2307 wall.

Deliberately unchanged: the `web-typecheck` workflow's explicit repair step. It becomes redundant (the postinstall performs the same install) but is idempotent and harmless; dropping it can be a follow-up once this has baked. `@runanywhere/web-root` is `"private": true` and never published, so the script cannot reach consumers of the published packages.

## Type of Change
- [x] Bug fix

## Testing
- [x] Lint passes locally
- [ ] Added/updated tests for changes (not applicable — one-line install lifecycle change; validated by the existing suite below)

On Windows 11 (Node v24.18.0, npm 11.16.0 — CI's `web-typecheck` uses v24.19.0 / npm 11.17.0, so this is not version skew), with TS IDL codegen already run:

Fail-before (this change reverted; `bindings/web/node_modules` and `bindings/proto-ts/node_modules` wiped):

```
cd bindings/web && npm ci --no-audit --no-fund   # proto-ts/node_modules absent afterwards
npm run build:ts
# src/chat.ts(8,44): error TS2307: Cannot find module '@bufbuild/protobuf/wire' ... (x39)
```

Pass-after (change applied, same wipe):

```
cd bindings/web && npm ci --no-audit --no-fund   # postinstall restores proto-ts/node_modules (@bufbuild, typescript)
npm run build:ts                                  # exit 0
npm test                                          # 56/56 test files, 244/244 tests pass
```

Before the fix the same `npm test` run had 49 of 56 test files failing (all downstream of the missing proto-ts install). Not run here: WASM builds, browser Playwright suites, and the other SDKs' builds — this change only touches the npm install lifecycle of the private web workspace root; CI remains the authority for those.

**Web SDK / Web Sample:** not applicable — no runtime code changes; browser checklist items unaffected.

## Labels
- [x] `Web SDK` - Changes to Web SDK (`bindings/web`)

## Checklist
- [x] Self-review completed
- [ ] Documentation updated (if needed) — intentionally not touched; with this change the documented commands work as already written

## Summary by CodeRabbit

* **Chores**
  * Improved installation setup by automatically installing required protocol dependencies after package installation.


sanchitmonga22:
Thanks for this, @MohammedAlkindi! The change looks good, and it's a neat way to give local installs the same proto-ts fix CI already applies. Because it adds a `postinstall` script to `bindings/web/package.json`, which runs on every `npm install` / `npm ci` in the Web SDK, including CI and release packaging, a maintainer will give that part a final look before merging.

The two red checks, `centralization` and `flutter-pubget`, aren't caused by your change (one is a release-tag check, the other a pub.dev auth error), so there's nothing for you to fix there.

PR #729, opened 2026-08-17
fix(gates): drop mapfile so the hardcoded-defaults gate runs on macOS

ayaangazali:
## What is wrong

`check_no_hardcoded_defaults.sh` cannot run on macOS. On a stock Mac:

```
$ /bin/bash scripts/validation/gates/check_no_hardcoded_defaults.sh
scripts/validation/gates/check_no_hardcoded_defaults.sh: line 70: mapfile: command not found
$ echo $?
127
```

`mapfile` is bash 4. macOS ships bash 3.2 as `/bin/bash` (3.2.57 here), and the shebang is `#!/usr/bin/env bash`, so it resolves to that unless someone happens to have a newer bash earlier on PATH. The gate exits 127 before reading a single file, and it only ever passes because `pr-build.yml`'s `centralization` job runs on `ubuntu-22.04`.

## Why this is a rule and not my preference

Two sibling scripts already state the constraint and avoid these builtins because of it:

`bindings/swift/scripts/sync-dist-repo.sh:178`
> Built with read loops rather than `mapfile`: macOS ships bash 3.2, which has no mapfile, and this script must run on a stock macOS release runner.

`scripts/build/build-core-android.sh:65`
> ...array (`declare -A`) so this script works on macOS' default /bin/bash 3.2

I swept the tree for the bash 4 builtins (`mapfile`, `readarray`, `declare -A`, `${x^^}`, `${x,,}`) and this line is the only violation left. The other two hits are those comments.

It bites at the worst moment too: a contributor on a Mac reproducing a red gate locally gets a bash error instead of the list of offending files.

## What this does

Replaces `mapfile -t FILES` with the read loop the sibling scripts use. `FILES` stays an array, so `${FILES[@]}`, `${#FILES[@]}`, the empty-scope guard and `--list` all behave as before. 8 insertions, 1 deletion.

## Verification

On this Mac, `/bin/bash` = 3.2.57, no Homebrew bash installed.

Before: exit 127, `mapfile: command not found`.

After:

```
$ /bin/bash scripts/validation/gates/check_no_hardcoded_defaults.sh
OK: no re-declared defaults in 781 scanned SDK source files.
$ echo $?
0
```

I checked the array is complete rather than assuming it, since a read loop can silently drop a final line without a trailing newline. Running the find pipeline standalone gives exactly `781`, matching the gate's count, and `--list` prints 781 paths plus its `scanned 781 files` line.

No test added: there is no harness for these gate scripts, and the check that matters is the one above, running it under the bash version that used to fail.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation compatibility with older Bash versions.
  * Preserved existing SDK file discovery and scope filtering behavior.
  * Validation now handles discovered files consistently across supported Bash environments without changing the checks performed or the files included in scope.

### DIFF
--- a/scripts/validation/gates/check_no_hardcoded_defaults.sh
+++ b/scripts/validation/gates/check_no_hardcoded_defaults.sh
@@ -67,7 +67,14 @@ LIST_ONLY=0
 #   node_modules, build, dist      not source
 #   DevTools, Playground, examples not shipped SDK surface
 # ---------------------------------------------------------------------------
-mapfile -t FILES < <(
+# Read loop rather than `mapfile`: macOS ships bash 3.2, which has no mapfile,
+# and this gate has to be runnable locally to reproduce a CI failure. Same
+# reasoning as bindings/swift/scripts/sync-dist-repo.sh and
+# scripts/build/build-core-android.sh.
+FILES=()
+while IFS= read -r _file; do
+  FILES+=("$_file")
+done < <(
   find \
     bindings/swift/Sources \
     bindings/kotlin/src/main \


ayaangazali:

ayaangazali:
Rebased onto `0b95af9ae` (0.20.27), diff unchanged.

Still reproduces on `main`. The gate is unrunnable on macOS:

```
$ bash scripts/validation/gates/check_no_hardcoded_defaults.sh
scripts/validation/gates/check_no_hardcoded_defaults.sh: line 70: mapfile: command not found
$ echo $?
127
```

`mapfile` is bash 4; macOS ships bash 3.2. It only ever passes because `pr-build.yml` runs it on `ubuntu-22.04`.

Two sibling scripts already avoid these builtins for exactly this reason and say so: `bindings/swift/scripts/sync-dist-repo.sh:178` ("macOS ships bash 3.2, which has no mapfile") and `scripts/build/build-core-android.sh:65` (a `case` block "instead of an associative array (`declare -A`)"). This line is the only remaining violation in the tree.

With the patch applied the gate does its job rather than merely not erroring: it scans **782** SDK source files and passes.

ayaangazali:
Still reproduces on current `main` (`b9032dff5`, 0.20.31).

New data point since I opened this: I ran the whole gate suite locally on a stock Mac today, not just this one file. Eleven of the twelve gates under `scripts/validation/gates/` pass. This is the only one that cannot execute at all:

```
  ok       check_agents_claude_sync.sh
  ok       check_deprecated_surfaces.sh
  ok       check_engine_prebuilt_pins.sh
  ok       check_flutter_centralization.sh
  ok       check_gradle_centralization.sh
  FAIL(127) check_no_hardcoded_defaults.sh
       scripts/validation/gates/check_no_hardcoded_defaults.sh: line 70: mapfile: command not found
  ok       check_qairt_pairing.sh
  ok       check_release_version_coherence.sh
  ok       check_swift_dist_repo_sync.sh
  ok       check_typescript_centralization.sh
  ok       check_wasm_provenance_contract.sh
```

(The twelfth, `check_rn_cpp_headers_compile.sh`, exits 1 with a clear `run yarn install in bindings/react-native` message, which is a prerequisite rather than a failure to run.)

So this is not one script among several that are macOS-unfriendly. It is the single outlier in an otherwise portable suite, which is what the two sibling comments I quoted in the description already assert as the house rule.

Rebased state unchanged, no conflicts with `main`, CI green.

ayaangazali:
While re-verifying this gate I found a second problem in the same file, one line below the change here. Flagging it rather than opening a competing PR, because a separate patch would land a guaranteed conflict on a file that has not been reviewed yet.

The scope comment claims more than the `find` list delivers:

> Scope. First-party SDK source for the six platform SDKs plus the two Android plugin forks.

The list is:

```sh
find \
  bindings/swift/Sources \
  bindings/kotlin/src/main \
  bindings/flutter/packages \
  bindings/react-native/packages \
  bindings/web/packages \
  bindings/python/runanywhere \
```

`bindings/kotlin/modules/` is not there, so the Android plugin forks the comment promises are not scanned at all. That is 8 first-party `.kt` files across `runanywhere-core-{llamacpp,onnx,qhexrt}` that have never been checked. The comment also says "two" forks; there are three now.

I ran the gate's own eight patterns against those 8 files by hand: **zero violations today**, so this is a drift vector rather than a live bug, same character as the portability fix in this PR.

Two ways to take it, your call:

1. I fold a one-line `bindings/kotlin/modules \` addition plus the comment correction into this PR. It stays one concern ("make this gate actually do what it says"), and you review one diff instead of two conflicting ones.
2. I leave this PR focused on the bash 3.2 portability fix and open the scope fix separately once this lands.

I did not push anything either way.

Separately and lower confidence, since it is a scope judgement rather than a stated-but-unimplemented claim: `bindings/electron/src` is not scanned either, though `@runanywhere/electron` is a published, non-private SDK at 0.20.31 in the same language as the two TypeScript bindings that are scanned. It is also clean under the current patterns. Say the word if you want it included and I will add it.

ayaangazali:
Still reproduces on `45e91276e` (0.20.36), and I can now show it failing rather than just pointing at the line.

```
$ /bin/bash --version
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin25)

$ /bin/bash scripts/validation/gates/check_no_hardcoded_defaults.sh
scripts/validation/gates/check_no_hardcoded_defaults.sh: line 70: mapfile: command not found
```

`mapfile` arrived in bash 4. macOS still ships 3.2 as `/bin/bash`, and that is what a contributor gets from `#!/usr/bin/env bash` unless they have installed and PATH-ordered a newer one. So on a stock Mac this gate does not report violations, it dies on line 70 before scanning anything.

The failure mode is the part worth caring about: it exits non-zero with a bash error rather than a gate message, so locally it reads as "my environment is broken" rather than "this check cannot run here", and in a `|| true` context it would pass silently.

I hit the sibling of this myself while writing tooling against this repo this week: `declare -A` is also bash 4+, and the same 3.2 on the same machine silently produced empty variables rather than an error. That is the wider reason to prefer POSIX-compatible constructs in these gates.

The fix is a `while read` loop, which is what the other gates in this directory already use.

Merges clean, CI green.

sanchitmonga22:
@coderabbitai review

sanchitmonga22:
@coderabbitai full review

sanchitmonga22:
> Replying to [this comment](https://github.com/RunanywhereAI/runanywhere-sdks/pull/729#issuecomment-5456275331)

Thanks for flagging that separately — I'd rather keep this PR scoped to the `mapfile` → `while read` portability fix. I opened #860 for the stale "two Android plugin forks" comment and the missing `bindings/kotlin/modules`/`bindings/electron` scan roots — feel free to pick it up if you're interested!

sanchitmonga22 (approved):
Thanks a lot for this, @ayaangazali! This makes the `check_no_hardcoded_defaults.sh` gate actually runnable on macOS's stock bash 3.2 — it used to die with `mapfile: command not found` (exit 127), so Mac contributors couldn't reproduce this CI gate locally at all.

Checked: CodeRabbit reviewed the latest commit · CI green · built and linted locally merged into `main` (fast gates / shell script portability) · two independent code reviews.

Follow-ups, not blocking: #860 (stale scope comment + missing `bindings/kotlin/modules`/`bindings/electron` scan roots) and #861 (gate silently under-scans if a root is renamed) — you're welcome to pick these up.

PR #730, opened 2026-08-17
fix(scripts): stop the Linux Sherpa download falling back to a stale version

ayaangazali:
## What is wrong

`core/scripts/load-versions.sh` states the rule in its own header:

> The VERSIONS file is the SINGLE SOURCE OF TRUTH for all versions.
> DO NOT hardcode version fallbacks in scripts - always source this file.

Two download scripts carry one anyway, and both literals are stale:

| script | fallback | canonical in `core/VERSIONS` |
|---|---|---|
| `core/scripts/linux/download-sherpa-onnx.sh:26` | `1.12.18` | `1.13.5` |
| `core/scripts/windows/download-sherpa-onnx.bat:24` | `1.12.23` | `1.13.5` |

They were not mirrored through the 1.13.2 bump, nor the 1.13.5 one in #718, so they now sit two minor versions behind.

## Why it matters now

`engines/sherpa/sherpa_backend.cpp` refuses Nemotron 3.5 prompted streaming ASR below 1.13.5, with a comment added in #718 explaining that a 1.13.4 runtime loads the graph but decodes it incorrectly. A build that quietly picks up 1.12.x therefore produces a runtime that cannot serve a model the catalog now offers, and the failure surfaces later as a version-rejection at inference time rather than at build time.

**The Windows path is reachable today.** `:load_versions` does `if not exist "%VERSIONS_FILE%" exit /b 1`, but `exit /b` only leaves the subroutine and the caller never checks the result, so a missing VERSIONS file falls straight through to `1.12.23`.

On Linux the missing-file case is caught, since `source` returning non-zero trips `set -e`. A renamed or removed key is not: the loader succeeds without exporting, and `:-` supplies the stale literal.

## What this does

Makes both fail closed, which deletes the stale numbers instead of updating them.

This is what every other platform already does. The macOS and iOS sherpa scripts error out when the variable is absent, and on Linux the three variables on the lines immediately below this one already use `:?`:

```sh
REPOSITORY="${SHERPA_ONNX_REPO_DESKTOP:?SHERPA_ONNX_REPO_DESKTOP is not set}"
RELEASE_TAG="${SHERPA_ONNX_RELEASE_TAG_DESKTOP:?...}"
SOURCE_COMMIT="${SHERPA_ONNX_COMMIT_DESKTOP:?...}"
```

So `VERSION` was the one variable in that group that guessed.

## Verification

`bash -n` on the Linux script passes.

Normal path, with VERSIONS sourced as the script does:

```
$ source core/scripts/load-versions.sh && echo "${SHERPA_ONNX_VERSION_LINUX:?}"
1.13.5
```

The behaviour change with the key absent:

```
# before
$ V="${SHERPA_ONNX_VERSION_LINUX:-1.12.18}"; echo "built against: $V"; echo $?
built against: 1.12.18
0

# after
$ V="${SHERPA_ONNX_VERSION_LINUX:?...}"
bash: SHERPA_ONNX_VERSION_LINUX: SHERPA_ONNX_VERSION_LINUX is not set (load-versions.sh should export it from core/VERSIONS)
# non-zero
```

I also grepped the tree for `1.12.18` and `1.12.23` to be sure nothing else depended on those values; there are no other references.

I could not execute the `.bat` on this machine, so that half is a read of the control flow rather than something I ran. It is the same one-line shape as the Linux change and matches the file's existing `echo [ERROR] ... / exit /b 1` idiom used in four other places.

No test added: these are build-time download scripts with no harness, and the check that matters is the one above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Linux downloads now require an explicitly configured Sherpa-ONNX version.
  * Downloads stop immediately when the version configuration is missing, preventing use of an outdated fallback version.
  * Stale inherited version and release settings are cleared before configuration is loaded, ensuring downloads use the intended release metadata.


ayaangazali:

ayaangazali:
Rebased onto `01490bd86` and cut down: **this is now the Linux half only.**

#761 landed the Windows half independently, with the same fail-closed shape and the same reasoning:

> Fail closed rather than falling back. The old hardcoded default (1.12.23) had already drifted from `core/VERSIONS` (now 1.13.5), so an unreadable VERSIONS file would silently request a nonexistent release asset and the sherpa backend would then build as a non-routable stub.

That is this PR's argument, reached separately, so I dropped my version of `download-sherpa-onnx.bat` rather than compete with it. That file is now byte-identical to main here.

Linux is still unfixed on `main`:

```
core/scripts/linux/download-sherpa-onnx.sh:26
VERSION="${SHERPA_ONNX_VERSION_LINUX:-1.12.18}"

core/VERSIONS:230
SHERPA_ONNX_VERSION_LINUX=1.13.5
```

Same two-versions-behind drift #761 described for Windows, and #761 also documented what that costs on Linux specifically: `native_linux` shipped "a ~156 KB hollow `librac_backend_sherpa.so` with no `libsherpa-onnx-c-api.so` beside it and no working STT/TTS/VAD".

The remaining diff is one line, and it makes `VERSION` behave like the three lines directly beneath it, which already use `:?`:

```sh
REPOSITORY="${SHERPA_ONNX_REPO_DESKTOP:?SHERPA_ONNX_REPO_DESKTOP is not set}"
RELEASE_TAG="${SHERPA_ONNX_RELEASE_TAG_DESKTOP:?SHERPA_ONNX_RELEASE_TAG_DESKTOP is not set}"
SOURCE_COMMIT="${SHERPA_ONNX_COMMIT_DESKTOP:?SHERPA_ONNX_COMMIT_DESKTOP is not set}"
```

`bash -n` passes, and the guard fires as intended when the variable is absent:

```
$ bash: SHERPA_ONNX_VERSION_LINUX: SHERPA_ONNX_VERSION_LINUX is not set (load-versions.sh should export it from core/VERSIONS)
```

ayaangazali:
Rebased onto `0b95af9ae` (0.20.27), diff unchanged. Still a one-line Linux fix.

Still reproduces on `main`, and the drift has widened since I opened this:

```
core/scripts/linux/download-sherpa-onnx.sh:26
VERSION="${SHERPA_ONNX_VERSION_LINUX:-1.12.18}"

core/VERSIONS
SHERPA_ONNX_VERSION_LINUX=1.13.5
```

#761 landed the same fail-closed change for Windows with the same reasoning ("The old hardcoded default had already drifted from `core/VERSIONS` [...] an unreadable VERSIONS file would silently request a nonexistent release asset and the sherpa backend would then build as a non-routable stub"), so I dropped my copy of the `.bat` rather than compete with it. Linux is the half that never got it.

The change makes `VERSION` behave like the three lines directly beneath it, which already use `:?`.

ayaangazali:
Still reproduces on `c11f78ef4`, and the gap is wider than when I opened this.

The fallback literal is now seven patch releases behind the canonical value it is standing in for:

```
$ grep -n 'SHERPA_ONNX_VERSION_LINUX' core/scripts/linux/download-sherpa-onnx.sh core/VERSIONS
core/scripts/linux/download-sherpa-onnx.sh:26:VERSION="${SHERPA_ONNX_VERSION_LINUX:-1.12.18}"
core/VERSIONS:230:SHERPA_ONNX_VERSION_LINUX=1.13.5
```

That is the whole argument for the change. If `load-versions.sh` ever stops exporting this key, the script does not fail, it quietly downloads 1.12.18 and the build carries on against a Sherpa-ONNX seven versions off the pin, which is a much harder thing to notice than a missing variable.

The three variables immediately below it already fail fast, so this is one line out of step with its own neighbours rather than a new convention:

```
REPOSITORY="${SHERPA_ONNX_REPO_DESKTOP:?SHERPA_ONNX_REPO_DESKTOP is not set}"
RELEASE_TAG="${SHERPA_ONNX_RELEASE_TAG_DESKTOP:?...}"
SOURCE_COMMIT="${SHERPA_ONNX_COMMIT_DESKTOP:?...}"
```

Merges clean, CI green.

ayaangazali:
Rebased onto 0.20.35. The one red check is not from this diff.

`android-arm64` fails at **step 3, `Run ./.github/actions/setup-toolchain`**. The step that would exercise anything this PR touches is step 5, `Download vendored Sherpa-ONNX (Android)`, and it never runs:

```
FAILED step 3: Run ./.github/actions/setup-toolchain
  step 4 skipped: apt-get install ninja-build
  step 5 skipped: Download vendored Sherpa-ONNX (Android)
  step 6 skipped: cmake --preset android-arm64
```

Two further reasons it cannot be this PR: the file changed here is `core/scripts/linux/download-sherpa-onnx.sh`, the Linux downloader, not the Android path, and `android-arm64` is SUCCESS on all 34 of my other open PRs against the same base. So this is toolchain-setup flake, not a regression. Not pushing a speculative fix for it.

On the change itself, the gap is wider than when I opened this:

```
$ grep -n 'SHERPA_ONNX_VERSION_LINUX' core/scripts/linux/download-sherpa-onnx.sh core/VERSIONS
core/scripts/linux/download-sherpa-onnx.sh:26:VERSION="${SHERPA_ONNX_VERSION_LINUX:-1.12.18}"
core/VERSIONS:230:SHERPA_ONNX_VERSION_LINUX=1.13.5
```

If `load-versions.sh` ever stops exporting that key the script does not fail, it silently downloads a Sherpa-ONNX seven patch releases off the pin, which is far harder to notice than a missing variable. The three variables immediately below it already use fail-fast `:?`, so this is one line out of step with its own neighbours.

ayaangazali:
Correcting something I told you in my last comment here. I asserted a behaviour without testing it, and it is not what happens.

I said the fallback means the script "does not fail, it quietly downloads 1.12.18 and the build carries on against a Sherpa-ONNX seven versions off the pin". That is wrong. The release tag is pinned separately, so the fallback builds an asset name that does not exist under the tag that does:

```
$ TAG=v1.13.5-rac-desktop.4
$ curl -sIL -o /dev/null -w '%{http_code}' .../$TAG/sherpa-onnx-v1.13.5-linux-x64-shared-rac-ort1.28.0.tar.bz2
200
$ curl -sIL -o /dev/null -w '%{http_code}' .../$TAG/sherpa-onnx-v1.12.18-linux-x64-shared-rac-ort1.28.0.tar.bz2
404
```

`curl -L --fail` then fails the build. So there is no silent stale download. My apologies for putting that in your inbox.

The narrower, accurate case for this change:

`SHERPA_ONNX_REPO_DESKTOP`, `SHERPA_ONNX_RELEASE_TAG_DESKTOP` and `SHERPA_ONNX_COMMIT_DESKTOP` all use `:?`, so if `load-versions.sh` stops exporting wholesale the script aborts on line 8 naming the missing variable. The fallback is therefore only reachable when `SHERPA_ONNX_VERSION_LINUX` **specifically** goes missing, which is what a rename or typo in `core/VERSIONS` looks like. In that one case you do not get "SHERPA_ONNX_VERSION_LINUX is not set"; you get a 404 for an asset under a tag that plainly exists, which reads as a broken release rather than a broken variable.

That is a smaller claim than the one I made, and it is the one I can actually demonstrate. The one-word change still seems right to me on consistency grounds alone, since it is the only one of four adjacent variables not using `:?`, but I would understand if you weigh it differently now.

The stale literal is still real: fallback `1.12.18` against canonical `SHERPA_ONNX_VERSION_LINUX=1.13.5`.

ayaangazali:
Half of this is a real gap in my fix and I have taken it. The other half is already in main.

**Taken: clearing the keys before loading.** You are right that `:?` does not distinguish "VERSIONS defines it" from "an earlier shell exported it". `load-versions.sh` exports these, so a developer who sourced it once keeps a stale value that satisfies the guard after the key is renamed away. That is the same silent stale-version outcome the guard exists to prevent. Demonstrated with the key removed from a stand-in loader:

```
before:  SHERPA_ONNX_VERSION_LINUX=1.12.18 -> used VERSION=1.12.18
after:   SHERPA_ONNX_VERSION_LINUX=1.12.18 -> bash: SHERPA_ONNX_VERSION_LINUX: not set
```

I cleared all four keys the script requires, not just the version, since the repo, tag and commit keys have the identical exposure. Nothing in the repo sets any of them in the environment, so this removes no intended override.

**Already present: rejecting loader failure in the .bat.** `core/scripts/windows/download-sherpa-onnx.bat:28-34` already fails closed on both `SHERPA_ONNX_VERSION_WINDOWS` and `SHERPA_ONNX_WINDOWS_X64_SHA256`, with a comment giving the same reasoning. That is pre-existing, not part of this PR. It is actually the strongest argument for this PR: Windows already failed closed, and Linux was the one platform still falling back.

The .bat does share the inherited-value hole, since `call :load_versions` will not overwrite an already-set variable either. I have not touched it here. I cannot execute a `.bat` on this machine, and I would rather not ship an untested Windows change inside a Linux fix. Say the word and I will do it as its own change.

sanchitmonga22:
Thanks for this, @ayaangazali! The change looks good. Because it touches `core/scripts/linux/download-sherpa-onnx.sh` (the script the release workflow runs to fetch the pinned Sherpa-ONNX Linux binaries), a maintainer will give that part a final look before merging.

One optional thing while you wait: the PR description still covers the Windows half, which is already on `main`, and the "silently uses the old version" claim you later corrected. Trimming both would keep the squash message accurate.

sanchitmonga22:
> Replying to [this comment](https://github.com/RunanywhereAI/runanywhere-sdks/pull/730#issuecomment-5591876463)

Thanks for taking the clear-before-load fix, and for keeping an untested Windows change out of a Linux PR (re: https://github.com/RunanywhereAI/runanywhere-sdks/pull/730#issuecomment-5591876463). Yes please, as its own change: I opened #854 for it.

One correction for that PR: `:load_versions` does overwrite a variable that is already set, because `core/scripts/windows/download-sherpa-onnx.bat:225` runs `set "%%a=%%b"` for every key it reads. So the hole only opens when a key is gone from `core/VERSIONS`, same as on Linux. `SHERPA_ONNX_REPO_DESKTOP`, `SHERPA_ONNX_RELEASE_TAG_DESKTOP`, `SHERPA_ONNX_COMMIT_DESKTOP` (lines 37-39) and `ONNX_VERSION_WINDOWS` (line 59) have no guard yet, so they're worth covering too. You won't need a Windows machine to see it run: `.github/workflows/pr-build.yml` runs the `.bat` in two Windows jobs (lines 195 and 258).

PR #731, opened 2026-08-17
fix(kotlin): stop deprecations pointing at deprecated replacements

ayaangazali:
## What is wrong

Six `@Deprecated` messages on the legacy extension surface tell the caller to move to a namespace method that is itself deprecated, so following the advice lands them on a second deprecation:

| legacy extension | current advice | that target's own annotation |
|---|---|---|
| `unloadModel` | `models.unload(category)` | `@Deprecated("Use unloadAll(category).")` |
| `stopSynthesis` | `tts.stop()` | `@Deprecated("Use the SpeechHandle returned by speak().")` |
| `stopSpeaking` | `tts.stop()` | same |
| `streamVAD` | `vad.detectStream(audio, options)` | `@Deprecated("Use vad.openStream(format, options).")` |
| `resetVAD` | `vad.detectStream` | same |
| `transcribeStream` | `stt.transcribeStream(audio, options)` | `@Deprecated("Use stt.openStream(format, options).")` |

The IDE shows the strikethrough and the message, the caller migrates, and gets struck through again. `ReplaceWith` is not involved in any of these, so nothing auto-applies, but the written guidance is a dead end and no gate covers it. `check_deprecated_surfaces.sh` tracks which deprecated surfaces exist against an allowlist; it does not check that a deprecation's target is supported.

## What this changes

Each message now names the replacement the namespace's own annotation points at, so one hop reaches a supported API:

- `models.unloadAll(category)`
- `interrupt()` on the `SpeechHandle` returned by `tts.speak()`
- `vad.openStream(format, options)`, and `close()` on the `VadStream` it returns for the reset case
- `stt.openStream(format, options)`

I took each target from the deprecated namespace method's own annotation rather than picking one, and confirmed the members exist: `ModelsNamespace.unloadAll` (`:266`), `TtsNamespace.speak` (`:60`) returning `SpeechHandle` with `interrupt()` (`Results.kt:252`), `SttNamespace.openStream` (`:189`), and `VadStream.close()` (`Results.kt:315`).

Message strings only. No signature, body, or `ReplaceWith` target changes, so this cannot alter behaviour.

## How I found them

Collected the deprecated methods declared inside `public/api/*Namespace.kt`, then looked for any `@Deprecated` message elsewhere naming `<namespace>.<method>` for that set. Six hits, four files. After the change the same scan reports zero.

Worth noting the scan produced 31 raw hits at first and 25 were false positives, because a deprecated extension and the live namespace method often share a simple name (`generate`, `rerank`, `diarize`). Those messages are correct and I left them alone; only the six above name a target that really is deprecated.

## Verification

- `./gradlew :compileDebugKotlin --rerun-tasks -x :buildLocalJniLibs -x :downloadJniLibs -x :syncAndroidRuntimeLibs -Prunanywhere.useLocalNatives=false`: **BUILD SUCCESSFUL**
- `./gradlew ktlintMainSourceSetCheck`: clean
- `git status` shows only the four source files; the codegen the Gradle build runs left nothing behind

No test added: these are annotation strings, and a test asserting the text of a deprecation message would pin prose rather than behaviour.

## Summary by CodeRabbit

* **Documentation**
  * Updated deprecation guidance for model unloading, speech-to-text streaming, text-to-speech controls, and voice activity detection.
  * Deprecated APIs now point to recommended replacement workflows, including stream handles and lifecycle methods.

### DIFF
--- a/bindings/kotlin/src/main/kotlin/com/runanywhere/sdk/public/extensions/Models/RunAnywhereModelLifecycle.kt
+++ b/bindings/kotlin/src/main/kotlin/com/runanywhere/sdk/public/extensions/Models/RunAnywhereModelLifecycle.kt
@@ -97,7 +97,7 @@ suspend fun RunAnywhere.loadModel(model: RAModelInfo): RAModelLoadResult =
         ),
     )
 
-@Deprecated("Use RunAnywhere.models.unload(category).")
+@Deprecated("Use RunAnywhere.models.unloadAll(category).")
 suspend fun RunAnywhere.unloadModel(request: ModelUnloadRequest): ModelUnloadResult {
     if (!isInitialized) {
         return ModelUnloadResult(
--- a/bindings/kotlin/src/main/kotlin/com/runanywhere/sdk/public/extensions/STT/RunAnywhereSTT.kt
+++ b/bindings/kotlin/src/main/kotlin/com/runanywhere/sdk/public/extensions/STT/RunAnywhereSTT.kt
@@ -85,7 +85,7 @@ suspend fun RunAnywhere.transcribe(
     return result
 }
 
-@Deprecated("Use RunAnywhere.stt.transcribeStream(audio, options).")
+@Deprecated("Use RunAnywhere.stt.openStream(format, options).")
 fun RunAnywhere.transcribeStream(
     audio: Flow<ByteArray>,
     options: RASTTOptions = RASTTOptions.defaults(),
--- a/bindings/kotlin/src/main/kotlin/com/runanywhere/sdk/public/extensions/TTS/RunAnywhereTTS.kt
+++ b/bindings/kotlin/src/main/kotlin/com/runanywhere/sdk/public/extensions/TTS/RunAnywhereTTS.kt
@@ -177,7 +177,7 @@ fun RunAnywhere.synthesizeStream(
         }
     }
 
-@Deprecated("Use RunAnywhere.tts.stop().")
+@Deprecated("Use interrupt() on the SpeechHandle returned by RunAnywhere.tts.speak().")
 suspend fun RunAnywhere.stopSynthesis() {
     CppBridgeTTS.stop()
 }
@@ -241,7 +241,7 @@ private fun convertPcmToWav(pcmData: ByteArray, sampleRate: Int): ByteArray {
         ?: throw SDKException.tts("Failed to convert PCM to WAV")
 }
 
-@Deprecated("Use RunAnywhere.tts.stop().")
+@Deprecated("Use interrupt() on the SpeechHandle returned by RunAnywhere.tts.speak().")
 suspend fun RunAnywhere.stopSpeaking() {
     ttsAudioPlayback.stop()
     stopSynthesis()
--- a/bindings/kotlin/src/main/kotlin/com/runanywhere/sdk/public/extensions/VAD/RunAnywhereVAD.kt
+++ b/bindings/kotlin/src/main/kotlin/com/runanywhere/sdk/public/extensions/VAD/RunAnywhereVAD.kt
@@ -93,7 +93,7 @@ suspend fun RunAnywhere.detectVoiceActivity(
  * `error_code`) and the flow finishes so callers do not silently keep
  * pumping audio into a dead detector.
  */
-@Deprecated("Use RunAnywhere.vad.detectStream(audio, options).")
+@Deprecated("Use RunAnywhere.vad.openStream(format, options).")
 fun RunAnywhere.streamVAD(
     audio: Flow<ByteArray>,
     options: RAVADOptions? = null,
@@ -115,7 +115,7 @@ fun RunAnywhere.streamVAD(
             )
         }
 
-@Deprecated("Cancel the Flow returned by RunAnywhere.vad.detectStream instead.")
+@Deprecated("Close the VadStream returned by RunAnywhere.vad.openStream instead.")
 suspend fun RunAnywhere.resetVAD() {
     if (!isInitialized) {
         throw notInitializedException()


ayaangazali:

ayaangazali:
Rebased onto `0b95af9ae` (0.20.27), diff unchanged.

Still reproduces. Re-ran the check against current `main`: five Kotlin deprecations still send the caller to a replacement that is itself deprecated.

```
RunAnywhereModelLifecycle.kt:100   -> RunAnywhere.models.unload
RunAnywhereTTS.kt:180              -> RunAnywhere.tts.stop
RunAnywhereTTS.kt:244              -> RunAnywhere.tts.stop
RunAnywhereVAD.kt:96               -> RunAnywhere.vad.detectStream
RunAnywhereSTT.kt:88               -> RunAnywhere.stt.transcribeStream
```

Following any of those trades one deprecation warning for another. This retargets them at the live replacements.

For context on scope: the same defect exists in Swift and Web and is #762, and `ReplaceWith` targets elsewhere in the Kotlin SDK all resolve correctly, so this is the complete Kotlin set rather than a sample.

ayaangazali:
Still reproduces on `c11f78ef4`. Re-checked each of the four replacement targets this time rather than the diff, and every one of them is itself deprecated on `main`:

```
ModelsNamespace.kt:281  @Deprecated("Use unloadAll(category).", ReplaceWith("unloadAll(category)"))
                        public suspend fun unload(category: ModelCategory? = null)

SttNamespace.kt:205     @Deprecated("Use stt.openStream(format, options).", ReplaceWith("openStream(format, options)"))
                        public fun transcribeStream(

TtsNamespace.kt:80      @Deprecated("Use the SpeechHandle returned by speak().")
                        public suspend fun stop()

VadNamespace.kt:186     @Deprecated("Use vad.openStream(format, options).", ReplaceWith("openStream(format, options)"))
                        public fun detectStream(
```

So the six extension deprecations this PR touches send a caller to a symbol that immediately warns again and carries its own `ReplaceWith` pointing one hop further on. Following the IDE quick-fix gets you a second deprecation warning rather than working code. The fix is to name the end of the chain directly, which is what these six messages now do.

The replacements they point at all exist and are not deprecated (`ModelsNamespace.unloadAll:266`, `SttNamespace.openStream:189`, `VadNamespace.openStream:171`, and the `SpeechHandle` returned by `tts.speak()`).

Text-only change to `@Deprecated` messages, no behaviour. Merges clean, CI green.

ayaangazali:
Still reproduces on `45e91276e` (0.20.36), and there is now cross-SDK confirmation that this is a pattern rather than a Kotlin oversight.

I verified the same defect on the Swift and Web side while working #762, and it is the identical shape. Swift deprecates a helper by pointing at `tts.stop()`:

```swift
// bindings/swift/.../Public/Extensions/TTS/RunAnywhere+TTS.swift:55
@available(*, deprecated, renamed: "tts.stop()")
```

and that target is itself deprecated:

```swift
// bindings/swift/.../Public/API/Namespaces/TTSNamespace.swift:102
@available(*, deprecated, message: "Use the SpeechHandle returned by speak(_:options:) to interrupt one utterance")
public func stop() async {
```

So the chain this PR fixes in Kotlin exists in at least three SDKs, with the same end state (the `SpeechHandle` returned by `speak`). The four Kotlin targets are unchanged on `main`:

```
ModelsNamespace.kt:281  @Deprecated("Use unloadAll(category).", ReplaceWith("unloadAll(category)"))
SttNamespace.kt:205     @Deprecated("Use stt.openStream(format, options).", ...)
TtsNamespace.kt:80      @Deprecated("Use the SpeechHandle returned by speak().")
VadNamespace.kt:186     @Deprecated("Use vad.openStream(format, options).", ...)
```

The `ReplaceWith` entries are the part that stings, because that is what drives the IDE's automated migration: accepting the quick-fix rewrites one deprecated call into another.

Text-only change to `@Deprecated` messages, no behaviour. Merges clean, CI green.

sanchitmonga22:
@coderabbitai review

sanchitmonga22:
@coderabbitai full review

sanchitmonga22:
Thanks for this, @ayaangazali! Retargeting these `@Deprecated` messages away from other deprecated symbols makes the migration path for Models/STT/TTS/VAD callers a lot clearer, and the investigation you posted along the way is genuinely thorough.

Before we merge, a few things to sort out — all in what the new targets actually do, not the file list:

1. `bindings/kotlin/.../Models/RunAnywhereModelLifecycle.kt:100` — the new message sends every `unloadModel(request)` caller to `models.unloadAll(category)`, but `ModelUnloadRequest` can also target one model by `model_id`, and `unloadAll` never reads that field — it unloads a whole category (or everything) instead. Suggest: "Use RunAnywhere.models.unload(id) for a single model, or unloadAll(category) for a whole category."
2. `bindings/kotlin/.../TTS/RunAnywhereTTS.kt:180` — `stopSynthesis()` also stops audio started via `tts.synthesizeStream()`, but the new message only names the `SpeechHandle` from `tts.speak()`, which streaming callers never get. Suggest adding: "Cancel the Flow returned by RunAnywhere.tts.synthesizeStream(), or call interrupt() on the SpeechHandle returned by RunAnywhere.tts.speak()."
3. `bindings/kotlin/.../VAD/RunAnywhereVAD.kt:118` — `resetVAD()` resets the lifecycle detector's adaptive state (thresholds, timing) via `resetLifecycle()`; `VadStream.close()` only cancels that one stream session and never touches that state, so following the new message doesn't actually reset anything. There isn't a non-deprecated equivalent of the lifecycle reset today, which is worth saying plainly rather than pointing at something that doesn't do it.

Not blocking this PR — I opened #862 for the same deprecation-chain pattern in the Swift and Web SDKs (outside this PR's scope), #863 for giving batch `tts.synthesize` a non-deprecated way to cancel, and #864 for updating the Kotlin docs that still present these legacy extensions as the current API. You're welcome to pick any of those up if you're interested.

Once the three messages above point at something that actually does what the old one did, we'll take another look. Thanks again for the thorough work here!

PR #736, opened 2026-08-18
fix(ci): stop the python-linux wheel repair hanging on a symlink loop

ayaangazali:
## What is wrong

`python-linux (3.9)` and `(3.12)` are red on **main**, not just on PRs. Run [32011099890](https://github.com/RunanywhereAI/runanywhere-sdks/actions/runs/32011099890) at `b2d2ff6d`:

> The job has exceeded the maximum execution time of 1h30m0s

The log stops dead at the start of the repair step and never prints again:

```
08:47:58 INFO:auditwheel.main_repair:Repairing runanywhere-0.20.24-cp39-cp39-linux_x86_64.whl
10:05:51 ##[error]The operation was canceled.
10:05:52 Terminate orphan process: pid (13128) (auditwheel)
```

The same job took about **11 minutes** on 2026-08-16 (see #720's run, `python-linux (3.12)` in 10m16s). Every PR opened since inherits the red cell.

## Root cause

`pr-build.yml` did not change in that window, so the input did. #718 moved the Linux prebuilt off k2-fsa's upstream tarball and onto the RunAnywhere desktop asset, and the two lay ONNX Runtime out differently:

```
upstream v1.13.2       libonnxruntime.so                          real file, 25,008,897 bytes

rac-desktop.4          libonnxruntime.so   -> libonnxruntime.so.1
                       libonnxruntime.so.1 -> libonnxruntime.so.1.28.0
                       libonnxruntime.so.1.28.0                   real file, 24,268,848 bytes
```

The prep step runs:

```sh
ln -sf libonnxruntime.so "$lib/libonnxruntime.so.1"
```

On the old flat layout that created the versioned name `_core`/sherpa record as `DT_NEEDED`, exactly as the comment says. On the new layout it **overwrites the middle link of an existing chain**, leaving:

```
libonnxruntime.so   -> libonnxruntime.so.1
libonnxruntime.so.1 -> libonnxruntime.so
```

That is a cycle. The real 23 MB library is no longer reachable through either name, and auditwheel sits on it until the runner's 1h30m limit kills the job.

## What this changes

Create the versioned name only when the prebuilt does not already ship it. Both layouts then work, and the comment now records why overwriting is not safe.

## Verification

I downloaded both assets and ran the current command and the new one against each, then checked whether `libonnxruntime.so.1` still resolves to a real file:

| command | rac-desktop.4 asset | upstream v1.13.2 asset |
|---|---|---|
| `ln -sf ...` (main today) | **BROKEN: Too many levels of symbolic links** | OK, 25,008,897 bytes |
| `[ -e ... ] \|\| ln -sf ...` (this PR) | OK, 24,268,848 bytes | OK, 25,008,897 bytes |

So the failure is real and asset-dependent, and the guard fixes the new layout while leaving the old behaviour byte-identical.

`yaml.safe_load` parses the workflow (22 jobs).

I could not run `auditwheel` itself here, since this is a macOS host and the hang is Linux-only. The ELOOP above is the mechanism, and this PR's own `python-linux` cells are the real test: if they go green while main stays red, that confirms it end to end.

One caveat worth stating: this unblocks the job, it does not claim the vendored result is otherwise unchanged. auditwheel will now vendor `libonnxruntime.so.1.28.0` through the intact chain, which is what it did before #718 changed the asset.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Linux package builds by preserving existing shared-library links when compatible prebuilt files are already available.
  * Prevented unnecessary replacement of library symlink chains during wheel creation.
  * Improved package validation to correctly recognize versioned Linux shared libraries and compatible sidecar libraries.
  * Added validation coverage for wheels containing versioned runtime libraries, ensuring these packages are accepted correctly.

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -379,10 +379,14 @@ jobs:
       - name: Sherpa-ONNX prebuilts
         run: |
           ./core/scripts/linux/download-sherpa-onnx.sh
-          # The prebuilt ships libonnxruntime.so, but _core/sherpa record a versioned
-          # DT_NEEDED (libonnxruntime.so.1); give auditwheel a matching filename to vendor.
+          # _core/sherpa record a versioned DT_NEEDED (libonnxruntime.so.1), so that
+          # name has to exist for auditwheel to vendor it. Only create it when the
+          # prebuilt does not already ship it: the RunAnywhere desktop asset ships the
+          # usual libonnxruntime.so -> .so.1 -> .so.1.<ver> chain, and overwriting the
+          # middle link points .so.1 back at .so, which is a symlink loop that leaves
+          # the real library unreachable and hangs auditwheel.
           lib="core/third_party/sherpa-onnx-linux/lib"
-          ln -sf libonnxruntime.so "$lib/libonnxruntime.so.1"
+          [ -e "$lib/libonnxruntime.so.1" ] || ln -sf libonnxruntime.so "$lib/libonnxruntime.so.1"
       - name: Build wheel
         working-directory: bindings/python
         run: python -m build --wheel -o dist
--- a/bindings/python/scripts/test_validate_public_packages.py
+++ b/bindings/python/scripts/test_validate_public_packages.py
@@ -203,6 +203,27 @@ class ValidatePublicPackagesTest(unittest.TestCase):
                 _write_dist(self._tmp(), wheel=_build_wheel(sidecars=partial)), VERSION
             )
 
+    def test_versioned_linux_sidecar_is_accepted(self) -> None:
+        """auditwheel keeps the soname, so the vendored ORT is ``.so.1.28.0``.
+
+        The upstream sherpa tarball shipped a flat ``libonnxruntime.so``, so the
+        relocated copy ended in ``.so`` and the suffix check happened to pass. The
+        RunAnywhere desktop asset ships ``libonnxruntime.so.1.28.0``, and the
+        vendored name keeps that version.
+        """
+        dist = self._tmp() / "dist"
+        dist.mkdir(parents=True)
+        wheel = _build_wheel(
+            sidecars={
+                "runanywhere.libs/libonnxruntime-a1b2c3d4.so.1.28.0": CLEAN_BINARY,
+                "runanywhere.libs/libsherpa-onnx-c-api-e5f6a7b8.so": CLEAN_BINARY,
+            }
+        )
+        linux_wheel = f"runanywhere-{VERSION}-cp39-cp39-manylinux_2_39_x86_64.whl"
+        (dist / linux_wheel).write_bytes(wheel)
+        (dist / SDIST_NAME).write_bytes(_build_sdist())
+        V.validate_public_packages(dist, VERSION)
+
     def test_missing_core_extension_is_rejected(self) -> None:
         with self.assertRaisesRegex(V.PackageValidationError, "_core"):
             V.validate_public_packages(
--- a/bindings/python/scripts/validate_public_packages.py
+++ b/bindings/python/scripts/validate_public_packages.py
@@ -40,6 +40,18 @@ HOST_PATH_MARKERS: tuple[bytes, ...] = (b"/Users/", b"/home/", b"\\Users\\")
 # Compiled-binary suffixes that belong in the wheel but must NEVER appear in the sdist.
 BINARY_SUFFIXES: tuple[str, ...] = (".pyd", ".so", ".dylib", ".dll")
 
+
+def _is_binary_member(name: str) -> bool:
+    """True for compiled artifacts, including ELF names carrying a soname version.
+
+    ``endswith(BINARY_SUFFIXES)`` alone misses ``libonnxruntime-abc123.so.1.28.0``:
+    the ELF convention puts the version *after* the suffix, so a vendored sidecar
+    only ends in ``.so`` when the library it came from was unversioned. That is what
+    ``_basename_stem`` already documents, and both callers below need the same rule.
+    """
+    lowered = name.casefold()
+    return lowered.endswith(BINARY_SUFFIXES) or ".so." in lowered
+
 # The compiled extension is imported as ``runanywhere._native._core``; its file is named
 # ``_core`` with a platform extension suffix (``_core.cp312-win_amd64.pyd`` / ``_core.so`` /
 # ``_core.*.dylib``). Match the stem so the check is interpreter/platform independent.
@@ -152,7 +164,7 @@ def validate_wheel(wheel: Path, expected_version: str, *, gpu: bool = False) -> 
                 for n in names
                 if posixpath.dirname(n) == CORE_DIR
                 and _basename_stem(n) == CORE_STEM
-                and n.casefold().endswith(BINARY_SUFFIXES)
+                and _is_binary_member(n)
             ]
             if not core_members:
                 raise PackageValidationError(
@@ -182,7 +194,7 @@ def validate_wheel(wheel: Path, expected_version: str, *, gpu: bool = False) -> 
                 _basename_stem(n)
                 for n in names
                 if posixpath.dirname(n) == platform.libs_dir
-                and n.casefold().endswith(BINARY_SUFFIXES)
+                and _is_binary_member(n)
             }
             # Repair tools mangle a hash into the name (onnxruntime -> onnxruntime-a1b2c3),
             # so match a stem that equals the expected lib or begins with "<lib>-".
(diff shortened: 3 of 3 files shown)


ayaangazali:
The first commit did its job and uncovered the next problem, so this PR now carries both.

**Where it stands.** `python-linux` no longer hangs: it went from the 1h30m timeout to failing in ~15 minutes, and the CI log shows the chain intact, which is exactly what the guard was for:

```
libonnxruntime.so:        19       (symlink)
libonnxruntime.so.1:      24       (symlink)
libonnxruntime.so.1.28.0: 24268848 (real)
```

`auditwheel repair` then completes in about 10 seconds and writes the fixed-up wheel.

**The next failure, now fixed here too.** The release-boundary step rejected the result:

```
ERROR: runanywhere-0.20.24-cp39-cp39-manylinux_2_39_x86_64.whl:
       vendored sidecar libraries missing from runanywhere.libs/: ['libonnxruntime']
```

The library is in the wheel. `validate_public_packages.py` filtered candidates with `name.endswith((".pyd", ".so", ".dylib", ".dll"))`, and auditwheel keeps the soname, so the relocated file is `libonnxruntime-<hash>.so.1.28.0`. That does not end in `.so`, so it never reached the stem match.

The two halves of that check already disagreed: `_basename_stem`'s docstring gives `libonnxruntime-abc123.so.1` as an input it handles, while the filter feeding it rejected that exact shape. It passed until now only because the upstream tarball shipped a flat `libonnxruntime.so`, so the vendored copy happened to end in `.so`. This is a latent bug the first commit exposed rather than caused.

**Verification.** `scripts/test_validate_public_packages.py` already builds synthetic wheels, so this one is checkable without a Linux runner. I added a case with a `libonnxruntime-a1b2c3d4.so.1.28.0` sidecar under a manylinux wheel name. Reverting only the predicate reproduces CI's message verbatim:

```
PackageValidationError: runanywhere-...-manylinux_2_39_x86_64.whl:
  vendored sidecar libraries missing from runanywhere.libs/: ['libonnxruntime']
```

With the fix, 17/17 pass.

Two commits because they are two distinct defects with different blast radii: the workflow guard is CI-only, the validator predicate also governs what ships. Happy to split them if you would rather review separately.

ayaangazali:
Update: two of the three blockers are fixed here, and the third is not something this PR should decide.

**Where the job is now.** Each fix moved the failure one step later, which is the useful part:

| commit | outcome |
|---|---|
| before | hung in `auditwheel repair`, killed at the 1h30m timeout |
| `7f9de1a` symlink guard | repair completes in ~10s, fails in `Validate release boundary` |
| `56bef3a` binary-member filter | validate passes, fails in `Install + hermetic tests` |

The current error is:

```
ERROR: runanywhere-0.20.24-cp39-cp39-manylinux_2_39_x86_64.whl is not a supported wheel on this platform.
```

**Why, and why I stopped here.** pip is right to refuse. The job runs on `ubuntu-22.04`, which is glibc 2.35, and the wheel is tagged `manylinux_2_39`. On 2026-08-16 the same job tagged it `manylinux_2_35` and installed fine.

The tag moved because the vendored sherpa library moved. I pulled both assets and read the glibc symbol versions each one requires:

| library | upstream v1.13.2 | rac-desktop.4 (current) |
|---|---|---|
| `libonnxruntime` | GLIBC_2.16 | GLIBC_2.27 |
| `libsherpa-onnx-c-api` | GLIBC_2.17 | **GLIBC_2.38** |
| `libsherpa-onnx-cxx-api` | n/a | GLIBC_2.32 |
| `libonnxruntime_providers_shared` | n/a | GLIBC_2.2 |

So ORT is not the problem; `libsherpa-onnx-c-api.so` in the RunAnywhere desktop build needs glibc 2.38, up from 2.17. auditwheel then tags the wheel above what the runner provides, and the install step cannot pass on that runner no matter what this PR does. (I measured 2.38 as the highest symbol version; auditwheel settled on 2_39. Either way it is above 2.35.)

That leaves a choice I do not think a drive-by PR should make for you:

1. Build the desktop sherpa asset against an older glibc, keeping the published wheels installable on glibc 2.35 hosts, or
2. Move `python-linux` to `ubuntu-24.04` (glibc 2.39) and accept that published Linux wheels now require glibc 2.39.

Option 2 is a one-line change I am happy to push, but it silently raises the minimum glibc for everyone installing the wheel, so it is your call rather than mine.

Both commits here stand on their own regardless of which you pick: the symlink loop makes `auditwheel` hang forever, and the filter rejects any sidecar carrying an soname version. If you would rather land those two and track the glibc decision separately, say so and I will open an issue for it with this evidence.

ayaangazali:
One more red check appeared since my last comment, and it is infrastructure rather than this diff: `android-arm64`.

Its conclusion is `cancelled` with **zero failed steps**. The log shows it stuck in `apt-get update` against `archive.ubuntu.com` from 00:51 to 06:50, almost six hours, before the runner killed it:

```
00:51:36 Get:5 https://archive.ubuntu.com/ubuntu jammy-security InRelease [129 kB]
06:50:32 ##[error]The operation was canceled.
```

Nothing in this PR touches Android. I cannot re-run it myself (fork PR, no admin rights on the repo), so it needs either a maintainer re-run or the next push to this branch to clear.

ayaangazali:
Following up with something I should have found earlier: the workflow already documents why this cannot be fixed by changing the runner.

`pr-build.yml:335-340`, directly above `runs-on: ubuntu-22.04`:

> Linux sibling of python-windows: builds the runanywhere-python wheel via scikit-build-core, repairs it with auditwheel ... installs it into a clean interpreter, and runs the hermetic suite. **ubuntu-22.04 keeps the manylinux tag low (glibc 2.35) for a broadly-installable wheel.**

So the runner is a deliberate portability choice, not an accident. That settles the option I floated earlier: moving this job to `ubuntu-24.04` would make the check go green by raising the published wheel's floor to glibc 2.39, which is the exact thing this job exists to prevent. I am not going to push that.

The real conflict is on the asset side. #718 swapped the Linux prebuilt to the RunAnywhere desktop build, and its `libsherpa-onnx-c-api.so` requires **GLIBC_2.38** where the upstream tarball needed 2.17:

| library | upstream v1.13.2 | rac-desktop.4 |
|---|---|---|
| `libonnxruntime` | GLIBC_2.16 | GLIBC_2.27 |
| `libsherpa-onnx-c-api` | GLIBC_2.17 | **GLIBC_2.38** |

A wheel vendoring that library cannot be "broadly installable" at glibc 2.35, so the asset and the stated goal are in conflict regardless of CI. Rebuilding the desktop sherpa against an older glibc is the fix that keeps the goal intact, and that is yours to make, not something I can do from a fork.

One clarification on what this PR is worth in the meantime, since I do not think I stated it plainly: **the two commits here are necessary but not sufficient.** Merging them will not turn `python-linux` green. What they do is turn a 90-minute hang that burns a runner and reports nothing into a 15-minute failure that names the actual problem, and fix a validator that rejects any sidecar carrying an soname version. Both are correct independently of how you resolve the glibc question. The hang is also on `main` today, so this is worth landing on its own.

Happy to split the validator fix out if you would rather take that one first.

ayaangazali:
Closing the last option I had not tested, so this thread has a definite answer rather than another "still red".

I checked whether the pinned sherpa release ships a variant built against an older glibc, which would have made this a one-line change in `core/VERSIONS` and fully fixable here. It does not. `RunanywhereAI/sherpa-onnx` tag `v1.13.5-rac-desktop.4` has exactly three assets:

```
sherpa-onnx-v1.13.5-linux-aarch64-shared-rac-ort1.28.0.tar.bz2
sherpa-onnx-v1.13.5-linux-x64-shared-rac-ort1.28.0.tar.bz2
sherpa-onnx-v1.13.5-win-x64-shared-rac-ort1.28.0.tar.bz2
```

One Linux x64 build, the one requiring GLIBC_2.38. So all three in-repo levers are now ruled out: no alternative asset to pin, the runner is a documented portability choice I will not undo, and falling back to the upstream k2-fsa tarball would drop the Nemotron streaming decoder that 1.13.5 was pulled in for. Since the upstream build of the same version needs only GLIBC_2.17, this looks like the sherpa fork being built on a newer base image rather than anything in its source, which makes it a change in that repo's build workflow.

Nothing further from me on the glibc question. The two commits here stand on their own and are worth landing regardless, since the 90-minute `auditwheel` hang they fix is live on `main`.

ayaangazali:
Closing this: #757 removed the `python-linux` job, so both halves of this PR are superseded.

The symlink guard patched a step inside that job, and the line it fixed (`ln -sf libonnxruntime.so "$lib/libonnxruntime.so.1"`) no longer exists on main. This branch now conflicts with `pr-build.yml` for that reason, and rebasing it would just delete my own change.

For the record, since #757's message describes the hang as "a pre-existing issue unrelated to Python version choice": the cause was concrete. The RunAnywhere desktop sherpa asset ships `libonnxruntime.so -> .so.1 -> .so.1.28.0`, and that `ln -sf` overwrote the middle link so the pair pointed at each other. `auditwheel` then sat on an unreachable library until the 90-minute job timeout. Removing the job removes the symptom either way, and given the job was not a required check and Python is not published from `release.yml`, dropping it is a reasonable call that I am not going to argue with.

The second commit here is unrelated to the job and still technically correct: `validate_public_packages.py` filtered candidate sidecars with `endswith((".pyd", ".so", ".dylib", ".dll"))`, which misses `libonnxruntime-<hash>.so.1.28.0` because ELF puts the soname version after the suffix. Its own `_basename_stem` docstring already documents `libonnxruntime-abc123.so.1` as an input it handles, so the two halves of that check disagreed. It only ever mattered for Linux wheels, so with `python-linux` gone it guards nothing today and I am not going to reopen it as its own PR for the sake of it. If you ever restore a Linux wheel job, the fix plus its regression test is one commit on this branch and I am happy to bring it back.

Also worth flagging: several of my other open PRs were showing red only because of this job. Those cells should clear on their next run.

PR #737, opened 2026-08-18
ci: keep the rcli Homebrew tap current automatically

Siddhesh2377:
The tap served **rcli 0.20.10 from 14 July** while releases went on to 0.20.24, so `brew install runanywhereai/tap/rcli` handed people a month-old CLI. Nothing was wrong with the release: `release.yml` simply never calls `rcli/scripts/update-tap.sh`, so the formula only moved when somebody ran it by hand, and nobody had since July.

The formula itself is already fixed and merged (RunanywhereAI/homebrew-tap#1, now at 0.20.24). This is the part that stops it going stale again.

## Why a separate workflow and not a step in release.yml

`release.yml` creates the GitHub Release as a **draft**. `update-tap.sh` reads the published `.sha256` sidecars over HTTPS, so it cannot run inside that job — the assets are not downloadable yet. That is exactly why the script's own header says to run it manually after the draft is published.

`release: published` fires when a human publishes the draft, which is the first moment the tarballs are fetchable. Keeping it out of `release.yml` also means a tap failure can never fail a release.

## Degrades instead of failing

The tap is a different repository, so `GITHUB_TOKEN` cannot write to it. When `RCLI_TAP_TOKEN` is absent the job prints a warning naming the manual command and exits 0, mirroring how the rcli signing step degrades when the Developer ID secrets are missing. An unconfigured fork, or a release cut before the secret exists, stays green.

`workflow_dispatch` takes a version, so the tap can be re-pointed at any published release without cutting a new one.

## Verification

The install path in this PR is the one I exercised by hand before opening the tap PR: tapped the branch locally, `brew install runanywhereai/tap/rcli`, `brew test` green, and the installed binary reported `rcli 0.20.24 (commons 0.20.24)` with all five backends registering. The workflow's last step repeats that on a runner so a bad formula is caught where someone will see it.

## Needs one secret

`RCLI_TAP_TOKEN` — a token with `Contents: read and write` on `RunanywhereAI/homebrew-tap`. Until it exists this workflow no-ops with a warning, and the tap is updated by hand:

```bash
./rcli/scripts/update-tap.sh <version>
```

## Unrelated but worth knowing

The macOS rcli tarball is **ad-hoc signed and not notarized** — `Signature=adhoc`, `TeamIdentifier=not set`, `spctl` rejects it, no stapled ticket. `release.yml` has the whole signing path but it is gated on `RCLI_DEVELOPER_ID_CERT_P12_BASE64`, which is not configured; only the three `RCLI_NOTARY_*` secrets exist, so it takes the ad-hoc fallback. This does not affect `brew install` (Homebrew downloads are not quarantined, which is why the install above worked), but a browser download from the releases page is refused by Gatekeeper. Fixing it needs a Developer ID Application certificate from the existing Apple team, which only an Account Holder or Admin can create.

## Summary by CodeRabbit

* **Automation**
  * Added automated Homebrew tap updates when a release is published.
  * Added support for manually triggering tap updates with a specified version.
  * Added optional Homebrew installation and formula verification when credentials are configured.
  * Verification failures now clearly report as workflow failures, while unavailable credentials allow the workflow to complete without verification.

### DIFF
--- a/.github/workflows/rcli-tap.yml
+++ b/.github/workflows/rcli-tap.yml
@@ -1,0 +1,78 @@
+name: rcli Homebrew tap
+
+# Keeps RunanywhereAI/homebrew-tap's Formula/rcli.rb pointing at the newest
+# release. Nothing did this before, so the tap sat at 0.20.10 from 14 July while
+# releases went on to 0.20.24: `brew install runanywhereai/tap/rcli` handed
+# people a month-old CLI, which is what "brew install rcli doesn't work" was.
+#
+# This is a SEPARATE workflow on purpose. release.yml creates the GitHub Release
+# as a DRAFT, and update-tap.sh reads the published .sha256 sidecars over HTTPS,
+# so it cannot run inside that job — the assets are not downloadable yet. The
+# `release: published` event fires when a human publishes the draft, which is
+# exactly the moment the tarballs become fetchable. Keeping it out of
+# release.yml also means a tap failure can never fail a release.
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Version to point the tap at, without the leading v (e.g. 0.20.24)"
+        required: true
+        type: string
+
+permissions:
+  contents: read
+
+jobs:
+  update-tap:
+    name: Point the tap at the published release
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          persist-credentials: false
+
+      - name: Resolve version
+        id: version
+        env:
+          INPUT_VERSION: ${{ inputs.version }}
+          RELEASE_TAG: ${{ github.event.release.tag_name }}
+        run: |
+          version="${INPUT_VERSION:-${RELEASE_TAG#v}}"
+          if [ -z "${version}" ]; then
+            echo "::error::no version to publish"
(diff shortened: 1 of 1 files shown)


Siddhesh2377:
All three addressed in 222b439cb.

**`secrets` in a step `if` (critical)** — correct, and worse than it looked: that line started as `env.TAP_TOKEN` declared on the same step, which a step's own `if` also cannot read, and I "fixed" it into a `secrets` reference that is not permitted in `if` at all. Both spellings silently evaluated false, so the verification step would never have run. `TAP_TOKEN` now lives on the job's `env` and the condition reads `env.TAP_TOKEN`, which is the one form that resolves.

**`continue-on-error` on the brew verification** — removed. A check that cannot fail would have reported a broken formula as healthy, which is the failure mode this workflow exists to prevent.

**Serializing pushes** — added `concurrency: rcli-homebrew-tap` with `cancel-in-progress: false`. `update-tap.sh` does clone/commit/push with no rebase and no retry, so a release publish overlapping a manual dispatch could have one push rejected and leave the tap on the older release. Queued rather than cancelled on purpose: a cancelled run here means a tap that never got updated.

Re-validated the YAML after the change — concurrency present, `TAP_TOKEN` on job env, the condition resolving against `env`, and no `continue-on-error` on any step.

PR #740, opened 2026-08-18
fix(commons): report tool-calling output allocation failures

shubhamsinnh:
## Description
Several tool-calling output paths in `tool_calling.cpp` returned `RAC_SUCCESS` even when their required output allocation failed, while the matching non-empty paths already return `RAC_ERROR_OUT_OF_MEMORY`. This makes allocation failures indistinguishable from success:

- `rac_tool_call_parse_with_format` — the parse-fallback path returned `RAC_SUCCESS` with a `NULL` `clean_text` when the fallback `malloc` failed.
- `rac_tool_call_format_prompt_with_format` — the empty-definitions path returned `RAC_SUCCESS` with a `NULL` prompt when `malloc(1)` failed.
- `rac_tool_call_format_prompt_json_with_format` — the empty-tools JSON path returned `RAC_SUCCESS` with a `NULL` prompt when `malloc(1)` failed.
- `rac_tool_call_definitions_to_json` — the empty-definitions path returned `RAC_SUCCESS` with a `NULL` output when `malloc(3)` (the `"[]"` literal) failed.

Each now returns `RAC_ERROR_OUT_OF_MEMORY` on allocation failure, matching the existing non-empty-path behavior. No other behavior changes.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring

## Testing
- [x] Lint passes locally
- [ ] Added/updated tests for changes (no new tests per repo guidance)

Local:
- `git diff --check` — clean.
- `g++ -std=c++20 -fsyntax-only -I core/include -I core/src -I <scratch-nlohmann> core/src/features/llm/tool_calling.cpp` — exit 0 (nlohmann single header downloaded locally; `RAC_HAVE_PROTOBUF` not defined, so the protobuf-generated `tool_calling.pb.h` section is skipped).
- Full native build and `core/scripts/lint-cpp.sh` (clang-format) could not run on this Windows host; the `pr-build.yml` native build is the authoritative gate.

**Swift SDK / iOS Sample:**
- [ ] Tested on iPhone (Simulator or Device) — not run (no macOS host)
- [ ] Tested on iPad / Tablet — not run
- [ ] Tested on Mac (macOS target) — not run

- [ ] Tested on Android Phone (Emulator or Device) — not run (no Android toolchain on this host)
- [ ] Tested on Android Tablet — not run

- [ ] Tested on iOS — not run
- [ ] Tested on Android — not run

- [ ] Tested on iOS — not run
- [ ] Tested on Android — not run

**Web SDK / Web Sample:**
- [ ] Tested in Chrome (Desktop) — not run
- [ ] Tested in Firefox — not run
- [ ] Tested in Safari — not run
- [ ] WASM backends load (LlamaCpp + ONNX) — not run
- [ ] OPFS storage persistence verified (survives page refresh) — not run
- [ ] Settings persistence verified (localStorage) — not run

## Labels

**SDKs:**

**Sample Apps:**

The iOS, Android, Web, and Electron consumer apps live in their own
repositories (`RunanywhereAI/runanywhere-{ios,android,web,electron}`) — open
those PRs there.

## Checklist
- [x] Self-review completed
- [ ] Documentation updated (if needed) — no public API contract change

## Screenshots
Attach relevant UI screenshots for changes (if applicable):
- Mobile (Phone)
- Tablet / iPad
- Desktop / Mac

### DIFF
--- a/core/src/features/llm/tool_calling.cpp
+++ b/core/src/features/llm/tool_calling.cpp
@@ -331,6 +331,8 @@ static size_t skip_whitespace(const char* str, size_t pos, size_t len) {
     return pos;
 }
 
+enum class ToolParseStatus { kNoMatch, kSuccess, kOutOfMemory };
+
 /**
  * @brief Extract a JSON string value starting at the given position (must be after opening quote)
  *
@@ -339,10 +341,10 @@ static size_t skip_whitespace(const char* str, size_t pos, size_t len) {
  * @param len Length of input string
  * @param out_value Output: Allocated string value (caller must free)
  * @param out_end_pos Output: Position after closing quote
- * @return true if successful
+ * @return Structured parse status distinguishing malformed input from allocation failure
  */
-static bool extract_json_string(const char* str, size_t pos, size_t len, char** out_value,
-                                size_t* out_end_pos) {
+static ToolParseStatus extract_json_string(const char* str, size_t pos, size_t len,
+                                           char** out_value, size_t* out_end_pos) {
     std::string result;
     bool escaped = false;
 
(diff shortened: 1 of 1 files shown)


MohammedAlkindi:
The four public paths look right. The same shape survives in the `static bool` helpers that feed them: `extract_json_value:619`, `extract_tool_name_and_args:940,1030,1042,1049`, `parse_lfm2_format:1373` and `parse_default_format:1575` each `malloc`, guard with `if (ptr) { ... }`, and then report success.

Two are sharper. At 1373 and 1575 the allocation is `*out_clean_text` but the function ends `return *out_tool_name != nullptr;`, so a failed `clean_text` allocation returns `true` with a NULL `clean_text`. That is the same indistinguishable-from-success you are fixing above, one layer down.

Line numbers are against the base, all below your first hunk. Read only, no C++ build here.

shubhamsinnh:
> **[MohammedAlkindi](/MohammedAlkindi) **

Good catch, these helper-level allocation failures were still being collapsed into successful or no-match results. 

Addressed in 7e2469a34 by propagating a structured success/no-match/OOM status through the parser helpers and atomically cleaning partial outputs. Thanks for flagging this.

sanchitmonga22:
Thanks for this, @shubhamsinnh! Turning silent allocation failures in tool-call parsing into an explicit `RAC_ERROR_OUT_OF_MEMORY` (instead of a false "success" with NULL/partial output) is a real robustness fix, and the exception-mapping you added at the four public entry points closes off a class of process crashes too.

1. `core/src/features/llm/tool_calling.cpp:2305` -- when `parallel_tool_calls` is set and parsing a second-or-later envelope now returns `RAC_ERROR_OUT_OF_MEMORY` (a new possibility this PR introduces), the loop treats that the same as "no more envelopes" and `break`s; line 2326 then unconditionally reports `RAC_SUCCESS` with a truncated call list. That's exactly the failure mode this PR sets out to fix, just at one more call site. Propagating `next_rc` (return it, or set `result.set_error_code` before breaking) instead of discarding it should close it.

CodeRabbit also has an open point that still applies: https://github.com/RunanywhereAI/runanywhere-sdks/pull/740#pullrequestreview-5121454556 (the `found_key` leak-on-throw at line 725 -- Minor, not blocking, but worth a look while you're in this file).

Not blocking this PR: I opened #867 for the remaining unguarded proto entry points, and #868 for the strict-UTF-8 argument-parsing edge case -- you're welcome to pick either up if you're interested.

sanchitmonga22:
@coderabbitai review

shubhamsinnh:
Hi @sanchitmonga22 ,

Followed up on the remaining temporary-buffer ownership finding in `004b50f4c`.

- `args_value`, each flat-argument `kval`, and the non-standard-key `value` are now held by `std::unique_ptr` owners with `std::free` deleters before potentially throwing string operations.
- Object-valued argument buffers use `release()` only when ownership is intentionally transferred to `out_args_json`.
- Successful parsing behavior and output ownership remain unchanged; exception unwinding now releases these temporary allocations.

Validation passed with protobuf enabled under C++20, warnings treated as errors, plus diff whitespace validation.

PR #742, opened 2026-08-18
fix(rag): initialize the LLM service a RAG session creates

ayaangazali:
## What is wrong

#735 fixed this in `HttpServer::loadModel` and stated the rule in its comment: `rac_llm_create()` only routes to the plugin's `create` op, and a backend that defers the weight load to `initialize` (MLX) is not loaded when it returns. llama.cpp loads synchronously inside `create`, which is why the omission stays invisible until another backend reaches `generate`.

`rac_rag_session_create_proto` has the same shape. It creates the session's generation model:

```cpp
rc = rac_llm_create(llm_path.c_str(), &llm_handle);
```

and hands that handle straight to `RAGBackend`:

```cpp
session->backend = std::make_unique<RAGBackend>(backend_config, llm_handle, embed_handle, ...);
```

`rac_llm_initialize` does not appear anywhere in `rac_rag_proto_abi.cpp`. The handle is then driven directly through:

- `rag_pipeline_graph.cpp:85` (`rac_llm_generate`)
- `rag_pipeline_graph.cpp:357` (`rac_llm_generate_stream`)
- `rag_rerank.cpp:121` (`rac_llm_generate`)

So a RAG session backed by MLX fails at generation with "model is not loaded", for exactly the reason the server did.

## Why I am confident this is the same defect and not a deliberate difference

I checked every `rac_llm_create` call site in the tree. There are four, and the other three settle the question:

| call site | calls `rac_llm_initialize`? |
|---|---|
| `llm_module.cpp:335` (`llm_create_service`) | yes, and always has |
| `http_server.cpp:309` | yes, as of #735 |
| `rac_rag_proto_abi.cpp:781` | **no** |
| `core/tests/test_advanced_modality_proto_abi.cpp:1149` | test |

## What this does

Adds the `initialize` call, with the same failure handling as the sibling early-out just above it: destroy the LLM handle, release the embedding service created a few lines earlier, and publish the same `rag.sessionCreate` failure the other early-outs publish. 11 lines.

## Verification

Built and ran the full commons suite on this branch:

```
cmake -B build -DRAC_BUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Debug
cmake --build build -j$(sysctl -n hw.logicalcpu)   # clean, 0 errors
ctest --test-dir build -j4                          # 100% tests passed out of 100
```

No test added, and I want to be straight about why: reproducing the failure needs a backend whose `create` defers the load, which in practice means MLX. The commons suite has no such backend, so a test here would either need a fake plugin built for this one case or would pass on llamacpp whether or not the fix is present. #735 added no test for the same reason. What I did instead was enumerate the call sites above, so the claim rests on the invariant the other three already follow rather than on a test that cannot distinguish the two states.

## Summary by CodeRabbit

* **Bug Fixes**
  * RAG sessions now correctly initialize the language model service with the resolved model path.
  * Session creation now stops cleanly when model initialization fails.
  * Failed session creation also properly releases associated language model and embedding resources.
  * Clear failure handling prevents partially initialized RAG sessions from being created.

### DIFF
--- a/core/src/features/rag/rac_rag_proto_abi.cpp
+++ b/core/src/features/rag/rac_rag_proto_abi.cpp
@@ -785,6 +785,17 @@ rac_result_t rac_rag_session_create_proto(const uint8_t* config_proto_bytes,
             publish_failure(rc, "rag.sessionCreate", rac_error_message(rc));
             return rc;
         }
+        // `create` only builds the session; backends that defer the weight load
+        // (MLX) finish it in `initialize`, so generation would fail here with
+        // "model is not loaded". Same reason HttpServer::loadModel calls this
+        // (#735) and llm_create_service has always called it.
+        rc = rac_llm_initialize(llm_handle, llm_path.c_str());
+        if (rc != RAC_SUCCESS) {
+            rac_llm_destroy(llm_handle);
+            rac_embeddings_destroy(embed_handle);
+            publish_failure(rc, "rag.sessionCreate", rac_error_message(rc));
+            return rc;
+        }
     }
 
     try {


ayaangazali:

Still reproduces on `main`. `rac_rag_proto_abi.cpp:781` creates the LLM service and never initializes it:

```cpp
rc = rac_llm_create(llm_path.c_str(), &llm_handle);
```

`grep -c rac_llm_initialize` in that file is still 0.

Two things since I opened this make the case tighter rather than weaker:

`core/src/server/http_server.cpp` has the same pattern and already carries the fix, with a comment describing exactly this failure:

> `rac_llm_create()` only routes to the plugin and calls its `create` op [...] Some backends (e.g. MLX, whose `create` just registers a session and defers the actual weight load to `initialize`) do not finish loading until `rac_llm_initialize()` is called explicitly. llama.cpp's `create` happens to load synchronously, which made this omission invisible until an MLX model hit `generate` here and failed with "model is not loaded".

And the embeddings side of this very function is already safe for the same reason: `rac::embeddings::create_service` initializes internally, and its comment records that skipping it "made RAG report 'Embedding provider returned an empty vector'" and "fail 100% of document Q&A on the Apple SDKs".

I also checked the rest of the tree rather than assuming this was the only one: every `*_module.cpp` create path (llm 335, vlm 199, stt 607, tts 287) calls its `*_initialize` immediately after create. RAG's LLM is the last one that does not.

ayaangazali:
Still reproduces on `c11f78ef4`, and there is now some cross-PR context worth having.

The gap is unchanged: `rac_rag_proto_abi.cpp:781` calls `rac_llm_create` and nothing calls `rac_llm_initialize`, so a backend that defers the weight load finishes create with an unloaded model.

```
$ grep -n "rac_llm_initialize\|rac_llm_create" core/src/features/rag/rac_rag_proto_abi.cpp
781:        rc = rac_llm_create(llm_path.c_str(), &llm_handle);
```

Both other places that build an LLM service still pair the two calls, which is the precedent this PR follows:

```
core/src/features/llm/llm_module.cpp:342:    result = rac_llm_initialize(*out_service, model_id);
core/src/server/http_server.cpp:322:    rc = rac_llm_initialize(llmHandle_, modelPath.c_str());
```

The new context: I opened two more PRs today that touch this same file, #823 (sampling defaults in `execute_rag_query`) and #824 (a protobuf-off build break at the includes). They are disjoint from this one, at lines 422 and 27-46 against this PR's 785, and I checked they are not merely theoretically compatible:

```
$ git checkout -B probe origin/main
$ git merge fork/fix/rag-llm-initialize && git merge fork/fix/rag-generation-defaults && git merge fork/fix/rag-proto-off-build
merged OK: fix/rag-llm-initialize
merged OK: fix/rag-generation-defaults
merged OK: fix/rag-proto-off-build
```

So they can land in any order without a rebase on my side. Mentioning it only because three open PRs on one file is the kind of thing that is annoying to discover mid-review, not to bundle them.

Still merges clean and CI is green here.

ayaangazali:
Still reproduces on `45e91276e` (0.20.36): `rac_rag_proto_abi.cpp:781` calls `rac_llm_create` and nothing calls `rac_llm_initialize`, while `llm_module.cpp:342` and `http_server.cpp:322` both pair them.

Updating the merge note I left here earlier, because I checked it properly this time rather than pairwise.

When I last commented I had verified this PR against the two other RAG ones. Since then the set grew, and three of my open PRs now touch this one file (#742, #823, #824) while four touch `core/tests/test_solution_runner.cpp`. Pairwise-against-main cleanliness does not cover that: GitHub reported every one of them mergeable while two actually collided with each other.

I found one real conflict that way (#812 and #836 had both added a test at the same insertion point, since fixed) and then re-ran the probe across everything I have open:

```
merged clean: 42 / 42
conflicts: NONE
configure exit=0   build exit=0, 0 errors
ctest: 100% tests passed out of 103
```

So the combined tree of all 42 builds and passes, and this one can land in any order relative to the rest. Nothing in this PR changed; I am replacing an earlier claim that rested on a narrower check.

sanchitmonga22:
> Replying to [this comment](https://github.com/RunanywhereAI/runanywhere-sdks/pull/742#issuecomment-5415323899)

That makes sense -- thanks for confirming the rebase and for pointing to http_server.cpp's identical create->initialize fix (#735) as precedent; this PR mirrors that pattern correctly. Merging it now.

sanchitmonga22:
> Replying to [this comment](https://github.com/RunanywhereAI/runanywhere-sdks/pull/742#issuecomment-5482975278)

Thanks for flagging the overlap with your other PRs -- good to know, and no bundling needed here. This one's landing on its own now.

sanchitmonga22:
> Replying to [this comment](https://github.com/RunanywhereAI/runanywhere-sdks/pull/742#issuecomment-5565834243)

Appreciate you re-checking the merge-safety note and running the fuller 42-PR probe. Sorry for the wait -- merging this now.

sanchitmonga22 (approved):
Thanks a lot for this, @ayaangazali! This fixes RAG sessions failing every answer with "model is not loaded" (MLX) or a not-initialized error (NeuRT), because the LLM handle a RAG session creates was never initialized after `rac_llm_create`.

Checked: CodeRabbit reviewed the latest commit · CI green · built and linted locally merged into `main` (C++ commons/engines core/**, fast validation gates) · two independent code reviews.

Follow-up, not blocking: #871 (having RAG reuse the lifecycle-loaded LLM instead of creating its own) -- you're welcome to pick it up.

Merging now -- really appreciate the contribution, and thanks for the patience through three check-ins while this waited on a maintainer look!

PR #744, opened 2026-08-18
fix(jni): refuse a platform adapter missing a required method

ayaangazali:
## What is wrong

`racSetPlatformAdapter` resolves nine required method ids and checks none of them:

```cpp
g_method_log         = env->GetMethodID(adapterClass, "log", "(ILjava/lang/String;Ljava/lang/String;)V");
g_method_file_exists = env->GetMethodID(adapterClass, "fileExists", "(Ljava/lang/String;)Z");
...
g_method_now_ms      = env->GetMethodID(adapterClass, "nowMs", "()J");
```

`GetMethodID` throws `NoSuchMethodError` when a lookup misses, so a Kotlin-side rename or signature change leaves **both** a null id and a pending exception. Today neither is noticed, and the two failures compound:

1. The pending exception is absorbed by the first optional-probe check a few lines below, which attributes it to that probe and nulls `g_method_file_list_directory`:

```cpp
g_method_file_list_directory = env->GetMethodID(adapterClass, "fileListDirectory", ...);
if (env->ExceptionCheck()) {
    env->ExceptionClear();
    g_method_file_list_directory = nullptr;
}
```

2. The null required id is stored into `g_c_adapter` anyway and later passed to `Call*Method`, which is undefined behaviour rather than a clean failure.

The observable result of an adapter/SDK build mismatch is therefore "directory probes unavailable" followed by a crash inside an unrelated file or secure-storage call, with nothing pointing at the real cause.

## Why the optional probes are not the same case

That degrade-on-miss handling is deliberate for `fileListDirectory` / `isNonEmptyDirectory`, and the comment above them says so: older host apps miss those ids and commons falls through to its NULL-callback branches. The nine above are required by the adapter contract in `rac_platform_adapter.h`, so a miss is a build mismatch to report, not a capability to degrade.

## What this does

Clears the pending exception (so later JNI calls in this function are safe), then refuses to install a partially resolved adapter and returns `RAC_ERROR_INITIALIZATION_FAILED`, the code this function already returns when `rac_set_platform_adapter` fails. The log line names the required set so the mismatch is findable.

This makes an already-broken configuration fail at registration instead of crashing later. An app missing one of these methods does not work today either; it just fails somewhere unrelated.

## Verification, and the limit

This file is Android-only and not part of the macOS commons build, so the usual `cmake`/`ctest` gate does not cover it. I type-checked it directly against the JDK's `jni.h`:

```
clang++ -fsyntax-only -std=c++20 \
  -I$JAVA_HOME/include -I$JAVA_HOME/include/darwin \
  -Icore/include -Icore/src -Icore/src/generated/proto -I<nlohmann> \
  core/src/jni/runanywhere_commons_jni.cpp     # clean
```

(`rac_defaults_generated.h` comes from `idl/codegen/generate_cpp_defaults.py`; that output is gitignored and `git status` shows only the one source file.)

I confirmed the check actually bites by introducing a deliberate typo inside the new block, which it caught at that line, so a clean run means the edit type-checks rather than the compiler skipping the file. `kotlin-android` is the authoritative gate and compiles it for real.

No test added: exercising this needs a Java adapter class built with a deliberately wrong signature, loaded through a live JNI attach, which the commons suite cannot host.

Related but separate from #743, which fixes the same *family* of omission (unchecked pending exception) at the four app-callback sites in the two hybrid-router JNI files. This one is the adapter-registration path and a different failure mode, so I kept them apart.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of platform adapters during initialization.
  * Initialization now stops with a clear failure when required adapter capabilities are unavailable, preventing partially configured adapters.
  * Improved handling of setup errors during initialization for more reliable diagnostics.
  * Improved directory listing reliability when invalid inputs, memory limitations, or underlying errors occur, with safer cleanup of partial results.


ayaangazali:
Update on the verification caveat in the description: CI has now compiled this file for real, so the syntax-check limitation no longer applies.

`android-arm64` passed on this PR (14m4s) and its log shows the changed translation unit built with the NDK toolchain and linked:

```
[720/1056] Building CXX object core/src/jni/.../runanywhere_commons_jni.cpp.o
[736/1056] Linking CXX shared library core/src/jni/librunanywhere_jni.so
```

`kotlin-android` is still running and will exercise the same object through the AAR build. Every other check is green; nothing here is waiting on me.

ayaangazali:

Still reproduces. On `main` the nine required `GetMethodID` lookups run straight into the optional directory probes with nothing between them:

```cpp
g_method_now_ms = env->GetMethodID(adapterClass, "nowMs", "()J");
// Optional Kotlin-side directory probes. Method lookup is
// best-effort ...
g_method_file_list_directory = env->GetMethodID(adapterClass, "fileListDirectory", ...
```

`grep -c "platform adapter is missing a required method"` is still 0.

That ordering is what makes it quiet rather than loud. `GetMethodID` throws `NoSuchMethodError` on a miss, so a Kotlin-side rename leaves both a null id and a pending exception. The pending exception is then absorbed by the first optional probe's `ExceptionCheck`, which blames the probe and nulls it, and the null id from the required lookup is handed to `Call*Method` later.

The distinction this PR relies on is the one the existing comment already draws: `fileListDirectory` and `isNonEmptyDirectory` are documented as best-effort with NULL-callback fallbacks in commons, while the other nine are required by the adapter contract in `rac_platform_adapter.h`. A miss there is a build mismatch to report, not a capability to degrade, so it refuses to install the adapter and returns `RAC_ERROR_INITIALIZATION_FAILED`.

ayaangazali:

`runanywhere_commons_jni.cpp:1685` onwards resolves the platform-adapter methods with `GetMethodID` and does not check the results:

```cpp
g_method_log         = env->GetMethodID(adapterClass, "log", "(ILjava/lang/String;Ljava/lang/String;)V");
g_method_file_exists = env->GetMethodID(adapterClass, "fileExists", "(Ljava/lang/String;)Z");
g_method_file_read   = env->GetMethodID(adapterClass, "fileRead", "(Ljava/lang/String;)[B");
```

`GetMethodID` throws `NoSuchMethodError` on a miss, so a Kotlin-side rename or signature change leaves both a null `jmethodID` and a pending exception. Neither is noticed today: the pending exception is absorbed by the first optional-probe `ExceptionCheck` further down, which then misattributes the failure to that probe and nulls it, and the null id is later handed to `Call*Method`, which is undefined behaviour.

So a build mismatch between the Kotlin adapter and this JNI layer presents as a mysteriously missing optional capability plus, eventually, a crash in an unrelated call, rather than as the version-skew error it is.

The nine methods this checks are the ones `rac_platform_adapter.h` marks required, deliberately not `fileListDirectory` / `isNonEmptyDirectory`, which are genuinely optional and should keep degrading.

Merges clean, CI green.

sanchitmonga22:
Thanks for this, @ayaangazali! Refusing a platform adapter that's missing a required method instead of quietly installing it is exactly the right instinct, and re-checking the repro against main after every rebase kept this honest.

Before we merge, a couple of things to sort out:

1. `core/src/jni/runanywhere_commons_jni.cpp:1668-1717` -- the refusal isn't transactional yet: the previous adapter's global ref is deleted and the new (invalid) one is installed into `g_platform_adapter`, and all nine method ids are overwritten, *before* the missing-method check runs. So a refusal still leaves `g_platform_adapter` pointing at the rejected object. `racAuthInit` only checks `g_platform_adapter == nullptr` (line 4615), so it treats the refused adapter as registered. Could you resolve the ids into locals against `adapter` first, validate them, and only then commit the globals?
2. `core/src/jni/runanywhere_commons_jni.cpp:1706` -- `ExceptionCheck`/`ExceptionClear` runs once, after all nine `GetMethodID` calls, so a miss on anything before `nowMs` leaves a pending `NoSuchMethodError` while later lookups in the same run still execute. This file's own `jniClearPendingException` helper documents that ART/CheckJNI aborts a JNI call made with a pending exception, so under CheckJNI this can crash instead of returning the new error code. Clearing right after each lookup (like the optional `fileListDirectory`/`isNonEmptyDirectory` probes just below already do) would close that.

CodeRabbit's thread raises the same transactional-safety point as (1): https://github.com/RunanywhereAI/runanywhere-sdks/pull/744#discussion_r3809009717 -- worth a look even though it now shows as addressed, since that annotation landed on the same commit it's describing.

Not blocking this PR: I opened #872 for the Kotlin side silently continuing init after a refused registration, and #873 for the refusal message not reaching Android logcat -- you're welcome to pick either up if you're interested.

Once (1) and (2) are in we'll take another look. Thanks again!

ayaangazali:
Both taken, pushed above. You were right that the check I added did not actually refuse anything.

**Transactional now.** The nine ids resolve into locals against `adapter`, and the globals are written only once every one of them is non-null. On refusal nothing is committed, so `g_platform_adapter` stays null from the teardown further up, which is exactly the state `racAuthInit` already reads correctly at line 4615. Previously the global ref and all nine ids were committed first, so a refusal installed the rejected object and `racAuthInit` saw a non-null pointer and carried on.

**One ExceptionCheck per lookup.** You are right that the single check after all nine was wrong, and it is worse than just mis-attributing the error: after a miss on `log`, the following eight `GetMethodID` calls were running with a `NoSuchMethodError` already pending, which is undefined. Each lookup now clears its own, via a small local helper so the nine call sites stay readable.

One extra thing I noticed while restructuring and fixed rather than leaving: the refusal path returned without `DeleteLocalRef(adapterClass)`, so it leaked the local ref on every refusal. It releases it on both paths now.

I did not touch the null-adapter branch above. It tears down first and then returns `RAC_ERROR_INVALID_ARGUMENT`, which looks like the same shape, but it leaves `g_platform_adapter` as null rather than dangling, so `racAuthInit` reports "not registered" correctly. Happy to move it if you would rather a null argument left a working adapter in place, but that is a behaviour change rather than a fix so I left it alone.

**How far I verified this.** `core/src/jni` is not part of the macOS build here, and I could not run the Android build either: `./gradlew compileDebugKotlin` fails in this checkout at `:modules:runanywhere-core-onnx:downloadJniLibs`, "Failed to download or validate RABackendONNX-android-arm64-v8a-v0.20.36.zip", which is unrelated to this change. So I syntax-checked the translation unit directly against the real headers instead:

```
g++ -fsyntax-only -std=c++20 -I core/include -I core/src \
    -I <nlohmann> -I <jdk>/include -I <jdk>/include/darwin \
    -I <protobuf> -I core/src/generated/proto \
    core/src/jni/runanywhere_commons_jni.cpp
exit 0
```

That is a compile of this file, not a run. The behaviour above is reasoned from the code, not observed on a device.

PR #745, opened 2026-08-19
fix(rcli): fail when `voice --output` cannot be written

ayaangazali:
## What is wrong

`rcli voice --output reply.wav` exits 0 when it cannot write the file.

`cmd_voice.cpp:115-121`:

```cpp
std::ofstream file(output, std::ios::binary);
file.write(result.synthesized_audio().data(), ...);
if (file.good()) {
    reply_path = output;
} else {
    out::status_line("warning: cannot write " + output);
}
```

`exit_code` is never touched, so the command returns 0. In `--json` mode the object then reports `"reply_audio": ""`, which a consumer cannot distinguish from a turn that produced no audio at all. A caller doing

```sh
rcli voice --output reply.wav && play reply.wav
```

proceeds against a file that is not there, with only a line on stderr that a script capturing stdout never sees.

## Why this is the outlier and not a policy

`main.cpp` states the contract in its header comment:

> Exit codes: 0 success, 1 runtime/SDK error, 2 usage error.

Both sibling commands that write a user-requested output file already follow it:

| command | write failure | exit |
|---|---|---|
| `tts` (`cmd_tts.cpp:105`) | `out::error_line(error)` | **1** |
| `image` (`cmd_image.cpp:156`) | returns false with `"cannot write " + out_path` | **1** |
| `voice` (`cmd_voice.cpp:119`) | `out::status_line("warning: ...")` | **0** |

`tts` is the closest analogue: synthesis succeeded, only the file write failed, and it still exits 1. There is no comment anywhere justifying the difference, so this reads as an oversight rather than an intentional split.

## What this does

Reports the failure through `error_line` and sets `exit_code = 1`, matching `tts`. Seven lines, one of them the actual change.

The turn's text output is unaffected: transcription and response still print, so a user running interactively loses nothing and a script now learns that the file it asked for does not exist.

If you would rather keep exit 0 because the voice turn itself succeeded, the alternative fix is to leave the code and make the JSON distinguishable instead (a `reply_audio_error` field). I went with matching `tts` because the exit-code contract is documented and two of three commands already implement it that way.

## Verification

Built with the CLI enabled and ran the whole suite:

```
cmake -B build-cli -DRAC_BUILD_CLI=ON -DRAC_DESKTOP_ADAPTER=ON -DRAC_BUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Debug
cmake --build build-cli -j$(sysctl -n hw.logicalcpu)   # 0 errors; cmd_voice.cpp.o rebuilt, rcli linked
ctest --test-dir build-cli                            # 100% tests passed out of 107
```

(`RAC_BUILD_CLI` defaults to OFF, so a plain `cmake -B build -DRAC_BUILD_TESTS=ON` does not compile this file at all. I mention it because I initially built without it and had to redo the gate; `-DRAC_DESKTOP_ADAPTER=ON` is also required or rcli's CMakeLists fails outright.)

No test added: exercising this needs a loaded voice-agent model plus an unwritable destination, which the rcli suite does not set up. The change is a one-line contract alignment with `cmd_tts`, and the enumeration above is what the claim rests on.

## Summary by CodeRabbit

* **Bug Fixes**
  * Voice commands now clearly report output-write failures.
  * The command exits with a failure status and suppresses result output when voice output cannot be written.

### DIFF
--- a/rcli/src/commands/cmd_voice.cpp
+++ b/rcli/src/commands/cmd_voice.cpp
@@ -118,7 +118,13 @@ int run_voice(const GlobalOptions& options, const std::string& input, const std:
             if (file.good()) {
                 reply_path = output;
             } else {
-                out::status_line("warning: cannot write " + output);
+                // A requested --output that cannot be written is a runtime error,
+                // not a warning: `rcli tts --output` already exits 1 for exactly
+                // this (cmd_tts.cpp), and a caller doing
+                // `rcli voice --output reply.wav && play reply.wav` otherwise
+                // proceeds against a file that is not there.
+                out::error_line("cannot write " + output);
+                exit_code = 1;
             }
         }
 
--- a/rcli/src/commands/cmd_voice.cpp
+++ b/rcli/src/commands/cmd_voice.cpp
@@ -128,19 +128,27 @@ int run_voice(const GlobalOptions& options, const std::string& input, const std:
             }
         }
 
-        if (options.json) {
-            out::JsonWriter json;
-            json.begin_object()
-                .field("transcription", result.transcription())
-                .field("response", result.assistant_response())
-                .field("reply_audio", reply_path)
-                .end_object();
-            out::result_line(json.str());
-        } else {
-            out::result_line("you   " + result.transcription());
-            out::result_line("agent " + result.assistant_response());
-            if (!reply_path.empty()) {
-                out::result_line("audio " + reply_path);
+        // Only a turn that fully succeeded prints a result. cmd_tts.cpp emits
+        // its result in the `else if` branch of the same write check, so a
+        // failed --output write there produces an error and nothing else; a
+        // success-shaped line next to exit 1 would let
+        // `rcli voice --json -o reply.wav | jq -r .reply_audio` read an empty
+        // path as if the turn had produced no audio.
+        if (exit_code == 0) {
+            if (options.json) {
+                out::JsonWriter json;
+                json.begin_object()
+                    .field("transcription", result.transcription())
+                    .field("response", result.assistant_response())
+                    .field("reply_audio", reply_path)
+                    .end_object();
+                out::result_line(json.str());
+            } else {
+                out::result_line("you   " + result.transcription());
+                out::result_line("agent " + result.assistant_response());
+                if (!reply_path.empty()) {
+                    out::result_line("audio " + reply_path);
+                }
             }
         }
     }


ayaangazali:
Good catch, and it was right. Setting `exit_code` was only half the change: the result block after it ran unconditionally, so a failed write still printed

```json
{"transcription":"...","response":"...","reply_audio":""}
```

on stdout next to exit 1. `jq -r .reply_audio` reads that as "the turn produced no audio", which is a different thing from "the file could not be written".

What settled the shape of the fix is that `cmd_tts.cpp` already answers it. Its result is emitted in the `else if` branch of the same write check:

```cpp
        if (!wav::write_wav_f32(params.output, float_samples, sample_count, result.sample_rate,
                                &error)) {
            out::error_line(error);
            exit_code = 1;
        } else if (options.json) {
```

so a failed `--output` write there prints the error and nothing else. `cmd_voice` now does the same by guarding the emission on `exit_code == 0`. The earlier `voice turn failed` path in this function was already consistent with that, since it lives in the other arm of the branch.

I did consider keeping the transcription and response, since unlike `tts` those are real results that do not depend on the file. I decided against it: `--output` was requested and could not be honoured, and inventing a partial-success output shape here would put `cmd_voice` on its own convention for no clear gain.

Also rebased onto `c4a8fb283`. Rebuilt with `-DRAC_BUILD_CLI=ON -DRAC_DESKTOP_ADAPTER=ON`, `cmd_voice.cpp` compiles and `rcli` links. I have not exercised the failure path end to end, since reaching it needs a voice turn to fully succeed against real STT, LLM and TTS models first, and this host does not have them staged. The change is a control-flow guard, so the review here is reading it against `cmd_tts`.

ayaangazali:
Closing this: it is superseded rather than wrong.

#776 retired the in-tree CLI yesterday. `rcli/` is gone from this repo, the product now lives in `RunanywhereAI/RCLI` and consumes the C++ desktop kit, and the root `CMakeLists.txt` turns the old switch into a hard error:

```cmake
option(RAC_BUILD_CLI "Retired: in-tree rcli was moved to RunanywhereAI/RCLI. Leaving ON is a hard error." OFF)
```

So the file this patches, `rcli/src/commands/cmd_voice.cpp`, no longer exists here and there is nothing left for this PR to merge into.

The defect itself is still real, just in the other repo now. For whoever picks it up there: `voice --output <path>` set `exit_code = 1` when the write failed but then fell through to the result block, so it printed a success-shaped line next to the failure, `{"transcription":"...","response":"...","reply_audio":""}` on stdout. `jq -r .reply_audio` reads that as "the turn produced no audio", which is a different thing from "the file could not be written". `cmd_tts.cpp` already had the right shape, emitting its result in the `else if` branch of the same write check, so the fix was to guard the emission on `exit_code == 0`.

Happy to reopen it against `RunanywhereAI/RCLI` if that is useful, otherwise this is here for the record.

PR #746, opened 2026-08-20
feat(catalog): add Qwen3.5 rows to the Flutter and RN examples

Siddhesh2377:
Rebuilds the catalog against the current model generation, following the rules agreed in Slack: drop superseded families, keep only 1-bit, Q4 and Q8, and add the newest families across each modality.

Removed: Qwen2, Qwen2.5, Qwen3, LFM2, Llama 2, Llama 3.2, Mistral 7B and SmolLM2 rows, plus the Q2 and Q6 rows that fell outside the allowed quantizations.

Added from a verified list: Qwen3.5 (0.8B through 9B), Gemma 4, Granite 4.1, LFM2.5, the PrismML Bonsai 1-bit family, and Maple Preview.

Every download size is the real `Content-Length` of the asset rather than a rounded guess, and every URL was checked with a HEAD request. Several sizes in the old catalog were wrong by enough to break the download progress bar.

Private rows are untouched. HNPU/QHexRT and NeuRT bundles are exactly as they were.

**Maple Preview is listed as 1-bit, not Q4.** Its experts are already ternary in storage, so upstream ships only TQ1_0 and TQ2_0 and calls the uniform Q4_K_M a quantizer sanity check.

**LoRA registration is removed for now.** The only adapter we ship is trained for qwen2.5-0.5b, which this catalog no longer carries, so registering it would offer an adapter with no compatible base. A replacement adapter trained on a base we do ship is planned, and re-adding is a small edit.

## Summary by CodeRabbit

* **New Features**
  * Added expanded model catalog options, including Qwen3.5, LFM2.5 Thinking, Granite 4.1, Qwen3.5 9B, and Bonsai variants.
  * Updated Flutter and React Native examples with refreshed model availability and download options.
* **Updates**
  * Refreshed RAG and voice agent examples to use the LFM2.5 230M model.
  * Removed outdated model and unsupported adapter options from example catalogs.

### DIFF
--- a/bindings/flutter/example/lib/core/services/model_catalog_bootstrap.dart
+++ b/bindings/flutter/example/lib/core/services/model_catalog_bootstrap.dart
@@ -116,6 +116,33 @@ abstract final class ModelCatalogBootstrap {
       memoryRequirement: 2800000000,
       supportsThinking: true,
     );
+    await _registerLLM(
+      id: 'qwen3.5-0.8b-q4_k_m',
+      name: 'Qwen3.5 0.8B Q4_K_M',
+      url:
+          'https://huggingface.co/unsloth/Qwen3.5-0.8B-GGUF/resolve/main/Qwen3.5-0.8B-Q4_K_M.gguf',
+      framework: InferenceFramework.INFERENCE_FRAMEWORK_LLAMA_CPP,
+      memoryRequirement: 900000000,
+      supportsThinking: true,
+    );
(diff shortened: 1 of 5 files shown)


Siddhesh2377:
Closing: this changed the in-tree contributor harnesses under bindings/*/example, but the apps that needed the catalog refresh are the standalone flutter-starter-example and react-native-starter-app repos. Reopening the work there.

PR #755, opened 2026-08-21
fix(llamacpp): honour the same load options on the VLM create path

ayaangazali:
## What is wrong

a46517cf1 taught the LLM path to read `context_length` and `accelerator_policy` out of the `config_json` load options "instead of dropping them". The VLM adapter one file over still drops them.

`llamacpp_vlm_create_impl` already receives `config_json` and already parses it, but only for `"mmproj_path"`, and then passes a null config:

```cpp
rac_result_t rc = rac_vlm_llamacpp_create(model_id, mmproj_path, nullptr, &backend_handle);
```

`rac_vlm_llamacpp_config_t` carries exactly the knobs that were just wired up on the LLM side:

```c
int32_t context_size;   // 0 = auto-detect
int32_t gpu_layers;     // -1 = all layers on GPU
```

So a VLM load that asks for a context length, or pins placement to CPU, is silently ignored and the backend runs on `RAC_VLM_LLAMACPP_CONFIG_DEFAULT`. That is the same defect the LLM path just had, reached through the same ABI, in the same engine.

## What this does

Parses the same keys in the block that already parses `mmproj_path`, with the same precedence and mapping the LLM helper uses:

- `accelerator_policy` wins over the deprecated `use_gpu`, matching the proto comment that calls `use_gpu` an adapter onto it
- CPU maps to `gpu_layers = 0`, GPU to `gpu_layers = -1`
- AUTO and NPU are left alone, for the reasons `parse_load_options` already gives

The config is passed only when something was actually requested. A load with no placement or context options keeps passing `nullptr`, so today's behaviour is unchanged, which mirrors the "returns false when there was nothing to say" contract on the LLM side.

I mirrored the two `AcceleratorPolicy` constants locally with the same rationale the LLM file gives, rather than pulling protobuf into this TU.

## The one judgement call

On `ACCELERATOR_POLICY_CPU` I also clear `use_gpu_vision`. The LLM config has no vision encoder, so a strict mirror would leave it set, and a caller that explicitly asked for CPU would still get the vision tower on the GPU. I think honouring the request across both compute paths is what CPU means here, but it is the one place I went past a literal copy of a46517cf1 — say the word and I will drop it to `gpu_layers` only.

## Verification

Built with backends enabled so this TU actually compiles, then ran the whole suite:

```
cmake -B build -DRAC_BUILD_TESTS=ON -DRAC_BUILD_BACKENDS=ON -DCMAKE_BUILD_TYPE=Debug
cmake --build build -j6     # 0 errors; rac_llamacpp_vlm_ops.cpp rebuilt
ctest --test-dir build -j4  # 100% tests passed out of 127
```

No test added: exercising this needs a loaded VLM plus a way to observe llama.cpp's context/placement, which the commons suite has no fixture for. The claim rests on the enumeration above, and the sibling it mirrors landed yesterday.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring model context length during loading.
  * Added CPU and GPU accelerator placement options.
  * Retained compatibility with the existing GPU configuration setting.

* **Bug Fixes**
  * Invalid loading configuration values now fall back to safe defaults with a warning.
  * Existing behavior is preserved when no context or accelerator options are provided.

### DIFF
--- a/engines/llamacpp/rac_llamacpp_vlm_ops.cpp
+++ b/engines/llamacpp/rac_llamacpp_vlm_ops.cpp
@@ -116,9 +116,16 @@ static void llamacpp_vlm_vtable_destroy(void* impl) {
     rac_vlm_llamacpp_destroy(impl);
 }
 
-// `create` adapter for llama.cpp VLM. Parses the optional "mmproj_path" key
-// from config_json (so VLM's 2-path create signature maps cleanly into the
-// uniform rac_vlm_service_ops_t::create slot).
+// AcceleratorPolicy wire values, mirrored from idl/model_types.proto for the
+// same reason rac_backend_llamacpp_register.cpp mirrors them: they arrive as
+// JSON numbers over the config_json ABI and this engine links neither protobuf
+// nor the generated tree. Keep in step with ACCELERATOR_POLICY_CPU/GPU.
+constexpr int kAcceleratorPolicyCpu = 2;  // ACCELERATOR_POLICY_CPU
+constexpr int kAcceleratorPolicyGpu = 3;  // ACCELERATOR_POLICY_GPU
+
+// `create` adapter for llama.cpp VLM. Parses "mmproj_path" (so VLM's 2-path
+// create signature maps cleanly into the uniform rac_vlm_service_ops_t::create
+// slot) plus the same load knobs the LLM path honours.
 rac_result_t llamacpp_vlm_create_impl(const char* model_id, const char* config_json,
                                       void** out_impl) {
     if (!model_id || !out_impl) {
@@ -128,6 +135,10 @@ rac_result_t llamacpp_vlm_create_impl(const char* model_id, const char* config_j
 
     std::string mmproj_path_owned;
     const char* mmproj_path = nullptr;
+    rac_vlm_llamacpp_config_t config = RAC_VLM_LLAMACPP_CONFIG_DEFAULT;
+    // Stays false when config_json said nothing about placement or context, so
+    // the create call below keeps passing nullptr and behaviour is unchanged.
+    bool have_config = false;
     if (config_json && config_json[0] != '\0') {
         try {
             auto json = nlohmann::json::parse(config_json);
@@ -136,6 +147,30 @@ rac_result_t llamacpp_vlm_create_impl(const char* model_id, const char* config_j
                 mmproj_path = mmproj_path_owned.c_str();
                 RAC_LOG_DEBUG(LOG_CAT, "Parsed mmproj_path from config_json: %s", mmproj_path);
             }
+            if (json.contains("context_length") && json["context_length"].is_number_integer()) {
+                config.context_size = json["context_length"].get<int32_t>();
+                have_config = true;
+            }
+            // accelerator_policy wins over the deprecated use_gpu when both are
+            // present, matching the proto comment that calls use_gpu an adapter
+            // onto it.
+            int policy = 0;
+            if (json.contains("accelerator_policy") &&
+                json["accelerator_policy"].is_number_integer()) {
+                policy = json["accelerator_policy"].get<int>();
+            } else if (json.contains("use_gpu") && json["use_gpu"].is_boolean()) {
+                policy = json["use_gpu"].get<bool>() ? kAcceleratorPolicyGpu
+                                                     : kAcceleratorPolicyCpu;
+            }
+            if (policy == kAcceleratorPolicyCpu) {
+                config.gpu_layers = 0;       // every layer stays on the CPU
+                config.use_gpu_vision = 0;   // and so does the vision encoder
+                have_config = true;
+            } else if (policy == kAcceleratorPolicyGpu) {
+                config.gpu_layers = -1;  // offload all of them
+                have_config = true;
+            }
+            // AUTO and NPU are left alone for the same reasons as the LLM path.
         } catch (const std::exception& e) {
             RAC_LOG_WARNING(LOG_CAT, "config_json parse failed (%s); using defaults", e.what());
         }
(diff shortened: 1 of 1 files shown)


ayaangazali:
Adding the end-to-end half of the evidence, since the description only showed that the config struct has the fields, not that the keys actually reach this adapter.

They do. `model_lifecycle.cpp:275-279` builds the VLM create payload by merging the mmproj object with the same load options the LLM path gets:

```cpp
const std::string config_json =
    merge_json_objects(vlm_config_json(mmproj_path), options_json);
rc = vt->vlm_ops->create(resolved_path.c_str(),
                         config_json.empty() ? nullptr : config_json.c_str(), &impl);
```

`options_json` is the same string described at `model_lifecycle.cpp:184` as forwarding "the v4 load knobs (context_length/threads/use_gpu/accelerator_policy) as advisory config_json". So `context_length` and `accelerator_policy` were already arriving at `llamacpp_vlm_create_impl` on every VLM load; the adapter read `mmproj_path` out of the object and dropped the rest on the floor.

Worth noting the ABI contract cited in that same comment: plugins that do not understand a key MUST ignore it, so nothing here was violating the contract. It is just that llama.cpp's VLM backend does understand these two, and its LLM sibling started honouring them yesterday.

ayaangazali:
Still valid on today's `main` (`00e879fa8`), a week and four releases later (0.20.28 through 0.20.31). Re-verified rather than assumed:

`llamacpp_vlm_create_impl` still passes a null config, so the VLM path still discards the load options the LLM path reads:

```
engines/llamacpp/rac_llamacpp_vlm_ops.cpp:148
    rac_result_t rc = rac_vlm_llamacpp_create(model_id, mmproj_path, nullptr, &backend_handle);
```

and the only `context_length` references in that file are the unrelated `out_info->context_length` model-info fields at lines 82-94, not load-option parsing. So nothing has started honouring `context_length` or `accelerator_policy` on this path in the meantime.

The branch merges clean into current `main` with no conflicts, and CI is green (25 passing, 0 failing).

No changes pushed, since a rebase would be a no-op force-push that only resets the review state. Happy to rebase or split it differently if that helps it land.

ayaangazali:

```
$ sed -n '148p' engines/llamacpp/rac_llamacpp_vlm_ops.cpp
    rac_result_t rc = rac_vlm_llamacpp_create(model_id, mmproj_path, nullptr, &backend_handle);
```

The third argument is the load config, and it is still hard-coded `nullptr`, so the VLM create path discards whatever the caller configured. The LLM create path in the same engine passes its config through, which is the asymmetry this PR closes.

Checked the obvious counter-argument again, that the config might be applied elsewhere for VLM: the only `context_length` references in this file (lines 82-94) are `out_info->context_length` on the model-info struct, which is reporting, not load configuration.

Merges clean, CI green.

ayaangazali:
The rule you are citing is the right rule, but applying it here would mean this PR stops doing the one thing it is for.

Those two constants are not an invention of this PR. They already exist, byte for byte including the comments, in the sibling file in the same engine, on `main`:

```
engines/llamacpp/rac_backend_llamacpp_register.cpp:59  constexpr int kAcceleratorPolicyCpu = 2;  // ACCELERATOR_POLICY_CPU
engines/llamacpp/rac_backend_llamacpp_register.cpp:60  constexpr int kAcceleratorPolicyGpu = 3;  // ACCELERATOR_POLICY_GPU
engines/llamacpp/rac_backend_llamacpp_register.cpp:83  if (parsed.contains("accelerator_policy") && parsed["accelerator_policy"].is_number_integer()) {
engines/llamacpp/rac_backend_llamacpp_register.cpp:86  policy = parsed["use_gpu"].get<bool>() ? kAcceleratorPolicyGpu : kAcceleratorPolicyCpu;
```

That is the LLM create path. This PR's entire purpose is that the VLM create path was ignoring load options the LLM path honours, so it reads the same JSON keys, in the same order of precedence, against the same values. Diverging from the sibling here would defeat the parity the PR exists to establish.

On the underlying point, you are correct that nothing links these to the schema. I checked for the canonical alternative and it does not exist yet:

- `engines/llamacpp/CMakeLists.txt` has no protobuf linkage, so the engine cannot use the generated C++ enum.
- `grep -rn ACCELERATOR_POLICY core/include/` returns nothing, so there is no protobuf-free C header exposing these values either. They live only in `idl/model_types.proto:524-528`.

So the correct fix is the one you describe as the fallback: generate a C enum for `AcceleratorPolicy` into the public headers and have both engine files use it. That is a codegen change plus `schema_lock.sh --update` and an `idl/VERSION` bump, touching every SDK's generated output. Folding it into a 40-line VLM parity fix would bury it, and it would leave the two files inconsistent until it landed.

No change made here. If you would rather see the generated enum first and this PR rebased on top, I am happy to do it in that order, but I did not want to expand this diff into the IDL without you asking for it.

sanchitmonga22:
@coderabbitai review

sanchitmonga22:
@coderabbitai full review

sanchitmonga22:
> Replying to [this comment](https://github.com/RunanywhereAI/runanywhere-sdks/pull/755#issuecomment-5591940169)

That makes sense — thanks for explaining. You're right that this only extends a pattern already on `main` (`rac_backend_llamacpp_register.cpp:59-60` has the identical CPU/GPU copy, and `engines/llamacpp` links no protobuf), so there's no need to hold this PR on a codegen change that doesn't exist yet. I opened #876 to track a canonical, protobuf-free AcceleratorPolicy header separately. Treating this as addressed.

sanchitmonga22 (approved):
Thanks a lot for this, @ayaangazali! VLM loads through the llama.cpp engine used to ignore the requested context length and accelerator placement (CPU/GPU, or the older use_gpu flag) — this makes them take effect, the same way they already do for LLM loads.

Checked: CodeRabbit reviewed the latest commit · CI green · built and linted locally merged into `main` (C++ commons and engines (engines/llamacpp/**), 104/104 tests passing) · two independent code reviews (Opus 5 and Codex GPT-5.6 Sol).

Follow-up, not blocking: #876 (a generated, protobuf-free AcceleratorPolicy header so CPU/GPU/NPU/UNSPECIFIED stop being hand-copied across three files) — you're welcome to pick it up.

One small optional note for later: a caller-set `context_length` now reaches the VLM loader as-is (`rac_llamacpp_vlm_ops.cpp:151` → `rac_vlm_llamacpp.cpp:1096-1107`, which only caps the auto-sized value), while the LLM path fits the context to device memory and falls back to 2048 when that fails — worth guarding the same way if you're back in this file, not required here.

PR #758, opened 2026-08-22
docs: two stale references (removed CI job, renamed codegen script)

ayaangazali:
Two documentation references that point at things which do not exist. Bundled rather than sent separately.

## 1. `bindings/python/PACKAGING.md:18`

> CI (`.github/workflows/pr-build.yml`) already builds + repairs + validates + hermetic-tests the wheel on `python-windows`, **`python-linux`**, and `python-macos`.

#757 removed `python-linux` earlier today, so a reader following this goes looking for a job that is not there. The row above it in the artifact table (`runanywhere-*-manylinux_2_35_x86_64.whl`) is also no longer produced by CI, so I noted that in the same sentence rather than leaving the table silently implying otherwise. I left the table row itself alone, since a manylinux wheel is still a thing the project can produce, just not from `pr-build.yml`.

## 2. `scripts/validation/verify_default_pool.sh:26`

The exit-code legend says:

> `2  A generated file is missing (run idl/codegen/generate_defaults_pool.sh).`

The generator is `generate_defaults_pool.py`. There is no `.sh`. This is self-evidently the outlier rather than a judgement call, because the same script's runtime message on that exact path already prints the right name:

```
$ bash scripts/validation/verify_default_pool.sh --quiet
...
Run: idl/codegen/generate_defaults_pool.py
```

## Verification

`bash -n` on the shell script passes, and I ran it to confirm the behaviour is untouched (it is a comment change) and to capture the runtime line quoted above. `ls idl/codegen/generate_defaults_pool.*` returns only the `.py`.

Docs only, no behaviour change.

## Summary by CodeRabbit

- **Documentation**
  - Clarified that CI builds Windows and macOS wheels only.
  - Documented the manual process for building and repairing manylinux wheels with `auditwheel`.

- **Bug Fixes**
  - Updated validation guidance to reference the current command for generating missing default files.

### DIFF
--- a/bindings/python/PACKAGING.md
+++ b/bindings/python/PACKAGING.md
@@ -23,8 +23,10 @@ wheel on `python-windows` and `python-macos`. The release just collects those ar
 1. **Version** — bump `core/VERSION`; `scripts/release/sync-versions.sh` propagates
    it into `pyproject.toml`. Confirm `pip show`/`__version__` match.
 2. **Build sdist** — `python -m build --sdist -o dist`.
-3. **Build + repair each wheel** on its platform (Win/Linux/macOS) via the CI recipe (build →
-   delvewheel/auditwheel/delocate).
+3. **Build + repair each wheel** on its platform (Win/Linux/macOS) (build →
+   delvewheel/auditwheel/delocate). Windows and macOS can follow the `python-windows` /
+   `python-macos` jobs in `pr-build.yml`; since #757 removed `python-linux` there is no CI recipe
+   left to copy for the manylinux wheel, so it has to be built and `auditwheel`-repaired by hand.
 4. **Validate the release boundary** — `python scripts/validate_public_packages.py --dist dist
    --expected-version <VERSION>` (asserts: `_core` + sidecar libs present in the right dir per
    platform, no host-path leaks in text/metadata members, the sdist has sources only, version matches).
--- a/bindings/python/PACKAGING.md
+++ b/bindings/python/PACKAGING.md
@@ -15,8 +15,8 @@ wheel. The sdist is source-only.
 | `runanywhere-*.tar.gz` (sdist) | scikit-build-core | — | source only (no binaries) |
 
 CI (`.github/workflows/pr-build.yml`) already builds + repairs + validates + hermetic-tests the
-wheel on `python-windows`, `python-linux`, and `python-macos`. The release just collects those
-artifacts.
+wheel on `python-windows` and `python-macos`. The release just collects those artifacts.
+`python-linux` was removed in #757, so the manylinux wheel above is not built by CI today.
 
 ## Dry-run checklist (before `twine upload`)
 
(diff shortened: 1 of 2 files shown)


ayaangazali:
Fair, and I had stopped one line short. Step 3 of the dry-run checklist still said to build and repair each wheel "on its platform (Win/Linux/macOS) via the CI recipe", which is exactly the pointer that #757 invalidated for Linux.

Reworded so the checklist says where each wheel actually comes from: Windows and macOS can follow the `python-windows` / `python-macos` jobs, and there is no Linux job left to copy, so the manylinux wheel has to be built and `auditwheel`-repaired by hand.

I deliberately left the `manylinux_2_35` row in the "What gets published" table alone. Whether that wheel keeps shipping is a maintainer call rather than a doc fix, and the note now states the build situation plainly enough that the decision is visible either way.

Rebased onto `c4a8fb283`.

ayaangazali:
Both references are still stale on today's `main` (`00e879fa8`). Re-verified rather than assumed:

**1. `python-linux`.** `pr-build.yml` has no such job any more; the only Python wheel jobs are `python-windows` (line 230) and `python-macos` (line 300). `bindings/python/PACKAGING.md:18` still tells the reader all three build in CI.

**2. `generate_defaults_pool`.** Only `idl/codegen/generate_defaults_pool.py` exists; there is no `.sh`. Worth adding that `scripts/validation/verify_default_pool.sh` contradicts itself on this, which I did not spell out in the description:

```
26:#   2  A generated file is missing (run idl/codegen/generate_defaults_pool.sh).
74:    print("\nRun: idl/codegen/generate_defaults_pool.py", file=sys.stderr)
```

The runtime error message on line 74 already points at the real file; only the exit-code comment on line 26 was missed when it was renamed. So this is a one-word fix that makes the file agree with itself.

Branch merges clean into current `main`, CI green. No changes pushed, since a rebase would be a no-op force-push that only resets the review state.

ayaangazali:
Both claims re-verified on `45e91276e` (0.20.36).

The CI one:

```
$ grep -n 'python-' .github/workflows/pr-build.yml
230:  python-windows:
300:  python-macos:
$ grep -n 'python-linux' bindings/python/PACKAGING.md
18: wheel on `python-windows`, `python-linux`, and `python-macos`.
```

There is no `python-linux` job, and there is no other workflow defining one.

The codegen one is a file contradicting itself four dozen lines apart:

```
verify_default_pool.sh:26: run idl/codegen/generate_defaults_pool.sh
verify_default_pool.sh:74: print("\nRun: idl/codegen/generate_defaults_pool.py", ...)
```

Only the `.py` exists. The rename reached the runtime message and missed the exit-code comment above it, so the file tells you to run a script that is not there and then, at the point of failure, tells you the right one.

Both are one-line text fixes. Merges clean.

ayaangazali:
Closing in favour of #849, which carries this finding along with the three other stale-reference fixes I had drip-fed as separate PRs.

Four docs PRs is four review threads for what should be one review. That was my mistake in splitting them; #849 has all four with the same evidence, re-verified on `45e91276e`. Nothing is dropped.

PR #759, opened 2026-08-22
fix(electron): coerce producer throw/reject to SDKException in bridgeStream

ayaangazali:
## What is wrong

`bindings/electron/src/api/iter.ts` opens with the contract it exists to enforce:

> This file is a thin Electron binding that coerces stream failures through `asSDKException` so renderer and main consumers **always** see the house error type.

It is not always. It holds when a producer calls `sink.fail()`, and breaks when a producer throws or rejects.

The wrapper only wraps the sink:

```ts
return sharedBridgeStream<T>((sink) => producer(withTypedFail(sink)), onCancel);
```

`pushStream` reports a producer failure through the sink **it** owns, which is the unwrapped one:

```ts
    try {
      const maybe = producer(sink);
      if (maybe != null && typeof (maybe as Promise<void>).then === 'function') {
        (maybe as Promise<void>).then(
          () => sink.end(),
          (e: unknown) => sink.fail(e),   // raw sink
        );
      }
    } catch (e) {
      sink.fail(e);                       // raw sink
    }
```

So an error that reaches the consumer by escaping the producer never passes through `asSDKException`.

## Evidence

Driving the real `pushStream` from `@runanywhere/proto-ts/streams/push` with `withTypedFail` and `bridgeStream` copied verbatim out of `iter.ts`:

```
A  producer calls sink.fail(raw)  -> SDKException  (contract held)
B  async producer rejects         -> Error  <-- RAW, contract broken
C  sync producer throws           -> Error  <-- RAW, contract broken
```

Most of the 20-odd `bridgeStream` call sites are `async (sink) => { ... }` producers, and `bridge.ts` already funnels native rejections through `asSDKException`, so a native failure is unaffected. What does leak is anything thrown outside a bridge call. The clearest case is a caller-supplied iterable: `stt.transcribeStream(input)` does `for await (const chunk of input)` inside the producer, so if the app's audio iterable rejects with a plain `Error`, the consumer of the returned stream gets that plain `Error` instead of an `SDKException`.

The sibling adapter in this same SDK already guarantees and tests this. `test/unit/stream.test.ts` on `toAsyncIterable`:

```ts
      // Coerced to SDKException so consumers always see the house error type.
      assert.ok(isSDKException(thrown));
```

`bridgeStream` states the same contract in prose and does not deliver it on two of its three failure paths.

## What this changes

Wraps the producer call rather than only the sink, so a throw and a rejection route through the same `withTypedFail` sink that an explicit `fail()` uses.

The one thing worth reviewing is that a synchronous producer has to stay synchronous. `pushStream` auto-`end()`s the stream when the callback's promise settles, so returning a promise unconditionally would truncate a sync producer that registers a callback and keeps pushing (`assets.ts:1462` is one). The callback therefore returns a promise only when the producer returned one.

Re-running the probes against the patched wrapper, with two cases added to pin the semantics that must not move:

```
A  producer calls sink.fail(raw)  -> SDKException  (contract held)
B  async producer rejects         -> SDKException  (contract held)
C  sync producer throws           -> SDKException  (contract held)
D  sync producer, pushes later    -> ended, values=["late"]
E  async producer, normal end     -> ended, values=["x"]
```

## Verification

`npm ci && npm run build` in `bindings/electron` on this branch and on unpatched `main`: 118 type errors both times, identical sets. The only difference is a line number moving on a pre-existing error in this file (`iter.ts(71)` to `iter.ts(93)`) because the diff inserts lines above it. Those 118 are all `Cannot find module '@runanywhere/proto-ts/...'` and their cascade, because I have not run the IDL codegen that `electron-unit` does before building. The diff adds none.

No test here, following the repo's standing instruction not to add unit tests unless asked. Happy to add the `bridgeStream` counterpart to `test/unit/stream.test.ts`'s coercion case if you would like it, since it fails on the unpatched code.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stream error handling for failures during synchronous or asynchronous data production.
  * Prevented streams from completing prematurely when processing synchronous producers.

### DIFF
--- a/bindings/electron/src/api/iter.ts
+++ b/bindings/electron/src/api/iter.ts
@@ -45,7 +45,29 @@ export function bridgeStream<T>(
   producer: (sink: StreamSink<T>) => void | Promise<void>,
   onCancel?: () => void | Promise<void>
 ): AsyncIterableIterator<T> {
-  return sharedBridgeStream<T>((sink) => producer(withTypedFail(sink)), onCancel);
+  return sharedBridgeStream<T>((sink) => {
+    // `pushStream` reports a producer that throws or rejects through the sink
+    // it owns, not the one the producer was handed, so those failures would
+    // skip `withTypedFail`. Route them ourselves.
+    //
+    // A synchronous producer must stay synchronous: returning a promise from
+    // this callback makes `pushStream` auto-`end()` the stream once it
+    // settles, which would truncate a producer that is still pushing.
+    const typed = withTypedFail(sink);
+    let pending: void | Promise<void>;
+    try {
+      pending = producer(typed);
+    } catch (error) {
+      typed.fail(error);
+      return;
+    }
+    if (pending != null && typeof (pending as Promise<void>).then === 'function') {
+      return (pending as Promise<void>).catch((error: unknown) => {
+        typed.fail(error);
+      });
+    }
+    return pending;
+  }, onCancel);
 }
 
 /** Wrap a ready array as a bridge-safe stream (used for synthetic events). */


ayaangazali:
Still reproduces on today's `main` (`00e879fa8`), and I can now show it rather than argue it.

`iter.ts:48` on `main` is still the unwrapped form:

```ts
return sharedBridgeStream<T>((sink) => producer(withTypedFail(sink)), onCancel);
```

I built the Electron package at `main` and drove `bridgeStream` with a producer that fails each way:

```
sync-throw:   threw Error (SDKException=false) msg="boom-sync"
async-reject: threw Error (SDKException=false) msg="boom-async"
```

Then rebuilt at this branch and ran the identical script:

```
sync-throw:   threw SDKException (SDKException=true) msg="boom-sync"
async-reject: threw SDKException (SDKException=true) msg="boom-async"
```

So on `main` a producer failure reaches the consumer as a raw `Error`, bypassing `withTypedFail` exactly as the description says, and this branch is what makes both paths arrive as `SDKException`. That matters for callers doing `catch (e) { if (e instanceof SDKException) ... }`, which is the pattern the rest of the Electron surface documents.

Worth noting the existing suite does not cover it: I ran `npm test` at `main` earlier today and it is 241 passing, 0 failing. The `toAsyncIterable` tests in `stream.ts` exercise a throwing `start()`, but nothing exercises a throwing producer through `bridgeStream` in `iter.ts`, which is why this went unnoticed.

Branch merges clean into current `main`, CI green. Nothing pushed, since a rebase here would be a no-op force-push. Happy to add a regression test for the two cases above if you would like one in the same PR.

ayaangazali:
Still reproduces on `45e91276e` (0.20.36), and I re-checked specifically whether it had been fixed upstream, because `iter.ts` on `main` does already mention `asSDKException`.

It has not. `main` wraps the sink it hands the producer:

```ts
return sharedBridgeStream<T>((sink) => producer(withTypedFail(sink)), onCancel);
```

but `pushStream` reports a producer that throws or rejects through the sink *it* owns, not the one the producer was handed, so those failures bypass `withTypedFail` entirely and reach the consumer as a bare `Error` rather than an `SDKException`. The partial handling at lines 34 and 72 covers the sink-side path, not the producer-side one.

The empirical before/after I ran when I opened this still stands: built at `main`, drove `bridgeStream` with a throwing and a rejecting producer, got `Error (SDKException=false)` both times; rebuilt at this branch, identical script, `SDKException=true` both times.

Merges clean, CI green.

sanchitmonga22:
@coderabbitai review

sanchitmonga22:
@coderabbitai full review

sanchitmonga22:
> Replying to [this comment](https://github.com/RunanywhereAI/runanywhere-sdks/pull/759#issuecomment-5459890930)

That makes sense — thanks for the clear repro. The fix in this PR addresses the underlying issue you found, so I'm merging it now; a regression test would be a nice follow-up whenever you have time, but it isn't required for this PR.

sanchitmonga22:
> Replying to [this comment](https://github.com/RunanywhereAI/runanywhere-sdks/pull/759#issuecomment-5555015383)

Confirmed, and merged now — thanks for checking back and for your patience. Feel free to open a follow-up PR with a regression test if you'd like to add one, but it's optional.

sanchitmonga22 (approved):
Thanks a lot for this, @ayaangazali! When an Electron `bridgeStream` producer throws synchronously or its returned promise rejects, the error now reaches the consumer as an `SDKException` instead of a raw `Error`, exactly as `iter.ts` already promises elsewhere — and you tracked down and demonstrated the root cause (the shared `pushStream` reports producer failures through its own raw sink, bypassing `withTypedFail`) really thoroughly.

Checked: CodeRabbit reviewed the latest commit · CI green · built and linted locally merged into `main` (electron) · two independent code reviews.

Two small optional notes for later, no action needed: `iter.ts:65` calls `.catch` on a value only checked for `.then` (an edge case for a producer returning a non-native thenable); and a regression test for this fix — which you offered — would be a nice addition whenever you have time, e.g. in `test/unit/stream.test.ts`.

Merging now — really appreciate the contribution, and thanks for your patience waiting on this one!

PR #765, opened 2026-08-23
fix(core): refuse RAC_BUILD_BACKENDS when configuring core/ standalone

ayaangazali:
#761 established the rule and put it in `core/AGENTS.md`:

> **Configure from the REPO ROOT, never `core/`.** `core/CMakeLists.txt` is a standalone `project(RunAnywhereCommons)` that never calls `add_subdirectory(engines)` — only the repo-root `CMakeLists.txt` does. Configuring `core/` creates *no engine targets at all*.

Nothing enforces it, and `core/` still accepts the flag that cannot work.

## What is wrong

`core/CMakeLists.txt:42` advertises:

```cmake
option(RAC_BUILD_BACKENDS "Build ML backends (LlamaCPP, ONNX, RAG)" OFF)
```

Setting it does nothing here. On current `main`:

```
$ cmake -S core -B /tmp/on  -DRAC_BUILD_BACKENDS=ON  ; echo $?
0
$ cmake -S core -B /tmp/off -DRAC_BUILD_BACKENDS=OFF ; echo $?
0
$ diff <(cmake --build /tmp/on --target help | sort) <(cmake --build /tmp/off --target help | sort)
$
```

709 targets either way, identical lists, no warning. The only `rac_backend_*` target present is `rac_backend_rag`, which is folded into commons regardless of this flag; there is no llamacpp, onnx, sherpa or cloud target and no `runanywhere_*` carrier. Grepping confirms why: outside the `option()` line and three comments, `RAC_BUILD_BACKENDS` is never read in this file.

That is the same silent no-op #761 was written about, one layer up: the request is accepted, nothing happens, exit 0.

## The docs walk you into it

Three places in this directory instruct the failing form, all with `core/` as the working directory:

- `core/AGENTS.md:25` — "Build with all backends enabled" → `cmake -B build -DRAC_BUILD_BACKENDS=ON`
- `core/AGENTS.md:279` — "Tests requiring backends" → same flag
- `core/README.md:89` — after an explicit `cd runanywhere-sdks/core`

The first of those sits 220 lines above the rule quoted at the top of this description, in the same file.

## What this changes

A `FATAL_ERROR` when `core/` is the top-level project and the flag is on, using the `CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR` idiom this file already uses at line 868 for the standalone-vs-monorepo split:

```
CMake Error at CMakeLists.txt:51 (message):
  RAC_BUILD_BACKENDS=ON cannot be honoured here: engines/ is added by the
  repo-root CMakeLists.txt, not by core/.  Configure the repo root instead
  (from core/: cmake -S .. -B build -DRAC_BUILD_BACKENDS=ON), or drop the
  flag for a commons-only build.
```

and the three instructions repointed at the root with `-S ..`. The backends-test snippet also gets its binary path corrected: with the root as source dir, tests land in `build/core/tests/`, not `build/tests/`.

I did not touch the `option()` line itself. It has to stay declared here for the root and `engines/` to consume, which the comment above it already explains.

## Verification

All three paths, on this branch:

| configure | result |
|---|---|
| `cmake -S core -B … -DRAC_BUILD_BACKENDS=ON` | fails with the message above |
| `cmake -S core -B …` (no flag) | configures, rc 0 — commons-only builds unaffected |
| `cmake -S . -B … -DRAC_BUILD_BACKENDS=ON` | configures, rc 0, guard does not fire |

`check_agents_claude_sync.sh` passes (20 symlinks intact).

Note on overlap: #747 also edits `core/CMakeLists.txt`, deleting the `RAC_INCLUDE_LOCAL_DEV_CONFIG` block immediately above the line I append after. A trial merge of this branch with #747 is clean, but they are adjacent, so whichever lands second may want a glance.

## Summary by CodeRabbit

* **New Features**
  * Full-page OCR support is now included in builds with protobuf enabled.

* **Documentation**
  * Updated ML-backend build instructions to use the repository root and a separate build directory.
  * Clarified that standalone core builds do not include backend engines.
  * Updated test executable paths for root-configured builds.

* **Bug Fixes**
  * Standalone configurations now provide a clear error when backend builds are enabled.
  * Backend-enabled and core-only builds no longer conflict through shared build configuration.


ayaangazali:
Good catch, and it was a real break in the snippets I added, not just a style point.

The block starts with a core-only configure that stamps `core/build` with `core/` as its source dir, and my replacement then pointed `-S ..` at that same directory. Running the two in the order the file lists them fails immediately:

```
$ cmake -B build -DCMAKE_BUILD_TYPE=Release      # core-only, as documented above
$ cmake -S .. -B build -DRAC_BUILD_BACKENDS=ON
CMake Error: The source ".../CMakeLists.txt" does not match the source
".../core/CMakeLists.txt" used to generate cache.  Re-run cmake with a
different source directory.
```

So I had swapped a silent wrong result for a loud wrong one. Better, but still wrong.

Now pointed at `../build/backends`, which the root `.gitignore` already covers via `build/`. Running the documented sequence in order:

```
step 1: cmake -B build ...                      rc=0
step 2: cmake -S .. -B ../build/backends ...    rc=0
both caches coexist:  build  ../build/backends
```

(I turned the three heavy engines off for that run to keep it quick. The clash was about the build directory, not which backends are enabled, and the error above reproduces with any flag set.)

The test snippet moved with it, so its binary path is now `../build/backends/core/tests/test_llm`, and both comments say why the directory differs rather than leaving it looking arbitrary.

The `FATAL_ERROR` in `core/CMakeLists.txt` is unchanged.

ayaangazali:
Still reproduces on today's `main` (`00e879fa8`). The instructions that silently produce a backend-less build are unchanged:

```
core/AGENTS.md:25   cmake -B build -DRAC_BUILD_BACKENDS=ON -DCMAKE_BUILD_TYPE=Release
core/AGENTS.md:279  cmake -B build -DRAC_BUILD_TESTS=ON -DRAC_BUILD_BACKENDS=ON -DRAC_BACKEND_LLAMACPP=ON
core/README.md:89   cmake -B build -DRAC_BUILD_BACKENDS=ON
```

All three configure `core/` as the source dir, and `engines/` is only added by the repo-root `CMakeLists.txt`, so the flag is accepted and does nothing. `core/CMakeLists.txt` declares the option at line 42 and still has no standalone guard; the only `standalone` matches in that file are the project description, the `RAC_WASM_RAG_STANDALONE` option and unrelated comments.

The failure mode is what makes it worth a guard rather than only a doc fix: the configure step succeeds, the build succeeds, and the first sign anything is wrong is a missing engine at runtime. `core/AGENTS.md:279` is the sharper case, since it tells you to run `./build/tests/test_llm` after a build that cannot contain an LLM backend.

Branch merges clean into current `main`, CI green (0 failing). Nothing pushed, since a rebase would be a no-op force-push.

ayaangazali:

`core/CMakeLists.txt:42` declares `RAC_BUILD_BACKENDS`, but `engines/` is added by the repo-root `CMakeLists.txt`, never by `core/`. So a standalone `cmake -S core -DRAC_BUILD_BACKENDS=ON` configures successfully, reports nothing unusual, and produces a build with zero engines. The option is accepted and silently ignored.

The reason to fail loudly rather than document it is in the history: that silent no-op is what shipped `RACommons-windows-x64` with no engines for several releases. A build that quietly does less than asked is indistinguishable from one that worked, right up until someone loads a model and gets no backend.

This PR makes the unsatisfiable combination a configure-time `FATAL_ERROR` naming the fix (configure the repo root instead), which is the same treatment `RAC_BUILD_SERVER=ON` already gets at line 1422 when `RAC_ENABLE_PROTOBUF` is off.

Merges clean, CI green.

ayaangazali:
Correct, fixed in the push above. The command my own error message printed does not work, and I reproduced it rather than taking it on faith:

```
$ cmake -S . -B build            # standalone configure from core/, succeeds
$ cmake -S .. -B build -DRAC_BUILD_BACKENDS=ON
CMake Error: The source "/.../CMakeLists.txt" does not match the source
"/.../core/CMakeLists.txt" used to generate cache.
Re-run cmake with a different source directory.
```

That is the exact situation the new error fires in: someone configured `core/` standalone, hit the FATAL_ERROR, and followed the instructions. `-B build` from `core/` is `core/build`, which already holds the standalone cache.

Your suggested path is also the one the repo already documents, in three places I had not checked when I wrote the message:

```
core/README.md:90    cmake -S .. -B ../build/backends -DRAC_BUILD_BACKENDS=ON
core/AGENTS.md:30    cmake -S .. -B ../build/backends -DRAC_BUILD_BACKENDS=ON -DCMAKE_BUILD_TYPE=Release
core/AGENTS.md:286   cmake -S .. -B ../build/backends -DRAC_BUILD_TESTS=ON -DRAC_BUILD_BACKENDS=ON ...
```

Confirmed the error still fires and now prints the working command.

ayaangazali:
`windows-x64` went red here. It is not this diff, and I re-ran it rather than only saying so.

The failing step is `vcpkg curl (static)`, a dependency install that runs before anything configures CMake. This PR changes one `message(FATAL_ERROR)` string in `core/CMakeLists.txt`. There is no path by which that affects vcpkg fetching curl, and if the new guard did fire it would surface as a CMake configure error, not a vcpkg one.

Same base, same afternoon, other branches of mine:

```
#848  windows-x64 SUCCESS
#813  windows-x64 SUCCESS
#801  windows-x64 SUCCESS
#765  windows-x64 FAILURE  (vcpkg curl (static))
```

#813 and #801 were rebased onto the same `main` commit within minutes of this one, so the toolchain is fine on that base. `main` itself also has a red run on 2026-09-07 between two green ones.

I force-pushed with a byte-identical tree to re-trigger the run:

```
tree before f35f577d7ead62aee9399ba912fc008dd29076b3
tree after  f35f577d7ead62aee9399ba912fc008dd29076b3
```

Same content, new run, so whatever it reports is about the runner and not the diff. If it comes back red again with the same step I will dig further rather than keep re-running it.

sanchitmonga22:
@coderabbitai review

sanchitmonga22:
@coderabbitai full review

sanchitmonga22:
Thanks for this, @ayaangazali! The fix looks solid -- it closes the standalone `core/` configure gap that used to let `RAC_BUILD_BACKENDS=ON` silently produce a backend-less build (the same silent no-op that previously shipped RACommons-windows-x64 with zero engines).

One thing worth sorting out while it waits: the core test runners configure `core/` itself with the flag on, so the new guard now stops them at configure time. `core/tests/scripts/run-tests.sh:53` sets `RAC_ROOT` to `core/`, and line 153-155 runs `cmake -S "${RAC_ROOT}" ... -DRAC_BUILD_BACKENDS=ON`; `run-tests-android.sh:216-223` and `run-tests-ios.sh:185-190` / `233-235` do the same. No CI workflow calls these scripts, which is why CI is still green. Having them configure the repo root (as your updated `core/README.md` shows), or dropping `-DRAC_BUILD_BACKENDS=ON` there, would keep them working.

Because this touches `core/AGENTS.md` -- the build/test command reference contributors and agents follow for `core/` -- a maintainer will give that part a final look before merging.

ayaangazali:
Good catch, and thank you for finding the callers rather than just the symptom. Fixed in the push above.

I took the first of your two options, having them configure the repo root, because dropping `-DRAC_BUILD_BACKENDS=ON` would quietly change what those runners test. Verified both directions rather than assuming:

```
# what the runners did
cmake -B ... -S core -DRAC_BUILD_TESTS=ON -DRAC_BUILD_BACKENDS=ON
CMake Error at CMakeLists.txt:51 (message):
  RAC_BUILD_BACKENDS=ON cannot be honoured here: engines/ is added by the ...

# what they do now
cmake -B ... -S <repo root> -DRAC_BUILD_TESTS=ON -DRAC_BUILD_BACKENDS=ON
-- Configuring done (129.5s)
-- Generating done (0.8s)
-- Build files have been written to: ...
```

All four call sites: `run-tests.sh:159`, `run-tests-android.sh:222`, `run-tests-ios.sh:191` and `:239`. Each script now resolves `REPO_ROOT` next to `RAC_ROOT` and passes that to `-S`; build directories stay under `core/` and `RAC_ROOT` keeps its other uses, so nothing else in the runners moves.

Worth saying plainly: this is not just the guard being inconvenient. Those configures were already producing a backend-less build, so the backend tests in those runners were running against no engines. The guard turned a silent no-op into a loud one, which is the same failure that shipped RACommons-windows-x64 with zero engines. As you note, no CI workflow calls these scripts, which is why CI stayed green either way.

All three scripts pass `bash -n`. I have not executed the Android or iOS runners here (no device or emulator on this machine), so those two are verified by the shared configure change and syntax check, not by a run.

PR #771, opened 2026-08-23
docs: two stale RAC_PLUGIN_API_VERSION numbers

ayaangazali:
Two comments name the wrong plugin ABI version. `core/include/rac/plugin/rac_plugin_entry.h:104` is the source of truth:

```c
#define RAC_PLUGIN_API_VERSION 9u
```

**`CMakeLists.txt:13`** — the banner describing what `engines/` publishes, in the file the same banner calls "the SINGLE entry point for native builds":

> Each publishes a `rac_engine_vtable_t` at `RAC_PLUGIN_API_VERSION=7u`

**`core/tests/test_rerank.cpp:7`** — the file comment contradicts the file's own assertion 170 lines later:

```
 *   - RAC_PLUGIN_API_VERSION == 8 and the primitive/vtable-slot wiring.
```
```c
    check(RAC_PLUGIN_API_VERSION == 9u, "RAC_PLUGIN_API_VERSION must be 9");
```

Both now say 9.

I checked the rest of the tree rather than only fixing what I tripped over. Every other statement of the current value is already right: `AGENTS.md:152`, `core/AGENTS.md:117` and `core/docs/ARCHITECTURE.md:152` all say 9u. The remaining `ABI v2` / `v4` / `v8` mentions are historical ("retired in ABI v4", "promoted from reserved_slot_2 in ABI v8") and are accurate as history, so I left them alone.

Comments only, no behaviour change.

## Summary by CodeRabbit

- **Documentation**
  - Updated documented plugin API version references from 7u to 11u.

- **Tests**
  - Updated reranking test expectations for plugin API version 11.
  - Added an Apple/NeuRT OCR end-to-end test using the lifecycle API.
  - The OCR test is skipped when no OCR bundle is configured.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,7 +10,7 @@
 #   - engines/                 — engine plugins (llamacpp, onnx, sherpa,
 #                                qhexrt, neurt). Each publishes
 #                                a `rac_engine_vtable_t` at
-#                                `RAC_PLUGIN_API_VERSION=7u` via
+#                                `RAC_PLUGIN_API_VERSION=9u` via
 #                                `RAC_STATIC_PLUGIN_REGISTER` or a dlopen'd
 #                                `rac_plugin_entry` symbol.
 #   - tools/plugin-loader-smoke/ — CLI smoke test for the plugin loader
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,7 +10,7 @@
 #   - engines/                 — engine plugins (llamacpp, onnx, sherpa,
 #                                qhexrt, neurt). Each publishes
 #                                a `rac_engine_vtable_t` at
-#                                `RAC_PLUGIN_API_VERSION=9u` via
+#                                `RAC_PLUGIN_API_VERSION=10u` via
 #                                `RAC_STATIC_PLUGIN_REGISTER` or a dlopen'd
 #                                `rac_plugin_entry` symbol.
 #   - tools/plugin-loader-smoke/ — CLI smoke test for the plugin loader
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,7 +10,7 @@
 #   - engines/                 — engine plugins (llamacpp, onnx, sherpa,
 #                                qhexrt, neurt). Each publishes
 #                                a `rac_engine_vtable_t` at
-#                                `RAC_PLUGIN_API_VERSION=10u` via
+#                                `RAC_PLUGIN_API_VERSION=11u` via
 #                                `RAC_STATIC_PLUGIN_REGISTER` or a dlopen'd
 #                                `rac_plugin_entry` symbol.
 #   - tools/plugin-loader-smoke/ — CLI smoke test for the plugin loader
--- a/core/tests/test_rerank.cpp
+++ b/core/tests/test_rerank.cpp
@@ -4,7 +4,7 @@
  *
  * Backend-neutral: registers a small in-test fake rerank engine (no model file,
  * no llama.cpp) and asserts:
- *   - RAC_PLUGIN_API_VERSION == 8 and the primitive/vtable-slot wiring.
+ *   - RAC_PLUGIN_API_VERSION == 9 and the primitive/vtable-slot wiring.
  *   - A registered rerank engine routes via rac_plugin_find(RAC_PRIMITIVE_RERANK)
  *     and rac_engine_vtable_slot() resolves rerank_ops.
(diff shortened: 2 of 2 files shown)


ayaangazali:
Still reproduces on today's `main` (`00e879fa8`). The canonical value is `rac_plugin_entry.h:104`:

```c
#define RAC_PLUGIN_API_VERSION 9u
```

and both references this PR touches are still behind it:

```
CMakeLists.txt:13              `RAC_PLUGIN_API_VERSION=7u` via
core/tests/test_rerank.cpp:7    - RAC_PLUGIN_API_VERSION == 8 and the primitive/vtable-slot wiring.
```

One detail I did not spell out in the description: `test_rerank.cpp` contradicts itself. The header comment says 8, while the assertion the file actually executes at line 177 says 9:

```c
check(RAC_PLUGIN_API_VERSION == 9u, "RAC_PLUGIN_API_VERSION must be 9");
```

So the test would fail loudly if the real value drifted, and only the prose above it was missed on the 8 to 9 bump. That is the same shape as the `CMakeLists.txt` line, which was left on 7 across two bumps.

I also re-swept the tree for any other reference pinned below 9 and these two are the only ones, so this PR closes the set rather than a sample of it.

ayaangazali:
Rebased onto `caed2daf4` and retargeted, because v0.20.32 moved the goalposts and made this PR's own numbers stale.

`RAC_PLUGIN_API_VERSION` is now `10u` (`core/include/rac/plugin/rac_plugin_entry.h:104`). The release updated the assertion but not the comment eight lines above it in the same file:

```
core/tests/test_rerank.cpp:7:   *   - RAC_PLUGIN_API_VERSION == 8 and the primitive/vtable-slot wiring.
core/tests/test_rerank.cpp:184:     check(RAC_PLUGIN_API_VERSION == 10u, "RAC_PLUGIN_API_VERSION must be 10");
CMakeLists.txt:13:              `RAC_PLUGIN_API_VERSION=7u` via
```

So the drift is now wider than when I opened this: the two comments say 8 and 7 against an actual 10. Both now point at 10.

That the same file can carry `== 10u` on one line and `== 8` in its header comment is the argument for the change, and it is why I have not tried to turn this into anything cleverer than a text fix: the assertion at :184 is the thing that actually fails when the ABI moves, and it is already correct.

Text-only, no behaviour. Merges clean, and `centralization` should now be green here since runanywhere-swift is cut at 0.20.32.

ayaangazali:
Retargeted again, to 11. Rebased onto `45e91276e` (0.20.36).

The numbers moved under this PR a second time while it sat. `RAC_PLUGIN_API_VERSION` is now `11u` (`rac_plugin_entry.h:104`), and the drift on `main` is three-way:

```
CMakeLists.txt:13                `RAC_PLUGIN_API_VERSION=7u` via
core/tests/test_rerank.cpp:7      - RAC_PLUGIN_API_VERSION == 8 and the primitive/vtable-slot wiring.
core/tests/test_rerank.cpp:184    check(RAC_PLUGIN_API_VERSION == 11u, "RAC_PLUGIN_API_VERSION must be 11");
```

Two comments and one assertion, in two files, disagreeing by four versions. The assertion is right because it fails the build when it is wrong; the comments are wrong because nothing checks them.

That is now the argument for this PR rather than the original "two numbers are stale". I have had to retarget it twice, 9 to 10 and now 10 to 11, purely because releases kept landing while it waited. Every ABI bump updates `:184` and leaves `:7` and `CMakeLists.txt:13` behind, so the gap widens on its own.

If a cheap guard is wanted instead of periodic correction, the same `static_assert`-style pin `:184` already uses could cover the doc strings, but that is a separate change and I have not made it here. This PR still just corrects the two numbers.

Text-only, no behaviour. Merges clean, CI green.

sanchitmonga22 (approved):
Thanks a lot for this, @ayaangazali! This fixes two stale `RAC_PLUGIN_API_VERSION` doc comments — the root `CMakeLists.txt` banner (said 7u) and the `core/tests/test_rerank.cpp` file header (said 8) — bringing both in line with the canonical 11u already defined in `rac_plugin_entry.h` and already checked by the test's own runtime assertion.

Checked: CodeRabbit reviewed the latest commit · CI green · built and linted locally merged into `main` (fast gates + C++ commons/engines build+test) · two independent code reviews.

Follow-ups, not blocking: #877 (doc references to the ABI version are stale elsewhere too) and #878 (a silent stale-version fallback in CppDesktopKit.cmake) — you're welcome to pick either up. The header's own version history, which stops at v9, is already being filled in by your #849.

PR #773, opened 2026-08-23
fix: MLX tool calls, workflow model loading, reasoning, and text-only VLM

Siddhesh2377:
Tool calling never worked on the MLX backend. Any model that emitted a tool call
failed the whole turn with a bare `-130` (`RAC_ERROR_GENERATION_FAILED`), while
plain generation on the same model was fine.

## What was happening

Commons parses tool calls out of the model's text, in the DEFAULT wire form from
`core/src/features/llm/tool_calling.cpp`:

```
<tool_call>{"tool":"name","arguments":{}}</tool_call>
```

MLX-LM's `ToolCallProcessor` buffers exactly that text and consumes it, handing
back a structured `Generation.toolCall` instead. `MLXRuntime` then dropped it:

```swift
case .toolCall:
    break
```

So commons' run loop received output with no tool call in it, `run_generate_once`
returned non-success, and the loop reported `LLM generation failed`.

The rejection path made it worse. Commons renders the tool schema into the
prompt and never declares tools to MLX-LM, so MLX-LM cannot match a call against
its own (empty) tool list and emits `.rejectedToolCall(reason: .undeclaredTool)`.
`MLXRuntime` turned that into a thrown `RejectedToolCallError`, meaning the model
producing a correct tool call was what killed the turn.

## The change

`case .toolCall` now serializes the call back to the wire form and emits it
through the token callback, so commons parses it the same way it does for every
other engine. No new representation is introduced on either side.

`case .rejectedToolCall` forwards the raw text when it arrived intact and lets
commons judge it. A truncated preview still throws, keeping the fail-closed
contract the original comment was protecting: half a call is worse than none.

Both generation loops are covered, LLM and VLM.

## Testing

Verified in the macOS example app against `mlx-lfm2.5-vl-3b-4bit` and
`MLX Qwen3.5 0.8B 4bit`. Asking "what is the current local time and date?" with
tools enabled now runs `get_current_time` and answers from the result. Before the
change the same prompt failed every time with `-130`.

Not covered here: commons still does not declare tools to MLX-LM, so the
rejection path is load-bearing rather than a fallback. Declaring them through
`applyChatTemplate(messages:tools:)` would be the cleaner fix and needs a way to
pass tool schemas to the backend.

## Summary by CodeRabbit

* **New Features**
  * Added agent workflow creation, validation, execution, cancellation, event streaming, scheduling, and run-history access.
  * Added workflow and reusable node-pack management, including import/export bundles.
  * Added tool-provider registration, automatic tool dispatch, and live progress reporting.
  * Added an integrated web-research tool with cited, validated results.
  * Added optional progress callbacks for tool-enabled generation and flexible secure storage.

* **Bug Fixes**
  * Improved rejected-call handling, generation cancellation, token flushing, interrupted download recovery, and secure file operations.

---

## Also on this branch

Six more fixes share the branch rather than getting their own PRs.

**Workflow nodes could not load a model at all.** `ensure_model_loaded` passed a
null registry handle to `rac_model_lifecycle_load_proto`, which rejects that
before it looks at anything else, so every LLM Generate node failed with
"could not load model 'x'" whatever model was set. The message came from the
node, not from the ABI, so it read as a missing model. The rerank node had the
same bug against `rac_model_lifecycle_resolve_paths_proto`.

**Required tool-call arguments were never checked.** The validator covered pack
nodes only, and it counted a key holding whitespace as configured. Tool nodes now
get the same check and a blank value counts as unset.

**`web_research` cut answers off mid-sentence** at a 640-token compose budget.
Now 1536.

**The thinking toggle did nothing on MLX.** The runtime passed `enable_thinking`
only when suppressing, and Qwen's template reads an undefined flag as off,
emitting a pre-closed `<think></think>` pair. It is now stated in both
directions. Two follow-on faults are fixed with it: the registry fallback in
`thinking_tags_from_request_or_model` overwrote the MLX prefill signal with
`false`, and the terminal result recomputed the split from raw text, which cannot
see an opening tag that lived in the prompt. Together these made a reasoning turn
arrive as one 7,800-character reply with the chain of thought in the answer.

**A vision model could not hold a text conversation.** Commons required exactly
one image, and both the MLX engine adapter and the MLX runtime rejected a null
one, so a VLM loaded for chat failed every text turn with "no lifecycle LLM model
loaded".

**Two public inputs could never work.** `stt.transcribe(.file(...))` sent a path
and commons refuses one, reporting a missing platform adapter; `ImageInput.bytes`
sent container bytes, which commons refuses outright to avoid feeding JPEG data
to a backend expecting raw pixels. Both now decode on the Swift side, which is
the layer that owns platform I/O.

## Testing

`test_agent_workflow` passes at 38/38, including three new cases for the tool
argument validation. The rest was verified end to end against real models on
macOS through the iOS app's test suite, which loads models off disk and runs
inference rather than mocking the SDK.

## Review round

All 20 review comments were worked through; the per-finding writeup is in a
comment below. Two more fixes came out of verifying them.

**A failure now names its cause.** `NativeProtoABI.invoke` is the choke point
every modality goes through, and it reported the proto buffer's message while the
engine's actual reason sat unread in `rac_error_get_details`. An MLX speech model
that could not run said "Inference failed"; it now says
`Inference failed: MLX speech transcription failed: unsupportedAudioFormat: MLX
speech inference currently accepts 16-bit mono PCM audio.` The detail is
thread-local and unstamped, so a stale one can in principle be appended; it is
read immediately after a failed call, the same window Solutions and Workflows
already rely on.

**Audio inputs are decoded before they cross the ABI.** That MLX message was
telling the truth. `AudioInput.file` was lowered as a container, and commons
hands `audio_data` to the engine *without reading `encoding`* — so a WAV arrived
with its 44-byte header as the first samples. Sherpa tolerated it, which is why
its transcripts looked fine and hid the bug; MLX refused, which was correct. The
SDK now decodes to mono PCM16 through AVFoundation, where platform I/O belongs,
and states `audioFormat = .pcm`, the field MLX actually gates on. All three
installed speech models now return accurate transcripts, where one previously
failed outright.

Worth flagging for other modalities: commons ignoring `encoding` means any caller
that sends a container is feeding header bytes to an engine.

## Still open

- Reasoning is honoured by one of the four installed models that advertise it.
  The plumbing above is fixed; what remains is per-model. `mlx-qwen3.5-0.8b`
  reasons but never closes its `<think>` block, and the two llama.cpp models do
  not honour the toggle in either direction.
- `stt.transcribe` now reaches the model and fails inside it with
  "Inference failed". Untouched here.
- `llm.generate` quietly runs the whole tool-calling loop whenever anything is in
  the global tool registry, because `toolChoice` defaults to `.auto`. Worth
  deciding whether that is the intended meaning of auto.


Siddhesh2377:
Addressed in 0ee016cb9.

`onToken` returns the consumer's stop signal and the forwarding paths were discarding it, so a caller that stopped reading would have kept the model generating. Both loops now honor it, each following its own existing convention rather than a single imposed one:

- The loop that carries held tokens sets `shouldFlushHeldTokens = false`, cancels, and breaks `generationLoop`, matching its `.chunk` case. That also closes the flush-after-stop path the review called out.
- The second loop only cancels. Its `.chunk` case does the same: the `break` there leaves the switch, not the loop, and the `isCancelled` check at the top of the next iteration is what actually ends it. Adding a label just for these two cases would have made them behave differently from the chunk path beside them.

The truncated-preview branch still throws and is unchanged.

Built for macOS and iOS.

Siddhesh2377:
Worked through all 20 review comments. Each was checked against the code before
changing anything; 19 were real and are fixed, one was already handled and one
is answered rather than applied. Commits `d7dae36`..`6d884a8`.

**web_research and search**
`quoted_spans` no longer opens a span on an apostrophe, so "Apple's outlook
improved and the company's revenue" stops yielding "s outlook improved and the
company" as a term to check a source for. Sentence splitting skips a full stop
between two digits, so "revenue of 94.9 billion" stays one sentence and keeps its
citation. The bullet test compares the three bytes `E2 80 A2` instead of a `char`
against `0x2022`, which never matched. A line starting with a space no longer
indexes `line[SIZE_MAX]`. `search()` refuses a 2xx with a null body rather than
constructing `std::string(nullptr, n)`, matching the guard `fetch_page_text`
already had.

**Tool provider ABI**
`rac_tool_provider_t` gained `abi_version`, checked at registration against
`RAC_TOOL_PROVIDER_ABI_VERSION`, so a provider built against another layout is
rejected instead of dispatched into. The progress sink now quiesces on replace:
an in-flight emit holds a count, and `register` waits for it before returning, so
a caller freeing `user_data` cannot pull it out from under a running callback. A
thread-local guard keeps a callback that re-registers from waiting on itself.

**ToolProgress identity**
`ToolProgress` gained `execution_id` (field 9), distinct per `execute()` call.
`sequence` restarts at 0 per execution, so with `parallel_tool_calls` two
providers under one run-loop handle emitted colliding keys; Swift's
`Identifiable.id` is now `"\(executionID).\(sequence)"`.

**Agent runner**
`execute_node` catches `nlohmann::json` and `std::exception` and reports through
the return code, since several executors parse user-supplied strings and every
caller reports failure by code. HTTP nodes get a bounded 30s default rather than
whatever a transport makes of `timeout_ms == 0`. `parse_request` accepts a
zero-length payload, which is how an all-default message encodes. A run destroyed
before `start()` is now persisted, as the header promises. The loop's
current-item context is cleared on the failure path, so a later node's
`{{ item.* }}` no longer resolves against a stale item. `store_save_document`
refreshes `updated_at_ms`, which the header documents and the code did not do.
Bundle import skips a document whose `schema_version` is newer than this build's,
instead of stamping it down and losing fields.

**Tool ownership**
The offer loop said a host tool of the same name wins, but dispatch asked only
whether any provider was registered under that name, so commons ran instead. Host
names are now tracked and consulted at dispatch too.

**Swift**
The MLX repetition guard's held tokens are flushed before a tool event is
forwarded, so commons receives text in the order the model produced it. A failed
`chmod` on the secure-store fallback file now removes the file and throws rather
than leaving a secret at the default mode silently.

**Not applied, with reasons**

*Honor callback cancellation for tool-call output* — already the case. Both
`.toolCall` branches check `onToken`'s return and cancel; the comments there say
so. The review predates the current diff.

*Use generated protobuf types for tool-progress contracts* — declining, because
`rac_tool_progress_status_t` is stated in C precisely so a provider never links
protobuf, which is the point of the `rac_*` surface. The real risk named is
drift, so the two enums are now pinned with `static_assert` at the one place they
meet, and a change to either fails the build.

Verified: `test_agent_workflow` 38/38, `swift build` clean, and the iOS app's
end-to-end suite green on the chat, VLM and TTS paths against real models.

PR #775, opened 2026-08-23
fix(stt): report audio duration at the configured sample rate

ayaangazali:
## What is wrong

`stt_module.cpp` computes `input_audio_duration_ms` for `VoiceLifecycleEvent` in two places, two different ways, and the batch path hard-codes 16 kHz:

```cpp
// batch, line 947
double audio_length_ms = (audio_size / 2.0 / 16000.0) * 1000.0;

// streaming, line 1116
double audio_length_ms = (audio_size * 1000.0) / (component->config.sample_rate * 2);
```

The batch one is wrong for any component not running at 16 kHz, and the rate it should be using is already in scope: `sample_rate` is assigned from `component->config.sample_rate` at line 916, and the same event sets it thirteen lines later via `voice.set_sample_rate(sample_rate)`. So the emitted message carries a duration and a sample rate that contradict each other.

For one second of mono 16-bit audio:

| rate | bytes | true ms | batch (947) | streaming (1116) |
|---|---|---|---|---|
| 8000 | 16000 | 1000 | **500** | 1000 |
| 16000 | 32000 | 1000 | 1000 | 1000 |
| 22050 | 44100 | 1000 | **1378** | 1000 |
| 44100 | 88200 | 1000 | **2756** | 1000 |
| 48000 | 96000 | 1000 | **3000** | 1000 |

The streaming one gets the rate right but divides by it unguarded, so a zero `sample_rate` is a division by zero and `static_cast<int64_t>` of the resulting infinity is undefined behaviour.

## The file already has the answer

`estimate_audio_length_ms(audio_size, sample_rate)` is defined at line 392 and does both things correctly: it guards the rate and uses the named constants rather than a magic `2` and a magic `16000`.

```cpp
int64_t estimate_audio_length_ms(size_t audio_size, int32_t sample_rate) {
    const int32_t rate = sample_rate > 0 ? sample_rate : RAC_STT_DEFAULT_SAMPLE_RATE;
    return static_cast<int64_t>(
        (static_cast<double>(audio_size) / static_cast<double>(RAC_STT_BYTES_PER_SAMPLE) /
         static_cast<double>(rate)) * 1000.0);
}
```

Four other call sites already use it (lines 432 and 1481, plus a three-argument overload at 1706 used at 1841 and 2052). These two were the ones that open-coded it.

## What this changes

Both sites now call the helper. Nine insertions, four deletions, no new code.

I left the four `static_cast<int64_t>(audio_length_ms)` at the consumers alone: the variable is now already `int64_t`, so those casts are no-ops, and removing them would grow the diff without changing behaviour.

## Scope

I swept the rest of `core/src` for the same shape rather than only fixing what I tripped over. Every other hard-coded `16000` is a legitimate fallback default (`param_int_or(spec, "sample_rate_hz", 16000)`, `options.sample_rate > 0 ? options.sample_rate : 16000`) rather than a computation ignoring a rate it already has. `rac_vad_stream.cpp`'s equivalent position math is fine: `audio_samples` is `uint64_t` so the `* 1000U` widens, and it already guards `sample_rate <= 0`.

## Verification

Built commons with `-DRAC_BUILD_TESTS=ON` (clean, zero compile errors) and ran the suite:

```
100% tests passed out of 100
```

The table above is both formulas evaluated directly, not an estimate.

Reachability, since a hard-coded default is only a bug if the value can differ: `sample_rate` is a caller-settable proto field (`idl/stt_options.proto:29`), reaching the component through `config->sample_rate` and `request.audio().sample_rate()`.

Note on overlap: #747 also edits this file, at `rac_stt_component_load_model` around line 826. It does not touch `audio_length_ms`, `sample_rate` or `16000`, and a trial merge of this branch with it is clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved audio-duration reporting for batch and streaming transcription.
  * Duration estimates now reflect configured sample rates and safely use a default when needed.
  * Telemetry duration calculations are now limited to supported protobuf-enabled configurations.

### DIFF
--- a/core/src/features/stt/stt_module.cpp
+++ b/core/src/features/stt/stt_module.cpp
@@ -943,11 +943,16 @@ extern "C" rac_result_t rac_stt_component_transcribe(rac_handle_t handle, const 
     }
     // Lock released — safe to do long-running transcription
 
+#if defined(RAC_HAVE_PROTOBUF)
     // Estimate audio length at the component's configured rate. The 16 kHz this
     // used to hard-code is only a default: the very next event field below is
     // set from `sample_rate`, so a component at any other rate reported a
     // duration that disagreed with the rate in the same message.
+    //
+    // Guarded because every consumer is: the lifecycle events below are the only
+    // readers, and estimate_audio_length_ms itself lives inside this same guard.
     const int64_t audio_length_ms = estimate_audio_length_ms(audio_size, sample_rate);
+#endif
 
     RAC_LOG_INFO("STT.Component", "Transcribing audio");
 
@@ -1115,10 +1120,13 @@ rac_stt_component_transcribe_stream(rac_handle_t handle, const void* audio_data,
         RAC_LOG_DEBUG("STT.Component", "STT streaming transcription using model_id: %s", model_id);
     }
 
+#if defined(RAC_HAVE_PROTOBUF)
     // Calculate audio length in ms. Via the shared helper so a zero sample rate
-    // falls back to the default instead of dividing by zero.
+    // falls back to the default instead of dividing by zero. Guarded for the
+    // same reason as the batch path above.
     const int64_t audio_length_ms =
         estimate_audio_length_ms(audio_size, component->config.sample_rate);
+#endif
 
     // Generate transcription ID for tracking
     std::string transcription_id = generate_unique_id();
--- a/core/src/features/stt/stt_module.cpp
+++ b/core/src/features/stt/stt_module.cpp
@@ -943,8 +943,11 @@ extern "C" rac_result_t rac_stt_component_transcribe(rac_handle_t handle, const 
     }
     // Lock released — safe to do long-running transcription
 
-    // Estimate audio length (assuming 16kHz mono 16-bit audio)
-    double audio_length_ms = (audio_size / 2.0 / 16000.0) * 1000.0;
+    // Estimate audio length at the component's configured rate. The 16 kHz this
+    // used to hard-code is only a default: the very next event field below is
+    // set from `sample_rate`, so a component at any other rate reported a
+    // duration that disagreed with the rate in the same message.
+    const int64_t audio_length_ms = estimate_audio_length_ms(audio_size, sample_rate);
 
     RAC_LOG_INFO("STT.Component", "Transcribing audio");
 
@@ -1112,8 +1115,10 @@ rac_stt_component_transcribe_stream(rac_handle_t handle, const void* audio_data,
         RAC_LOG_DEBUG("STT.Component", "STT streaming transcription using model_id: %s", model_id);
     }
 
-    // Calculate audio length in ms (assume 16kHz, 16-bit mono)
-    double audio_length_ms = (audio_size * 1000.0) / (component->config.sample_rate * 2);
+    // Calculate audio length in ms. Via the shared helper so a zero sample rate
+    // falls back to the default instead of dividing by zero.
+    const int64_t audio_length_ms =
+        estimate_audio_length_ms(audio_size, component->config.sample_rate);
 
     // Generate transcription ID for tracking
     std::string transcription_id = generate_unique_id();


ayaangazali:
`wasm` was mine, not environmental. Fixed in `b17e11a9`.

```
stt_module.cpp:950:37: error: use of undeclared identifier 'estimate_audio_length_ms'
stt_module.cpp:1121:9: error: use of undeclared identifier 'estimate_audio_length_ms'
```

`estimate_audio_length_ms` is inside `#if defined(RAC_HAVE_PROTOBUF)`, and the two computations I routed through it are unconditional. The wasm job configures without protobuf (its `em++` line carries `-DRAC_BUILDING_COMMONS=1 -DRAC_PLUGIN_MODE_STATIC=1 -DRAC_VERSION_STRING=...` and no `RAC_HAVE_PROTOBUF`), so the calls had no declaration in scope. My local build had protobuf on, which is why I did not see it.

Walking the preprocessor nesting for this file makes the right fix obvious:

| line | guard |
|---|---|
| 392 `estimate_audio_length_ms` | `#if defined(RAC_HAVE_PROTOBUF)` |
| 432 existing caller | `#if defined(RAC_HAVE_PROTOBUF)` |
| 950, 1121 my computations | **unconditional** |
| 963, 1034, 1135, 1193 consumers | `#if defined(RAC_HAVE_PROTOBUF)` |

Every reader of `audio_length_ms` is already inside the guard, so the value is protobuf-only data that was being computed outside it. I put the computation inside the guard rather than widening the helper's scope, which also drops what would otherwise be an unused-variable warning in non-protobuf builds (the same shape as the pre-existing `unused variable 'load_duration_ms'` at line 824 that shows up in this job's log).

Reproduced locally rather than guessing, by compiling the translation unit with `RAC_HAVE_PROTOBUF` undefined:

```
── PRE-FIX:   exit=1
core/src/features/stt/stt_module.cpp:950:37: error: use of undeclared identifier 'estimate_audio_length_ms'
core/src/features/stt/stt_module.cpp:1121:9: error: use of undeclared identifier 'estimate_audio_length_ms'
── POST-FIX:  exit=0, errors=0
```

Same two lines the wasm job reported. And the protobuf-on path is unaffected: full rebuild clean, `100% tests passed out of 100`.

ayaangazali:
Still reproduces on today's `main` (`00e879fa8`). Both open-coded sites are unchanged:

```
core/src/features/stt/stt_module.cpp:947
    double audio_length_ms = (audio_size / 2.0 / 16000.0) * 1000.0;
core/src/features/stt/stt_module.cpp:1116
    double audio_length_ms = (audio_size * 1000.0) / (component->config.sample_rate * 2);
```

The clearest way to see the first one is that the inconsistency lives inside a single message. A few lines below 947, the same `VoiceLifecycleEvent` is filled with both values:

```cpp
voice.set_input_audio_duration_ms(static_cast<int64_t>(audio_length_ms));  // computed as if 16 kHz
...
voice.set_sample_rate(sample_rate);                                        // the component's real rate
```

So a component running at 48 kHz emits one event that reports `sample_rate = 48000` alongside a duration derived from 16 kHz, a 3x error that a consumer cannot detect because both fields look authoritative.

The second site has a different failure: it divides by `component->config.sample_rate` with no guard. The shared helper already at line 392 in the same file has one:

```cpp
const int32_t rate = sample_rate > 0 ? sample_rate : RAC_STT_DEFAULT_SAMPLE_RATE;
```

so routing both through it is what this PR does, rather than adding a new convention.

ayaangazali:
Narrowing this, following up on my note above.

Two of the three files were fixed on `main` independently since I opened this, so their hunks had become no-ops that GitHub still displayed because the diff renders against the merge base:

- `core/cmake/FetchONNXRuntime.cmake` — `main` already carries `core/scripts/...` at lines 85, 161, 224 and 255
- `engines/sherpa/CMakeLists.txt:259` — `main` already carries `"bash core/scripts/linux/download-sherpa-onnx.sh"`, the exact string this proposed

I rebased onto `00e879fa8`, which dropped both cleanly, so the PR is now one file and one concern: the STT duration computation. My earlier comment verified only that half, which is the half that still reproduces.

Re-checked after the rebase:

```
cmake --preset macos-debug -DRAC_BUILD_BACKENDS=ON && ninja   ->  exit 0
ctest                                                        ->  128/128 passed
```

and, since the change sits inside `#if defined(RAC_HAVE_PROTOBUF)`, I re-ran the file's own compile command with `-DRAC_HAVE_PROTOBUF` stripped: exit 0, with only a pre-existing `unused variable 'load_duration_ms'` warning at line 824 that this diff does not touch. That guard is there because `estimate_audio_length_ms` lives inside the same guard, which is what broke the wasm build on my first attempt at this.

ayaangazali:
On the red `centralization`: it is not from this diff, and it is worth saying why it appeared only now.

One step of the job's seventeen failed, "Swift distribution repo (runanywhere-swift) is cut at this release". Everything else passed, including the two steps that actually read this area, "Release-train version coherence" and its gate tests. It is also the only failing job in the run.

The condition is documented in `scripts/release/sync-versions.sh:616`:

```
(scripts/validation/gates/check_swift_dist_repo_sync.sh fails every PR
 once v${NEW_VERSION} exists until that repo carries the ${NEW_VERSION} tag.)
```

`RunanywhereAI/runanywhere-swift` is tagged `0.20.30`; this repo has published `v0.20.31`.

This PR was green until an hour ago because its previous run predated that release. Rebasing it to drop the two superseded hunks triggered a fresh run, and any run from now on postdates v0.20.31, so it picks the gate up. The rebase did not cause the failure, it just moved the PR onto the far side of a release boundary.

No fix pushed, since the remedy is cutting the distribution repo rather than anything on this branch.

ayaangazali:
Re-verified on `45e91276e` (0.20.36), and the argument is now stronger than when I opened this: the correct helper already exists in this file, and these two call sites are the ones that do not use it.

`main` has a guarded helper, used on the transcribe-proto path:

```c
// stt_module.cpp:392
int64_t estimate_audio_length_ms(size_t audio_size, int32_t sample_rate) {
    const int32_t rate = sample_rate > 0 ? sample_rate : RAC_STT_DEFAULT_SAMPLE_RATE;
```

The two sites this PR touches still do the arithmetic inline:

```c
947:  double audio_length_ms = (audio_size / 2.0 / 16000.0) * 1000.0;
1116: double audio_length_ms = (audio_size * 1000.0) / (component->config.sample_rate * 2);
```

Line 947 hard-codes 16 kHz while the very next event field is populated from the component's actual `sample_rate`, so a component at any other rate emits a duration that contradicts the rate in the same message. Line 1116 uses the configured rate but divides by it with no zero guard, which is exactly what the helper's first line exists to prevent.

So this is no longer "please change this arithmetic", it is "three places compute the same quantity and one of them is already right".

Merges clean, no conflicts. The only red is `centralization`, which is the known dist-repo gate rather than anything here.

ayaangazali:
I checked this one carefully because a Critical protobuf-off compile break is exactly the kind of thing I have filed against this repo before (#824). It does not hold here: both changed calls are inside the same guard as the definition.

Scanning `stt_module.cpp` for the preprocessor state at each site, treating `#else` as inverting the enclosing condition:

```
line  392  2-arg definition   -> #if defined(RAC_HAVE_PROTOBUF)
line  954  call (this diff)   -> #if defined(RAC_HAVE_PROTOBUF)
line 1128  call (this diff)   -> #if defined(RAC_HAVE_PROTOBUF)
line 1494  call (pre-existing)-> NOT(#if !defined(RAC_HAVE_PROTOBUF))
```

954 is inside the `#if` at 946 that closes at 955; 1128 is inside the one at 1123 closing at 1129. Neither is unconditional. 1494 looks inverted at a glance but is in the `#else` arm, so it is also protobuf-on, which it has to be since it calls `event.mutable_final_output()`.

The `#else` handling is the part worth flagging for anyone re-checking this: a scan that tracks only `#if`/`#endif` reports 1494 as protobuf-off and makes this look like a real break.

There is also a comment already at line 953 stating this reasoning, which I put there when I wrote the change:

> readers, and estimate_audio_length_ms itself lives inside this same guard.

The non-protobuf path uses the 3-arg overload at 1719, which is outside every guard.

No change made. If you can point at a specific configuration where this fails to compile I will happily look again.

sanchitmonga22:
@coderabbitai full review

sanchitmonga22:
> Replying to [this comment](https://github.com/RunanywhereAI/runanywhere-sdks/pull/775#issuecomment-5591777491)

That makes sense -- thanks for walking through it a second time. We independently traced both call sites against the diff and confirmed they're correctly wrapped in `#if defined(RAC_HAVE_PROTOBUF)...#endif`, matching the green wasm build on this head. Treating this as addressed.

sanchitmonga22 (approved):
Thanks a lot for this, @ayaangazali! The STT component now reports transcription duration at the component's configured sample rate instead of a hard-coded 16 kHz, and the streaming path no longer risks a divide-by-zero/UB cast when the configured rate is zero.

Checked: CodeRabbit reviewed the latest commit · CI green · built and linted locally merged into `main` (core/src/features/stt, C++ commons) · two independent code reviews.

Follow-up, not blocking: #879 (making the duration/sample_rate telemetry track the effective per-call rate, not just the configured one) -- you're welcome to pick it up.

PR #784, opened 2026-08-24
fix(gates): match release asset names literally, not as regexes

ayaangazali:
## Description

`check_engine_prebuilt_pins.sh` fetches **one** release listing, from `NEURT_RELEASE_TAG`, and then greps it for both the NeuRT slices and the QHexRT ABIs:

```sh
listing="$(curl ... "https://api.github.com/repos/${repo}/releases/tags/${NEURT_RELEASE_TAG}" ...)"
...
for s in "${NEURT_SLICES[@]}"; do want+=("neurt-${s}-v${nv}.tar.gz"); done
# QHexRT has its OWN repo and tag pins; reusing NeuRT's would check the
# wrong release the moment the two diverge.
for a in "${QHEXRT_ABIS[@]}"; do want+=("qhexrt-${a}-v${qv}.tar.gz"); done
```

The comment states the hazard exactly. The filenames use `qv` (from `QHEXRT_RELEASE_TAG`), but the listing they are checked against is still NeuRT's tag. `QHEXRT_RELEASE_TAG` is an independent pin in `core/VERSIONS`, so as soon as the two differ the QHexRT assets live in a different release and the gate reports every ABI missing.

Latent today, because both pins happen to read `v0.20.26`. Not theoretical, though: #778 split QAIRT onto its own tag precisely because "the QAIRT SDK version and the engine version move independently", and QHexRT is pinned the same independent way.

This is also the one place in the function that does not already do the right thing. The QAIRT block a few lines below fetches its own listing for its own tag, and the QHexRT **checksum** loop already fetches with `$QHEXRT_RELEASE_TAG`. Only the presence check reuses NeuRT's listing.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring

## Testing
- [x] Lint passes locally — `bash -n` clean
- [ ] Added/updated tests for changes — no test harness exists for this gate (`scripts/validation/gates/` has one only for `check_release_version_coherence.sh`), and per `AGENTS.md` I do not add tests unless asked

The online section needs `NEURUN_TOKEN`, which I do not have, so I verified two ways.

Offline path unaffected:

```
$ bash scripts/validation/gates/check_engine_prebuilt_pins.sh --offline
  SKIPPED (offline or no NEURUN_TOKEN): cannot verify the release itself.
OK: engine prebuilt pins, consumers and release contents all agree.
```

And the divergent-pin case, replaying both versions of the loop against fixture listings where NeuRT is `v0.20.26` and QHexRT is pinned to its own `v0.21.0`, with both releases fully populated:

```
-- OLD (qhexrt grepped against the NeuRT listing):
   [FAIL] release is missing qhexrt-arm64-v8a-v0.21.0.tar.gz
   => gate FAILS although the QHexRT release is fine
-- NEW (qhexrt grepped against its own tag's listing):
   (passed, correctly)
```

### Platform-Specific Testing
Not applicable, this is a repo-level validation script.

## Labels
Closest is `core`, though this is `scripts/validation/gates/` rather than `core/` itself. Happy to relabel.

## Checklist
- [x] Code follows project style guidelines — mirrors the QAIRT block's structure in the same function
- [x] Self-review completed
- [ ] Documentation updated (if needed) — not needed, the existing comment already described the intended behaviour

## Screenshots
Not applicable.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved release asset validation for NeuRT, QHexRT, and QAIRT.
  - Asset names and checksum filenames containing dots are now matched exactly, preventing incorrect validation results.

### DIFF
--- a/scripts/validation/gates/check_engine_prebuilt_pins.sh
+++ b/scripts/validation/gates/check_engine_prebuilt_pins.sh
@@ -268,16 +268,19 @@ for a in rel.get('assets',[]): print(a['name'])" 2>/dev/null || true)"
         want=()
         for s in "${NEURT_SLICES[@]}"; do want+=("neurt-${s}-v${nv}.tar.gz"); done
         for w in "${want[@]}"; do
-            if ! grep -qx "$w" <<<"$listing"; then bad "release is missing $w"
-            elif ! grep -qx "${w}.sha256" <<<"$listing"; then bad "release is missing ${w}.sha256"
+            # -F: these asset names are full of dots, and without it grep reads
+            # each one as "any character", so a malformed name can satisfy the
+            # check for a well-formed one.
+            if ! grep -qxF "$w" <<<"$listing"; then bad "release is missing $w"
+            elif ! grep -qxF "${w}.sha256" <<<"$listing"; then bad "release is missing ${w}.sha256"
             fi
         done
         qwant=()
         for a in "${QHEXRT_ABIS[@]}"; do qwant+=("qhexrt-${a}-v${qv}.tar.gz"); done
         for w in "${qwant[@]}"; do
-            if ! grep -qx "$w" <<<"$qhexrt_listing"; then
+            if ! grep -qxF "$w" <<<"$qhexrt_listing"; then
                 bad "release $QHEXRT_RELEASE_TAG is missing $w"
-            elif ! grep -qx "${w}.sha256" <<<"$qhexrt_listing"; then
+            elif ! grep -qxF "${w}.sha256" <<<"$qhexrt_listing"; then
                 bad "release $QHEXRT_RELEASE_TAG is missing ${w}.sha256"
             fi
         done
@@ -300,8 +303,8 @@ for a in rel.get('assets',[]): print(a['name'])" 2>/dev/null || true)"
             done
             qairt_fail_before=$fail
             for w in "${qairt_want[@]}"; do
-                if ! grep -qx "$w" <<<"$qairt_listing"; then bad "QAIRT release is missing $w"
-                elif ! grep -qx "${w}.sha256" <<<"$qairt_listing"; then bad "QAIRT release is missing ${w}.sha256"
+                if ! grep -qxF "$w" <<<"$qairt_listing"; then bad "QAIRT release is missing $w"
+                elif ! grep -qxF "${w}.sha256" <<<"$qairt_listing"; then bad "QAIRT release is missing ${w}.sha256"
                 fi
             done
             [[ $fail -eq $qairt_fail_before ]] && ok "QAIRT release carries all $(( ${#qairt_want[@]} * 2 )) expected files"


ayaangazali:
Good catch, and it applies more broadly than the pair this PR added.

`grep -qx` reads the pattern as a basic regex, so every dot in an asset name matches any character:

```
$ printf 'qhexrt-arm64-v8a-v0X21X0Xtar4gz\n' | grep -qx "qhexrt-arm64-v8a-v0.21.0.tar.gz" && echo MATCHED
MATCHED

$ printf 'qhexrt-arm64-v8a-v0X21X0Xtar4gz\n' | grep -qxF "qhexrt-arm64-v8a-v0.21.0.tar.gz" && echo MATCHED
(no output — correctly rejected)
```

`-x` anchors the whole line but does nothing about the dots, so a malformed name of the right shape passes the presence check.

Fixed with `-F` at all six sites rather than just the two I introduced. The NeuRT and QAIRT pairs have the identical pattern against the identical kind of name, and leaving them regex while the QHexRT pair is literal would be the more confusing outcome. Comment added once above the first pair so the `-F` does not read as noise.

`bash -n` clean, `--offline` run still passes.

ayaangazali:
Half of this PR landed upstream, so I have cut it down to the half that did not.

**Superseded, removed from this PR.** `main` now reads QHexRT's own release listing, with the same shape this PR proposed, and its comment confirms the failure mode:

```
# QHexRT has its OWN tag pin, so it needs its OWN release listing -- the line below
# used to build the right FILENAME from $qv and then grep for it in NeuRT's listing,
# which passes only while the two tags happen to be equal. v0.20.32 was the first
# release to diverge them ... and the gate reported the QHexRT assets missing from a
# release that was never supposed to carry them.
```

That is the case this PR was opened to prevent, and it is fixed. Nothing for me to add there.

**Still open, and what this PR now is.** The asset-name comparison is still an unanchored regex match. `grep -qx` treats every `.` as "any character", and these names are almost entirely dots:

```
$ listing=$'qhexrt-arm64-v1X2X3Xtar.gz'
$ grep -qx 'qhexrt-arm64-v1.2.3.tar.gz' <<<"$listing" && echo MATCHED
MATCHED
$ grep -qxF 'qhexrt-arm64-v1.2.3.tar.gz' <<<"$listing" || echo "rejected with -F"
rejected with -F
```

So a release carrying a corrupted or wrongly-named asset can satisfy a gate whose entire job is to prove the right assets are present. It needs a name that differs only where the dots are, which is unlikely by accident but is exactly what the check exists to catch.

Six call sites, none of them using `-F`:

```
271, 272   NeuRT listing
278, 280   QHexRT listing
303, 304   QAIRT listing
```

All six now use `grep -qxF`. That is the whole diff.

Rebased onto `488cf27b2` (0.20.35). `bash -n` clean; I could not run shellcheck, it is not installed here.

sanchitmonga22 (approved):
Thanks a lot for this, @ayaangazali! Dots in release asset names (NeuRT, QHexRT and QAIRT) were being read as regex wildcards by `grep -qx`, so a malformed asset name could slip past the presence check meant for the well-formed one -- switching all six call sites to `grep -qxF` closes that gap.

Checked: CodeRabbit reviewed the latest commit · CI green · built and linted locally merged into `main` (scripts/validation/gates/**) · two independent code reviews.

PR #786, opened 2026-08-25
fix(flutter): forward contextLength, accelerator, and backendPreferences on model load

shubhamsinnh:
## What

`ModelGate.load()` (the shared auto-load gate behind every generation verb) logged `contextLength`, `threads`, and `useGpu` as "not carried by the commons load ABI yet", although the current IDL carries them: `ModelLoadRequest` has `context_length=6`, `use_gpu=8`, `backend_preferences=9`, and `accelerator_policy=10` (`idl/model_types.proto`), and commons reads them in `core/src/core/model_lifecycle.cpp`.

## Changes

- `bindings/flutter/packages/runanywhere/lib/public/api/internal/model_gate.dart`:
  - `ignoredLoadOptionKnobs` now reports only `threads` (retired from the load ABI at `ModelLoadRequest` reserved tag 7).
  - `load()` wires `contextLength`, `acceleratorPolicy` (from `resolvedAccelerator`, folding the deprecated `useGpu` alias), and `backendPreferences` (from `resolvedBackendPreferences`, folding the deprecated `framework` alias) onto `ModelLoadRequest`. The single-pin `framework` field stays aligned with the ranked list's winner.
- `bindings/flutter/packages/runanywhere/test/model_gate_load_options_test.dart`: updated to assert only `threads` is reported as not carried.

## Notes

- No open PR touches `model_gate.dart`; PR #755 is engine-only (llama.cpp VLM create path).
- Dart/Flutter/protoc toolchains are not installed on the contributor machine, so `flutter analyze`/`flutter test` run in CI (generated `model_types.pb.dart` is produced by `idl/codegen/generate_dart.sh`).

## Summary by CodeRabbit

* **New Features**
  * Model loading now reports detailed results, including resolved framework, category, device, runtime, ABI versions, and fallback reasons.
  * Load options now support context length, accelerator and backend preferences, and configurable force-reload behavior.
  * Already-loaded models can be recognized without reloading.
  * Model metadata is retained when load results omit optional details.

* **Bug Fixes**
  * Invalid required backend preferences are rejected.
  * Diagnostic values are normalized for clearer reporting.
  * Unsupported thread settings now produce a warning.


sanchitmonga22:
Thanks for this, @shubhamsinnh! Forwarding contextLength, accelerator, and backendPreferences into the load ABI closes a real gap -- Flutter callers can finally get the placement info Kotlin already returns, and `forceReload` now actually reloads an already-loaded model.

1. `bindings/flutter/packages/runanywhere/lib/public/api/internal/model_gate.dart:93` (the already-loaded shortcut in `ensureLoaded`) -- when the model is already loaded, this returns a `ModelLoadResult` built from `CurrentModelResult`, which carries no placement fields, so `models.load()` reports `actualDevice: 'unknown'`. It also never reaches commons, so its per-category option validation (`core/src/core/model_lifecycle.cpp:674-710`) doesn't run: `models.load(sttId, LoadOptions(contextLength: 4096))` now fails when the model isn't loaded but succeeds silently when it already is.
Commons' own `load()` runs that validation before its same-model check and then returns the resident model's full placement (`model_lifecycle.cpp:769`), so routing an explicit `models.load()` through `load()` -- keeping the Dart shortcut only for the generation verbs that call `ensureLoaded` implicitly -- would fix both.

Rejecting `BackendPreference(required: true)` up front is fine as it stands: `ModelLoadRequest` has no way to carry `required`, and Web does the same in #785. #882 tracks lining up these load options, including `required`, across the SDKs.

PR #789, opened 2026-08-25
chore: delete the three workflows under core/.github

ayaangazali:
## Description

`core/.github/workflows/` holds three workflow files. GitHub Actions only reads `.github/workflows/` at the **repository root**, so nothing in that directory has ever run here.

```
core/.github/workflows/build-commons.yml   137 lines
core/.github/workflows/release.yml         292
core/.github/workflows/size-check.yml       86
                                           515 total
```

They are leftovers from when commons was its own repository, carried along by the 0.20.17 restructure (#665, the last commit to touch that path) rather than maintained. Three independent signs of that:

- **Stale action pin.** They use `actions/checkout@v4`; every live workflow is on `@v7`.
- **Stale path filters.** `build-commons.yml` and `size-check.yml` filter on `include/**`, `src/**`, `cmake/**`, `CMakeLists.txt`. `paths:` is evaluated from the repo root, and this repo has no top-level `include/` or `src/` — those describe the pre-restructure standalone layout, not `core/include` and `core/src`.
- **Superseded.** The root `release.yml` publishes a superset: the dead copy names `RACommons-android-*` and `RACommons-ios-*`, the live one produces those plus `RACommons-linux-*` and the rest.

Nothing can reach them either. No file in the repo mentions `core/.github` (grepped `*.yml`, `*.sh`, `*.md`, `*.py`), none of the three declares `workflow_call`, so they are not reusable workflows another job could invoke, and `core/` is not synced or mirrored to any other repository where they would execute.

Deleting rather than fixing: a workflow that cannot run is not worth repairing, and leaving it invites someone to edit it expecting an effect.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [x] Refactoring

## Testing
- [ ] Lint passes locally — no lintable source changed; this is a pure deletion of three unreferenced YAML files
- [ ] Added/updated tests for changes — not applicable to deleting dead CI config

Ran the full gate suite on the branch. Everything passes except two failures that are pre-existing on `main` and unrelated:

- `check_no_hardcoded_defaults.sh` — `mapfile: command not found`, macOS bash 3.2, the subject of my open #729
- `check_rn_cpp_headers_compile.sh` — needs `yarn install` in `bindings/react-native`, environmental

Also confirmed `.github/workflows/` at the root is untouched (still 11 files) and `core/.github` is now empty.

### Platform-Specific Testing

## Labels
Closest is `core` by path, though nothing in the built `core` library changes.

## Checklist
- [x] Self-review completed
- [ ] Documentation updated (if needed) — no doc references these files

## Screenshots
Not applicable.

## Summary by CodeRabbit

- **Chores**
  - Removed automated build verification for macOS, iOS, Android, and C++.
  - Removed automated packaging and publishing of iOS and Android library releases.
  - Removed pull request size checks for the iOS framework, including generated size reports and size-limit validation.


ayaangazali:

The three workflow files are still there:

```
$ ls core/.github/workflows/
build-commons.yml
release.yml
size-check.yml
```

GitHub only ever runs workflows from `.github/workflows/` at the repository root, so a nested `core/.github/workflows/` is inert: these three have never executed and cannot. The risk is not the wasted files, it is that they read as coverage. `build-commons.yml` in particular looks like the commons build gate while the real one lives in the root `pr-build.yml`.

Deleting them is the whole diff. Merges clean, no conflicts.

sanchitmonga22:
@coderabbitai full review

sanchitmonga22:
Thanks for this, @ayaangazali! The change looks good — the diff is exactly what it says, a clean deletion of the three inert workflow files under `core/.github/workflows/`.

One small, optional thing while it waits: `core/AGENTS.md:290-293` still describes those workflows as live ("This directory carries its own `.github/workflows/`…", then `build-commons.yml` and `size-check.yml`). Dropping or rewording that section in this PR would keep the docs in step with the deletion.

Because it touches CI/automation files (`build-commons.yml`, `release.yml`, `size-check.yml` — files that would run with the repo's tokens if GitHub ever read them), a maintainer will give that part a final look before merging.

sanchitmonga22:
> Replying to [this comment](https://github.com/RunanywhereAI/runanywhere-sdks/pull/789#issuecomment-5535060352)

That makes sense — thanks for explaining! We verified independently that these three files never ran: GitHub only reads workflows from the repository root, and none of this repo's registered workflows live under `core/`, so your reasoning holds. The one leftover is `core/AGENTS.md:290-293`, which still describes them (see the note above). Because the change touches CI/workflow files, a maintainer will still give it a final look before merging — thanks for your patience.

PR #790, opened 2026-08-25
chore(gates): drop two deprecated-surface exemptions that no longer apply

ayaangazali:
## Description

`deprecated_surface_allowlist.txt` carries 120 entries; only 118 of them still correspond to a real detection. These two are vacuous:

```
bindings/kotlin/.../LLM/StructuredOutputProtoHelpers.kt|kotlin:json-serialisation
core/include/rac/features/llm/rac_llm_structured_output.h|cpp:json-param-in-public-api
```

Neither is a moved or deleted file. Both still exist; they simply no longer match the rule they are exempted from:

```
$ grep -cE "(toJson|fromJson|JsonObject)" .../StructuredOutputProtoHelpers.kt
0
$ grep -cE "(config_json|out_json|out_stats_json)" core/include/rac/features/llm/rac_llm_structured_output.h
0
```

That matters because of what an entry in this file means. The gate's own message calls each one a "narrowly scoped exception", and its header says it "Exits 1 when NEW (non-allowlisted) violations are found". An exemption for a surface that is already gone is not narrow, it is a standing pre-approval: put `toJson` back into that Kotlin helper, or a `config_json` parameter back into that header, and the gate stays silent on exactly the regression it exists to catch.

Both are structured-output paths, so this looks like the residue of that migration finishing rather than anything going wrong.

## How I found them

Ran the gate with the allowlist temporarily moved aside, which makes it report every detection rather than only unlisted ones, then diffed that set against the file:

```
detected:     118
allowlisted:  120
STALE (allowlisted, no longer detected):  the two above
detected but unlisted:                    none
```

The empty second set is the part worth stating: the gate is not hiding anything today, so this is purely about the two entries that have stopped doing work.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [x] Refactoring

## Testing
- [x] Lint passes locally — `check_deprecated_surfaces.sh` exits 0 with the two entries removed: "All detected surfaces match documented exceptions. No new regressions."
- [ ] Added/updated tests for changes — not applicable to removing two allowlist lines

Also confirmed the removal does not flip the gate: it passed before and after, and no previously-listed path became a new violation.

### Platform-Specific Testing

## Labels
Closest is `core` by path, though the change is to a validation allowlist rather than to shipped code.

## Checklist
- [x] Self-review completed
- [ ] Documentation updated (if needed) — the file documents itself through its entries

## Screenshots
Not applicable.

## Summary by CodeRabbit

* **Chores**
  * Removed validation allowlist entries for Kotlin and C++ structured-output JSON APIs.
  * Added validation allowlist entries for C++ image-embedding and OCR service headers.

### DIFF
--- a/scripts/validation/gates/deprecated_surface_allowlist.txt
+++ b/scripts/validation/gates/deprecated_surface_allowlist.txt
@@ -8,7 +8,6 @@
 # Kotlin typed JSON adapters and external cloud-provider payloads.
 bindings/kotlin/src/main/kotlin/com/runanywhere/sdk/hybrid/CloudSttProvider.kt|kotlin:org-json-usage
 bindings/kotlin/src/main/kotlin/com/runanywhere/sdk/hybrid/Cloud.kt|kotlin:org-json-usage
-bindings/kotlin/src/main/kotlin/com/runanywhere/sdk/public/extensions/LLM/StructuredOutputProtoHelpers.kt|kotlin:json-serialisation
 # Web-search tool parses the external DuckDuckGo JSON response (not an SDK DTO);
 # mirrors the already-allowlisted Swift RunAnywhere+WebSearchTool.swift.
 bindings/kotlin/src/main/kotlin/com/runanywhere/sdk/public/extensions/LLM/RunAnywhereWebSearchTool.kt|kotlin:json-serialisation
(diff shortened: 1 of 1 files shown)


ayaangazali:

Both exemptions this removes are still in `deprecated_surface_allowlist.txt`, and both name surfaces that no longer exist, so they exempt nothing. An allowlist entry that matches no path is not harmless: it is indistinguishable from a live exemption when someone reads the file to find out what is deliberately tolerated, and it survives forever because nothing fails when it stops matching.

The diff is the two lines. No gate behaviour changes, which is checkable by running `check_deprecated_surfaces.sh` before and after.

Merges clean.

ayaangazali:
Folded into #802 and closing this.

Both changes touch the same gate: #802 removes a scan arm from `check_deprecated_surfaces.sh`, and this removes two entries from the allowlist that same script reads. Splitting them into two PRs meant two reviews of one file pair, which was my mistake.

The finding is unchanged and re-verified on `45e91276e`: neither entry appears in the gate's detected-surface output, so neither suppresses a real detection, and the gate still passes with the allowlist at 120 entries instead of 122.

PR #795, opened 2026-08-26
fix(hybrid-stt): roll back service creation when model-id allocation fails

shubhamsinnh:
## Summary

- check the hybrid STT service model-id duplication before returning the service
- destroy the backend instance and free the service wrapper when duplication fails

## Validation

- `g++ -std=c++20 -fsyntax-only` targeted syntax check
- `git diff --check`

Full Windows debug configuration was unavailable because Visual Studio 17 2022 is not installed on the validation host.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved service creation error handling to safely clean up resources when allocation fails.
  * Prevented incomplete services from being returned after an initialization failure.


sanchitmonga22:
@coderabbitai full review

sanchitmonga22 (approved):
Thanks a lot for this, @shubhamsinnh! In `rac_stt_hybrid_router_create_service`, if copying the model-id tag failed for lack of memory, the hybrid STT service factory used to return a half-initialized service (NULL `model_id`) as if creation had succeeded. Now it tears the service back down and returns NULL, matching the pattern used a few lines above for the sibling allocation-failure case.

Checked: CodeRabbit reviewed the latest commit (including a full re-review at this exact head) · CI green · built and tested locally merged into `main` (C++ commons build+tests: cmake/ctest macos-debug, 104/104 passing) · two independent code reviews, no findings.

PR #801, opened 2026-08-26
fix(llamacpp): stop discarding gpu_layers=-1, the "offload every layer" request

ayaangazali:
## Description

`a46517cf1` ("honour accelerator_policy and context_length instead of dropping them") made the load knobs reach the llama.cpp backend. They reach it, and then the backend drops the GPU case anyway.

`rac_llm_llamacpp_create` emits `gpu_layers = -1` for `ACCELERATOR_POLICY_GPU`, and omits the key entirely when the caller has no opinion:

```cpp
// AUTO means "no opinion"; every other value, INCLUDING 0, is an
// explicit placement request and must reach the backend.
if (config->gpu_layers != RAC_LLM_LLAMACPP_GPU_LAYERS_AUTO) {
    model_config["gpu_layers"] = config->gpu_layers;
}
```

`LlamaCppTextGeneration::load_model` then re-uses `-1` as its own local "not set" sentinel:

```cpp
int user_gpu_layers = -1;  // -1 = not set by user
if (config.contains("gpu_layers")) {
    user_gpu_layers = config["gpu_layers"].get<int>();
}
...
if (user_gpu_layers >= 0) {
    model_params.n_gpu_layers = user_gpu_layers;
}
```

The two meanings of `-1` collide. In the public API `-1` is "offload every layer" (`RAC_LLM_LLAMACPP_CONFIG_DEFAULT` documents it as *All layers on GPU*); here it means "the caller said nothing". An explicit GPU request therefore never survives the `>= 0` guard.

`ACCELERATOR_POLICY_CPU` maps to `0`, which passes `>= 0` and applies correctly. That asymmetry is what hides this: pinning to CPU works, asking for GPU silently does not.

This is not only a lost hint. For any non-`SUCCESS` `common_fit_params` outcome the fit has already fallen back to `n_gpu_layers = 0`, and applying the override is precisely what reinstates the offload — the comment directly below the guard says so ("honouring the user override here reinstates the OOM risk the fit call was supposed to prevent ... keep honouring the user's explicit request"). So a caller who asked for GPU gets a fully CPU-resident model.

## The change

Uses **presence of the JSON key** as the signal rather than a sentinel value. The producer already omits the key when the caller has no opinion, so `contains("gpu_layers")` is exactly "the caller asked for something". That removes the collision by construction instead of trading `-1` for another magic number.

The CPU-only-build warning is widened the same way (`> 0` → `!= 0` with the presence flag), so a `-1` GPU request on a build with no GPU backend now reports that it is being ignored instead of staying silent.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring

## Testing

Both preprocessor arms of the edited region were compiled, since the two guards select different code on different CI targets:

```
cmake --preset macos-debug -DRAC_BUILD_BACKENDS=ON
ninja engines/llamacpp/CMakeFiles/rac_backend_llamacpp.dir/llamacpp_backend.cpp.o   # exit 0
```

That build defines neither `GGML_USE_METAL`/`CUDA`/`VULKAN`/`WEBGPU`, so it compiles the CPU-only arm. I re-ran the same compile command with `-DGGML_USE_METAL` added to cover the `#else` arm — also exit 0.

- [ ] Lint passes locally

No test added: the affected code is inside `load_model`, which needs a real GGUF model and a live llama.cpp context, and the existing `core/tests/test_llm.cpp` cases skip themselves when no model is present. A test that cannot fail on the unfixed code is not worth adding. Happy to add one if you have a fixture for this.

## Labels
- [ ] `Commons`

(The change is in `engines/llamacpp`, which none of the listed labels cover exactly.)

## Checklist
- [x] Self-review completed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GPU-layer override handling so omitted settings, zero, and unlimited-layer values are interpreted distinctly.
  * CPU-only builds now clearly report when explicit nonzero GPU-layer overrides are ignored.
  * GPU-capable builds consistently apply configured overrides, including unlimited-layer settings, and warn when they replace a failed or errored automatic fit.

### DIFF
--- a/engines/llamacpp/llamacpp_backend.cpp
+++ b/engines/llamacpp/llamacpp_backend.cpp
@@ -426,9 +426,14 @@ bool LlamaCppTextGeneration::load_model(const std::string& model_path,
     model_params.load_mode = LLAMA_LOAD_MODE_NONE;
 #endif
 
-    // If common_fit_params aborts, use the user-provided value
-    int user_gpu_layers = -1;  // -1 = not set by user
-    if (config.contains("gpu_layers")) {
+    // If common_fit_params aborts, use the user-provided value.
+    // Presence of the key is the signal, not a sentinel value: -1 is a real
+    // request ("offload every layer"), which is what rac_llm_llamacpp_create
+    // emits for ACCELERATOR_POLICY_GPU. It omits the key entirely when the
+    // caller has no opinion, so `contains` is exactly "the caller asked".
+    const bool have_user_gpu_layers = config.contains("gpu_layers");
+    int user_gpu_layers = 0;
+    if (have_user_gpu_layers) {
         user_gpu_layers = config["gpu_layers"].get<int>();
         RAC_LOG_INFO("LLM.LlamaCpp", "User-provided GPU layers: %d (will apply after fit)",
                      user_gpu_layers);
@@ -558,7 +563,7 @@ bool LlamaCppTextGeneration::load_model(const std::string& model_path,
                      "GPU memory but no GPU backend active. "
                      "Applying conservative CPU defaults.");
     }
-    if (user_gpu_layers > 0) {
+    if (have_user_gpu_layers && user_gpu_layers != 0) {
         RAC_LOG_INFO("LLM.LlamaCpp",
                      "CPU-only build: ignoring user gpu_layers=%d (no GPU backend "
                      "available)",
@@ -570,7 +575,7 @@ bool LlamaCppTextGeneration::load_model(const std::string& model_path,
         RAC_LOG_INFO("LLM.LlamaCpp", "CPU-only: capping context to %u", ctx_params.n_ctx);
     }
 #else
-    if (user_gpu_layers >= 0) {
+    if (have_user_gpu_layers) {
         // common_fit_params fell back to n_gpu_layers=0 for non-SUCCESS outcomes;
         // honouring the user override here reinstates the OOM risk the fit call
         // was supposed to prevent. Log a warning so it's visible in the event of
--- a/engines/llamacpp/llamacpp_backend.cpp
+++ b/engines/llamacpp/llamacpp_backend.cpp
@@ -21,6 +21,7 @@
 #include <cstdint>
 #include <cstring>
 #include <fstream>
+#include <optional>
 #include <string>
 #include <sys/stat.h>
 #include <vector>
@@ -431,12 +432,11 @@ bool LlamaCppTextGeneration::load_model(const std::string& model_path,
     // request ("offload every layer"), which is what rac_llm_llamacpp_create
     // emits for ACCELERATOR_POLICY_GPU. It omits the key entirely when the
     // caller has no opinion, so `contains` is exactly "the caller asked".
-    const bool have_user_gpu_layers = config.contains("gpu_layers");
-    int user_gpu_layers = 0;
-    if (have_user_gpu_layers) {
+    std::optional<int> user_gpu_layers;
+    if (config.contains("gpu_layers")) {
         user_gpu_layers = config["gpu_layers"].get<int>();
         RAC_LOG_INFO("LLM.LlamaCpp", "User-provided GPU layers: %d (will apply after fit)",
-                     user_gpu_layers);
+                     *user_gpu_layers);
     }
 
     // Set up context params early for common_fit_params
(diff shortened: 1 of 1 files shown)


ayaangazali:
On the typed-optional suggestion: I agree the tri-state (omitted / `0` / `-1`) deserves to be explicit somewhere, but I do not think this PR is the place, for two reasons.

`load_model` is declared as `bool load_model(const std::string& model_path, const nlohmann::json& config = {})`, so raw JSON is the parameter type, and all five knobs it reads across its 320 lines go through the same `contains` check:

```
408: config.contains("context_size")
411: config.contains("max_context_size")
431: config.contains("gpu_layers")
487: config.contains("fit_margin_mib")
495: config.contains("n_ctx_min")
```

Making `gpu_layers` the only typed one would leave the function reading four knobs one way and a fifth another. Deserializing the whole config into a typed structure means changing the signature and every caller, which is a real refactor with a different blast radius than a fix for an explicit GPU request being silently dropped.

The second reason is that presence already *is* the contract, not a proxy for it. `rac_llm_llamacpp_create` omits the key exactly when the caller has no opinion:

```cpp
if (config->gpu_layers != RAC_LLM_LLAMACPP_GPU_LAYERS_AUTO) {
    model_config["gpu_layers"] = config->gpu_layers;
}
```

So `contains("gpu_layers")` maps one to one onto `has_value()`. The bug was that the old code inferred the third state from a *value* (`-1`) that is also a legitimate request; keying on presence removes that collision rather than re-encoding it.

Happy to do the typed-config pass over all five knobs as a separate PR if that is wanted.

ayaangazali:

```
$ grep -n gpu_layers engines/llamacpp/llamacpp_backend.cpp
430:    int user_gpu_layers = -1;  // -1 = not set by user
431:    if (config.contains("gpu_layers")) {
```

The sentinel and a real setting are the same number. `-1` is llama.cpp's own value for "offload every layer to the GPU", and it is the value this code uses to mean "the caller said nothing", so a caller who explicitly asks for full offload is indistinguishable from one who did not ask at all, and the later `>= 0` guard drops the request.

Line 462 in this same file sets `model_params.n_gpu_layers = -1` deliberately for the WebGPU path, which is what makes `-1` an ordinary value here rather than an impossible one.

The fix keeps `config.contains("gpu_layers")` as the presence test, which the code already computes, instead of overloading the value.

Merges clean, CI green.

ayaangazali:
Half of this is what the PR already does, and I have taken the other half.

**Already the case: presence rather than a sentinel.** That is the entire content of this PR. On `main` the code initialises `user_gpu_layers = -1` as a "not set" marker and later tests `>= 0`, which is exactly why `-1` (offload every layer) was being discarded. The diff replaced the sentinel with the presence of the JSON key, so `has_value()` is already the contract; it was just spelled as a separate bool.

**Taken: `std::optional<int>`.** Pushed above. A bool beside an int lets a caller read the int without consulting the bool, and the optional makes that impossible. Also rebased onto `main` (0.20.36) while I was there. Rebuilt `rac_backend_llamacpp`, 0 errors.

**Not taken: deserialising the load configuration into a typed structure.** The guideline about structured types is about the cross-boundary surface, where `idl/*.proto` is canonical and hand-written equivalents drift. This is a local `nlohmann::json` blob read at one call site inside one engine, and a new config struct would have exactly one producer and one consumer. I would rather not add a type whose only job is to be constructed and immediately destructured.

If the intent is that this engine should stop parsing a JSON blob at all and receive a generated type instead, that is a real and much larger change to the engine boundary, and worth its own discussion rather than a rider on a three-state fix.

sanchitmonga22:
> Replying to [this comment](https://github.com/RunanywhereAI/runanywhere-sdks/pull/801#issuecomment-5592021653)

That makes sense -- thanks for explaining. Scoping this PR to the three-state fix and treating the full config-struct refactor as its own discussion is the right call: `config.contains(...)` already gives you `has_value()` semantics for all five `load_model` knobs today, so nothing functional is being left behind here. Treating this as addressed -- if you'd like to open a separate issue for the config-struct idea, we'd welcome it.

sanchitmonga22 (approved):
Thanks a lot for this, @ayaangazali! This fixes a real bug: an explicit GPU request (`accelerator_policy=GPU` / `use_gpu=true`, which the SDK sends to the engine as `gpu_layers=-1`) was silently discarded by the llama.cpp backend's own `-1` "not set" sentinel, so those loads got the automatic fit's placement or CPU-only instead of full GPU offload. Tracking "the caller set this key" with `std::optional<int>` instead of overloading `-1` closes that gap cleanly.

Checked: CodeRabbit reviewed the latest commit · CI green · built and linted locally merged into `main` (C++ engines, fast validation gates) · two independent code reviews.

Follow-ups, not blocking: #888 (warn when a GPU-layers override replaces a successful fit's placement) and #889 (align `RAC_LLM_LLAMACPP_CONFIG_DEFAULT.gpu_layers` with the `AUTO` macro) -- you're welcome to pick these up.

PR #802, opened 2026-08-26
fix(gates): the deprecated-surfaces gate has a dead scan arm and two vacuous exemptions

ayaangazali:
## Description

Two problems in the deprecated-surfaces gate. They are one concern, because the gate is one thing: `check_deprecated_surfaces.sh` and the allowlist it reads.

**1. Check 6d scans a directory that cannot match.** The check's stated principle is right:

> Packaged header mirrors are public API too. A clean canonical header is insufficient if SwiftPM or an AAR still advertises a retired declaration.

The SwiftPM arm holds, because `bindings/swift/Sources/RunAnywhere/CRACommons/include` is committed and can drift from `core/include` on its own. The React Native arm cannot:

```
$ ls -d bindings/react-native/packages/core/android/src/main/jniLibs/include
absent on HEAD

$ git check-ignore -v bindings/react-native/packages/core/android/src/main/jniLibs/include
.gitignore:365  bindings/react-native/packages/*/android/src/main/jniLibs/

$ git ls-files bindings/react-native/packages/core/android/src/main/jniLibs
(nothing tracked)
```

That tree is a verbatim `cp -R core/include/rac` performed by `package-sdk.sh`. It is gitignored, nothing is committed under it, and it does not exist when the gate runs. So it can hold no declaration check 6c has not already seen, and the scan is guaranteed to find nothing. A check that cannot fail is not coverage; it just reads like it in the gate's output. Same reasoning for that path in the `rac_extract_archive` list.

I kept the principle and rewrote the comment to say why there is deliberately no RN arm, so nobody adds it back.

**2. Two allowlist entries suppress nothing.** Neither `StructuredOutputProtoHelpers.kt|kotlin:json-serialisation` nor `rac_llm_structured_output.h|cpp:json-param-in-public-api` appears anywhere in the gate's detected-surface output, so neither is exempting a real detection. Both files still exist; they simply do not trip the patterns any more.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring

## Testing
- [ ] Lint passes locally

Ran the gate itself, which is the only thing that can judge this:

```
before:  exit 0, "All detected surfaces match documented exceptions", allowlist 122 entries
after:   exit 0, "All detected surfaces match documented exceptions", allowlist 120 entries
```

The two removed entries do not appear in either run's detected-surface list, which is the evidence that they were vacuous rather than load-bearing.

No test added. This is a shell gate with no test harness in the repo, and the gate's own exit status is the check.

None. Repository tooling only, nothing shipped.

## Labels
**SDKs:**
- [ ] `Swift SDK`
- [ ] `Kotlin SDK`
- [ ] `Flutter SDK`
- [ ] `React Native SDK`
- [ ] `Web SDK`
- [ ] `Commons`

None apply: `scripts/validation/gates/` is CI tooling, not an SDK surface.

## Checklist
- [x] Self-review completed

## Summary by CodeRabbit

* **Chores**
  * Updated validation checks to focus on available committed header mirrors.
  * Clarified handling of generated React Native header files during validation.
  * Removed obsolete exceptions for retired structured-output serialization and parameter formats.

### DIFF
--- a/scripts/validation/gates/check_deprecated_surfaces.sh
+++ b/scripts/validation/gates/check_deprecated_surfaces.sh
@@ -306,11 +306,14 @@ scan_grep "core/include" \
   "(RAC_DEPRECATED|__attribute__\\(\\(deprecated|\\[\\[deprecated)" "*.h" "cpp:deprecated-declaration"
 
 # 6d. Packaged header mirrors are public API too. A clean canonical header is
-# insufficient if SwiftPM or an AAR still advertises a retired declaration.
+# insufficient if SwiftPM still advertises a retired declaration, because that
+# mirror is committed and can drift from core/include on its own.
+# There is deliberately no React Native arm here: that package's header tree is
+# a verbatim `cp -R core/include/rac` performed by package-sdk.sh into a
+# gitignored jniLibs/include (.gitignore), so it cannot carry a declaration
+# 6c has not already seen, and it does not exist at all when this gate runs.
 scan_grep "bindings/swift/Sources/RunAnywhere/CRACommons/include" \
   "(RAC_DEPRECATED|__attribute__\\(\\(deprecated|\\[\\[deprecated)" "*.h" "swift-c:deprecated-declaration"
-scan_grep "bindings/react-native/packages/core/android/src/main/jniLibs/include" \
-  "(RAC_DEPRECATED|__attribute__\\(\\(deprecated|\\[\\[deprecated)" "*.h" "rn-c:deprecated-declaration"
 
 # 6e. Retired compatibility surfaces that must never reappear under a neutral
 # name or in one hand-maintained package mirror.
@@ -323,8 +326,7 @@ scan_git_grep \
 scan_git_grep \
   "rac_extract_archive[[:space:]]*\\(" \
   "all:retired-archive-adapter-api" "core/include" \
-  "bindings/swift/Sources/RunAnywhere/CRACommons/include" \
-  "bindings/react-native/packages/core/android/src/main/jniLibs/include"
+  "bindings/swift/Sources/RunAnywhere/CRACommons/include"
 
 # ---------------------------------------------------------------------------
 # Report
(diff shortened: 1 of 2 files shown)


ayaangazali:

The React Native arms of check 6d are still in the gate:

```
$ grep -n 'react-native' scripts/validation/gates/check_deprecated_surfaces.sh
247: scan_find "bindings/react-native" "json.hpp" "rn:vendored-nlohmann" ...
251: scan_grep_filtered "bindings/react-native/packages" ...
254: scan_grep_filtered "bindings/react-native/packages" ...
259: scan_grep_filtered "bindings/react-native/packages" ...
```

They scan `jniLibs/include`, which is gitignored, so the directory is absent in every clone and in CI. `scan_grep` opens with `[[ -d "${search_dir}" ]] || return 0`, so each arm returns success without looking at anything.

The evidence that removing them changes nothing observable is that the gate's output is byte-identical before and after: 125 lines, exit 0 both ways. That is what makes this a deletion rather than a behaviour change.

Merges clean.

ayaangazali:
Folded #790 into this PR and closed it, and rebased onto `main` (0.20.36).

Both changes are the same gate: this script and the allowlist it reads. Reviewing them as two PRs meant two passes over one file pair, which is my fault for splitting them. Title and body updated to cover both, with the gate run before and after as the evidence.

Nothing was dropped in the merge.

sanchitmonga22:
> Replying to [this comment](https://github.com/RunanywhereAI/runanywhere-sdks/pull/802#issuecomment-5535060487)

That makes sense — thanks for explaining. We independently confirmed jniLibs/include is gitignored, untracked, and absent on HEAD, so scan_grep's early return made those two arms provable no-ops. Treating this as addressed.

sanchitmonga22:
> Replying to [this comment](https://github.com/RunanywhereAI/runanywhere-sdks/pull/802#issuecomment-5607280308)

That makes sense — thanks for explaining. We checked: #790 is closed (not merged), its own diff described the same two removals plus the same two additions that landed here, and both new allowlist entries (rac_image_embedding_service.h, rac_ocr_service.h) do carry a real config_json parameter. Nothing was dropped in the fold-in. Treating this as addressed.

sanchitmonga22 (approved):
Thanks a lot for this, @ayaangazali! This removes two checks in the deprecated-surfaces CI gate that scanned a gitignored, untracked React Native header mirror that could never match, and drops two allowlist entries that no longer match anything in the two files they used to exempt — so the gate will actually catch a future regression there.

Checked: CodeRabbit reviewed the latest commit · CI green · built and linted locally merged into `main` (scripts/validation/gates/**) · two independent code reviews.

Follow-up, not blocking: #887 (making the gate report allowlist entries that match nothing) — you're welcome to pick it up.

PR #806, opened 2026-08-27
chore: release 0.20.30

sanchitmonga22:
Version-only patch release bundling four fixes validated during v0.20.29's
Windows Electron/RCLI device testing:

- `bridge.ts`'s `addSidecarDirToDspSearchPath` (#803) — the Electron
  ADSP_LIBRARY_PATH gap that broke `qwen3.8-27b-1bit-npu` generation
- `package-private-engine-overlay.sh`'s skel-file filter (#804) — RCLI's
  private QHexRT overlay silently dropped every skel file
- both already merged to main; this PR is version bump + lockfiles only

`scripts/validation/gates/check_release_version_coherence.sh` passes clean
with no TEMP-pin exceptions needed.

## Summary by CodeRabbit

* **New Features**
  * Released version **0.20.30** across supported SDKs and platform packages.
* **Bug Fixes**
  * Fixed Windows packaging so QHexRT models run correctly in packaged Electron applications.
  * Fixed RCLI QHexRT packaging to include required model support files, eliminating manual file copying.
* **Documentation**
  * Updated installation examples, version references, and changelogs to reflect version 0.20.30.
  * No Flutter-facing API changes are included in this release.


ayaangazali:
One checksum looks like it needs re-syncing before this ships: `RunAnywhereMLXRuntime` in `bindings/flutter/packages/runanywhere_mlx/ios/runanywhere_mlx.podspec:12`.

#805 recorded that v0.20.29 committed `d942788f...` while actually publishing `fc4b23a0...`. That literal is still on main, and this PR carries it forward unchanged while moving the podspec from `asset_version = '0.20.29'` to `'0.20.30'`.

Checked the four MLX assets in that podspec against what the v0.20.29 release actually published:

```
RABackendMLX               OK
RunAnywhereMLXRuntime      MISMATCH
    committed: d942788f1d8f08a9c0c8cc9b7d4c212c9b35382aa90978615eccf8a7e72eac5b
    published: fc4b23a067a497f1b3ff687f9fc436982f68882b93b946076d8b243d6a2b15fe
RunAnywhereMLXMetal        OK
RunAnywhereMLXResources    OK
```

For contrast, `Package.swift` at `sdkVersion = 0.20.29` is fully correct: all six of its binary-target checksums match the v0.20.29 `.sha256` assets. So the 0.20.29 sync landed everywhere except this one literal.

Two things make it worth catching here rather than after publish:

1. Per #805 the stripped Mach-O for this archive is not byte-reproducible, so a fresh 0.20.30 `native_ios` run will produce a digest different from both values above. Re-pointing `asset_version` at v0.20.30 while keeping a 0.20.29-era literal cannot resolve to a match.
2. `RAC_CHECKSUM_SKIP=RunAnywhereMLXRuntime` means the publish-time checksum gate is the one place that would otherwise flag it, and it skips exactly this archive.

The v0.20.29 instance only stayed harmless because `runanywhere_mlx` had not been pushed to pub.dev for that version yet. Re-running `bindings/swift/scripts/sync-checksums.sh` against the specific 0.20.30 run whose tag is being published, as #805 recommends, should cover it.

I have not touched the file, since this PR owns it.

PR #808, opened 2026-08-27
fix(desktop bindings): report unknown directory-entry sizes as zero

shubhamsinnh:
## Problem

`std::filesystem::file_size(path, error_code)` returns an error sentinel when an entry size cannot be determined. The Electron and Python desktop adapters cast that sentinel to `int64_t`, allowing `-1` to escape even though `rac_directory_entry_t::size_bytes` requires `0` for directories or unknown sizes.

## Fix

- set `size_bytes` to `0` whenever the size lookup reports an error
- apply the same fallback to Electron and Python on Windows and POSIX
- preserve successful regular-file sizes and existing directory enumeration behavior

## Scope

- `bindings/electron/native/win32_platform_adapter.cpp`
- `bindings/electron/native/posix_platform_adapter.cpp`
- `bindings/python/native/win32_platform_adapter.cpp`
- `bindings/python/native/posix_platform_adapter.cpp`

## Validation

- `git diff --check`
- targeted formatting checks for every changed range
- C++20 syntax compilation of both Win32 adapter translation units
- Electron Windows/macOS and Python Windows/macOS CI passed

POSIX compilation was not available locally because this Windows host has no Linux or WSL environment; the macOS binding jobs provide the corresponding platform compile coverage.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved directory listings when file-size information cannot be retrieved.
  * Entries with unavailable or invalid size data now report a size of zero instead of potentially displaying an incorrect value.
  * Applied consistently across supported desktop and Python integrations.


sanchitmonga22:
@coderabbitai full review

sanchitmonga22 (approved):
Thanks a lot for this, @shubhamsinnh! This makes the Electron and Python desktop directory listings report 0 (per the documented contract) instead of -1 when a file's size can't be read — for example a broken symlink, FIFO, or socket.

Checked: CodeRabbit reviewed the latest commit · CI green · built and linted locally merged into `main` (electron native addon + TS build/test, Python test suite) · two independent code reviews.

Follow-ups, not blocking: #890 (sharing the four hand-kept `list_dir` bindings' error handling with the core desktop adapter) — you're welcome to pick it up.

PR #812, opened 2026-08-27
fix(solutions): an explicit temperature of 0.0 never survives expansion

ayaangazali:
## Description

`idl/llm_options.proto` states the contract on the field itself:

```proto
// 0.0 = greedy decoding, and is honoured as an explicit request.
optional float temperature = 2 [
    (runanywhere.v1.rac_default) = "0.7",
    ...
];
```

The solutions path breaks it twice, on the same value, at both ends of the same hop.

**Writer** — `core/src/solutions/solution_converter.cpp`:

```cpp
if (cfg.has_generation() && cfg.generation().temperature() != 0.0f) {
    (*llm->mutable_params())["temperature"] = std::to_string(cfg.generation().temperature());
}
```

This uses the value as a presence test. `temperature` is `optional`, so `has_temperature()` exists and is the question actually being asked. An explicit `0.0` is indistinguishable from a caller who set nothing, and the param is never written into the operator spec. `core/src/foundation/rac_proto_adapters.cpp:351` already does this correctly with `if (in.has_temperature())`.

**Reader** — `core/src/solutions/operators/op_engine_backed.cpp`:

```cpp
options->set_temperature(temperature_ > 0.0f ? temperature_ : 0.8f);
```

Even when the param *is* present, a spec carrying `"temperature" = "0"` was re-read as unset and replaced. So both ends had to change for `0.0` to survive.

Net effect today: a caller asking for deterministic decoding through a voice-agent solution silently gets sampled decoding instead.

## The change

The reader now lets `param_float_or` supply the default, which it already does when the key is absent, and passes the parsed value through untouched. No sentinel value is involved on either side.

That also replaces a hardcoded `0.8f` with `RAC_DEFAULT_LLM_GENERATION_OPTIONS_TEMPERATURE`. **This changes the unset-case default from 0.8 to 0.7.** 0.7 is the canonical value in `llm_options.proto` and what every other temperature default in `core/src` uses (`rac_backend_platform_register.cpp:59,92`, `vlm_module.cpp:80`); this operator was the only `0.8` in the tree. Happy to restore the 0.8 if it was deliberate.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring

## Testing

Regression test added in `core/tests/test_solution_runner.cpp`, beside the existing voice-agent converter case rather than in a new file.

**Confirmed it fails on the unfixed code** by reverting only the converter guard and rebuilding:

```
[RUN ] voice_agent_explicit_zero_temperature_reaches_llm_operator
[FAIL] core/tests/test_solution_runner.cpp:1099 it != llm->params().end()
```

With the fix in place:

```
```

Both preprocessor configurations of the edited region were compiled, since `op_engine_backed.cpp` guards its body on `RAC_HAVE_PROTOBUF`: the normal build (define present) plus a re-run of the same compile command with `-DRAC_HAVE_PROTOBUF` stripped. Both exit 0.

- [ ] Lint passes locally

`clang-format` is not installed on this machine, so I matched surrounding style by hand and checked no added line exceeds the column limit rather than claiming a formatter run.

## Labels
- [x] `Commons`

## Checklist
- [x] Self-review completed

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed VoiceAgent configurations with an explicit temperature of `0.0` so the setting is preserved and correctly applied during text generation.
  * Unspecified temperature settings continue to use the default generation temperature.

* **Tests**
  * Added coverage to verify that an explicit zero temperature reaches the language model operator.

### DIFF
--- a/core/src/solutions/operators/op_engine_backed.cpp
+++ b/core/src/solutions/operators/op_engine_backed.cpp
@@ -48,6 +48,7 @@
 #include "rac/core/rac_platform_adapter.h"
 #include "rac/foundation/rac_proto_buffer.h"
 #include "rac/graph/pipeline_node.hpp"
+#include "rac/rac_defaults_generated.h"
 #include "rac/solutions/operator_registry.hpp"
 
 #if defined(RAC_HAVE_PROTOBUF)
@@ -192,7 +193,10 @@ class GenerateTextNode final : public OperatorNode {
         : PipelineNode(std::move(name), /*input*/ 8, /*output*/ 8, OverflowPolicy::BlockProducer),
           model_id_(spec.model_id()),
           max_tokens_(param_int_or(spec, "max_tokens", 0)),
-          temperature_(param_float_or(spec, "temperature", 0.0f)) {
+          // param_float_or already supplies this when the key is absent, so an
+          // explicit "0" in the spec stays 0 (greedy) instead of being re-read as unset.
+          temperature_(param_float_or(spec, "temperature",
+                                      RAC_DEFAULT_LLM_GENERATION_OPTIONS_TEMPERATURE)) {
         if (const std::string* p = find_param(spec, "system_prompt"); p) {
             system_prompt_ = *p;
         }
@@ -214,7 +218,7 @@ class GenerateTextNode final : public OperatorNode {
         message->set_role(runanywhere::v1::MESSAGE_ROLE_USER);
         message->set_content(item.text());
         auto* options = request.mutable_options();
-        options->set_temperature(temperature_ > 0.0f ? temperature_ : 0.8f);
+        options->set_temperature(temperature_);
         options->set_top_p(1.0f);
         if (!system_prompt_.empty()) {
             options->set_system_prompt(system_prompt_);
--- a/core/src/solutions/solution_converter.cpp
+++ b/core/src/solutions/solution_converter.cpp
@@ -65,7 +65,10 @@ void expand_voice_agent(const runanywhere::v1::VoiceAgentConfig& cfg, PipelineSp
     if (cfg.max_context_tokens() > 0) {
         (*llm->mutable_params())["max_context_tokens"] = std::to_string(cfg.max_context_tokens());
     }
-    if (cfg.has_generation() && cfg.generation().temperature() != 0.0f) {
+    // Presence, not value: llm_options.proto documents "0.0 = greedy decoding, and
+    // is honoured as an explicit request", so a caller asking for 0.0 must reach the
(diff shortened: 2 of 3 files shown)


ayaangazali:

```
$ grep -n 'temperature() != 0.0f' core/src/solutions/solution_converter.cpp
68:    if (cfg.has_generation() && cfg.generation().temperature() != 0.0f) {
```

`temperature` is `optional` in `llm_options.proto` with `rac_default "0.7"`, so `0.0` is a legal explicit setting (greedy decoding) and cannot double as "unset". A solution config asking for greedy silently gets 0.7 instead, which is the one temperature a caller picks deliberately.

Worth noting the shape has since turned up twice more in the same subsystem, which is why I think the presence-vs-value distinction here is worth taking rather than papering over: #823 fixed the same defect on the RAG query path (`top_k` unguarded, `temperature` guarded on the wrong presence bit), and #836/#839 fix knobs the solution loader parses that the expansion drops entirely.

Merges clean, CI green.

sanchitmonga22:
@coderabbitai full review

sanchitmonga22:
> Replying to [this comment](https://github.com/RunanywhereAI/runanywhere-sdks/pull/812#issuecomment-5535062352)

That makes sense — thanks for the fresh repro and for flagging #823/#836/#839, we'll take a look at those too. On the open question: we checked, and 0.7 (not 0.8) is the codebase's canonical temperature default — it's `RAC_DEFAULT_LLM_GENERATION_OPTIONS_TEMPERATURE`, generated from `idl/llm_options.proto`'s `rac_default="0.7"`, and already used the same way in `rac_llm_types.h:168` and `rac_vlm_types.h:261`. So we're keeping your change as-is rather than restoring 0.8. Merging now — thanks for the patience!

sanchitmonga22 (approved):
Thanks a lot for this, @ayaangazali! An explicit temperature of 0.0 (greedy decoding) on a Solutions voice agent was silently replaced with sampled decoding at 0.8 — you fixed both the converter's presence check and the operator's default, so the proto's documented behavior (0.0 is a legal explicit request, not "unset") is now honoured end to end.

Checked: CodeRabbit reviewed the latest commit · CI green · built and linted locally merged into `main` (C++ commons: core/src/solutions, core/tests) · two independent code reviews.

Follow-ups, not blocking: #891 (forwarding the rest of VoiceAgentConfig.generation through expansion) and #892 (options_from_request ignoring field presence for top_k/min_p) — you're welcome to pick these up, you clearly know this code well.

PR #813, opened 2026-08-27
fix(vad): an explicit activation_threshold of 0.0 cannot be requested

ayaangazali:
## Description

`VADOptions.activation_threshold` is `optional`, and `idl/vad_options.proto` puts the presence contract on the field itself:

```proto
// Unset = keep the loaded detector's calibrated value. Same normalized
// [0,1] scale as VADConfiguration.activation_threshold.
optional float activation_threshold = 1 [...];
```

Every reader instead uses the value as the presence test:

```cpp
const bool has_override = options.activation_threshold() > 0.0f;
```

`0.0` is inside the documented `[0,1]` range and the setter accepts it. `rac_vad_component_set_energy_threshold` rejects only out-of-range input:

```cpp
if (threshold < 0.0f || threshold > 1.0f) { ... return RAC_ERROR_INVALID_PARAMETER; }
```

and then warns separately at `< 0.002` ("may cause false positives"), so a near-zero threshold is plainly contemplated rather than nonsensical.

The energy detector's verdict is `mean_sq > energy_threshold_sq` (`energy_vad.cpp:523`), so a threshold of `0.0` means "treat any non-silent frame as speech", which is how a push-to-talk style caller disables gating. Today that request is indistinguishable from leaving the field unset, so it is silently dropped and the calibrated value stays in force.

## The change

Four `VADOptions` readers move to `has_activation_threshold()`: the override flag in `rac_vad_component_process_proto`, and both lifecycle-VAD `set_threshold` guards in `vad_module.cpp`.

`rac_vad_stream.cpp` carried the same conflation one level deeper. The session stored the override as a bare float whose `0.0f` doubled as "no override", so it needed a real flag rather than a different magic number. Worth noting the sibling field immediately below it already does this correctly:

```cpp
s.max_speech_ms = parsed.has_max_speech_duration_ms() ? std::max(0, parsed.max_speech_duration_ms()) : 0;
```

`VADConfiguration` is deliberately untouched: its fields are not `optional`, so value-based defaulting is the only thing available there and is correct as written.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring

## Testing

```
```

Both preprocessor configurations were compiled for each edited file, since both guard parts of their bodies on `RAC_HAVE_PROTOBUF`: the normal build, plus a re-run of each file's own compile command with `-DRAC_HAVE_PROTOBUF` stripped. All four compiles exit 0.

- [ ] Lint passes locally

**No test added, and I want to be straight about why rather than imply coverage I do not have.** I tried to add a behavioural regression test to `test_speech_proto_abi.cpp` (configure a high threshold, then process quiet audio with an explicit `0.0` override and assert it flips to speech). `rac_vad_component_start` returns non-success for a component with no model loaded, so the energy path cannot be driven through `rac_vad_component_process_proto` in that harness, and the mock-model path restores the original threshold after each call so the override is not observable afterwards. Rather than ship a test that passes on the unfixed code, I dropped it. If you can point me at the right seam I will add one.

`clang-format` is not installed here, so I matched surrounding style by hand rather than claiming a formatter run.

## Labels
- [x] `Commons`

## Checklist
- [x] Self-review completed

## Summary by CodeRabbit

* **Bug Fixes**
  * Explicitly setting the activation threshold to `0.0` now correctly applies it as an override.
  * Omitted thresholds are no longer sent as zero-valued overrides.
  * Streaming and lifecycle VAD processing now consistently distinguishes between omitted and explicitly provided thresholds.
  * Confidence calculations and detector behavior honor zero-valued threshold settings.
  * VAD processing reports threshold configuration failures instead of continuing with invalid settings.
  * Concurrent threshold overrides are handled consistently, preventing requests from interfering with one another.


ayaangazali:
Both review points are about the lifecycle path, so answering them together.

**1. "Restore the lifecycle model threshold after a per-request override" — real, but pre-existing, and not fixable at this layer today.**

The leak is real: nothing in `rac_vad_process_lifecycle_proto` restores the threshold after `ref.ops->set_threshold`. It is not introduced by this PR though. The same unrestored call is on `main`, just behind a different predicate:

```cpp
if (ref.ops->set_threshold && request.has_options() &&
    request.options().activation_threshold() > 0.0f) {
    (void)ref.ops->set_threshold(ref.impl, threshold);
}
```

So any non-zero override already persisted into the next request. This diff changes which values reach that call, not whether the call restores.

The suggested remedy is not available here: `rac_vad_service_ops_t` has `set_threshold` and **no getter**. The vtable is `process / start / stop / reset / set_threshold / is_speech_active / destroy / initialize / create`. With no way to read the current value there is nothing to preserve.

The component path can do save/restore only because it does not go through the vtable, it uses the component accessor:

```cpp
const float original_threshold = rac_vad_component_get_energy_threshold(handle);
```

which has no lifecycle-path equivalent. Doing this properly means adding a `get_threshold` slot to `rac_vad_service_ops_t`, which is a plugin ABI change (currently v9) and every backend has to implement it. I did not want to smuggle an ABI change into a four-line predicate fix. Happy to open that separately if you want it, or to fold it in here if you would rather it land together.

**2. "Reject an override that the lifecycle backend cannot apply" — I would rather you decide this one.**

You are right that both cases are silent today: a null `set_threshold` skips the override, and a failing one is discarded by the `(void)` cast, and `process` runs either way.

Two arguments pull against each other. The proto contract says a set `activation_threshold` is an explicit request, so honouring it partially and saying nothing is the same class of bug this PR is fixing. But turning that into `RAC_ERROR_NOT_SUPPORTED` makes `rac_vad_process_lifecycle_proto` start failing for any backend that does not implement `set_threshold`, for callers who are passing a threshold today and getting a usable result. That is a behavioural change with a blast radius well beyond this diff, and it is your call rather than mine to make in a review round.

If you want it, my preference would be to surface the setter's error and keep the null case degrading (with a log line), rather than failing both. Say which you want and I will push it.

ayaangazali:
Still reproduces on `45e91276e` (0.20.36). All five readers are still value-reads, and `has_activation_threshold()` appears nowhere in the file:

```
1213: proto.activation_threshold() > 0.0f ? proto.activation_threshold() : RAC_VAD_DEFAULT_ENERGY_THRESHOLD
1283: const bool has_override = options.activation_threshold() > 0.0f;
1639: if (request.has_options() && request.options().activation_threshold() > 0.0f)
1651:     request.options().activation_threshold() > 0.0f)
1743: proto.activation_threshold() > 0.0f ? proto.activation_threshold() : RAC_VAD_DEFAULT_ENERGY_THRESHOLD
```

Line 1283 is the clearest: a variable literally named `has_override` computed from a value comparison rather than presence. `activation_threshold` is `optional`, so `0.0` is a legal explicit setting (detect on any energy) and cannot also mean "unset".

Since I opened this, the same presence-vs-value defect has turned up in four more places in commons, which is why I think it is worth taking as a pattern rather than case by case: #812 (solutions temperature), #823 (RAG top_k and temperature), #828 (tool-calling auto_execute) and #841 (embeddings normalize).

Merges clean, CI green.

ayaangazali:
Two findings, one taken and one that cannot be built as described.

**Taken: rejecting an override the backend cannot apply.** Pushed above. The lifecycle path did `(void)ref.ops->set_threshold(...)` and then ran `process` regardless, so a caller whose threshold could not be applied still got a result, computed at the calibrated threshold, with nothing saying their request was dropped. It now returns `RAC_ERROR_NOT_SUPPORTED` when the backend has no setter and propagates the setter's error otherwise, releasing the lifecycle ref on both paths. That matches the `!ref.ops->process` check immediately below it. Build 0 errors, ctest 103/103.

**Cannot take as written: preserving and restoring the threshold.** The VAD ops vtable has no getter:

```
core/include/rac/features/vad/rac_vad_service.h:23
    rac_result_t (*set_threshold)(void* impl, float threshold);
```

That is the only threshold entry point. There is no `get_threshold`, so the lifecycle path cannot read the original value in order to put it back. Adding one means extending `rac_vad_service_ops_t`, which is a plugin ABI change and a `RAC_PLUGIN_API_VERSION` bump. I am not folding that into a one-line presence fix.

The concern behind it is real, and worth stating precisely so it is not lost: `rac_vad_component_process_proto` (same file, around line 1300) does a full get, set, process, restore under a per-handle mutex, and the lifecycle path does none of that. So the two entry points genuinely disagree about whether a per-request threshold is transient. That predates this PR, which only changes the condition from `> 0.0f` to `has_activation_threshold()`; the leak already applied to every threshold above zero, and this widens it to include an explicit `0.0`.

I have not added the sequential test you suggested for the same reason: with no getter, a test asserting the restore would be asserting behaviour the vtable cannot currently provide. Happy to file the getter plus restore as its own change if you want the two paths reconciled.

ayaangazali:
Taken, pushed above.

You are right that pinning the lifecycle ref only prevents an unload. It does nothing about a second request re-setting the threshold between this request's `set_threshold` and its `process`, which would hand this caller a result computed at the other caller's value.

`rac_vad_component_process_proto` already solves exactly this per handle, and says so:

> Without it, two threads on the same VAD handle (one one-shot, one streaming, or two of either) can interleave their set/restore calls and corrupt the persistent energy_threshold.

The lifecycle path has no handle, so the lock is module-wide (`lifecycle_threshold_mutex()`, sitting next to the existing `fallback_vad_mutex()` which is there for the same shared-state reason). It is a `unique_lock` taken only when an override is in effect, so the no-override path stays lock-free, matching the component path's tradeoff. It is held across `process` and released on every early return.

On restoring the original value inside that lock: still not possible. `rac_vad_service.h:23` exposes `set_threshold` and no getter, so there is nothing to read back. Your "when the lifecycle ABI can expose it" is the right framing, and it needs a vtable extension plus a `RAC_PLUGIN_API_VERSION` bump, which I am not putting inside this fix. The lock at least means the override and the processing it was meant for cannot be pulled apart.

Build 0 errors, ctest 103/103.

sanchitmonga22:
Thanks for this, @ayaangazali! Checking presence instead of value for `activation_threshold` is the right call — it's the only way to let an explicit `0.0` through, and it matches how the rest of the SDK treats `optional` proto fields.

Before we merge, a couple of things to sort out:

1. `bindings/kotlin/src/main/kotlin/com/runanywhere/sdk/public/api/MappingOptions.kt:162` — `activation_threshold = activationThreshold ?: 0f` turns an unset Kotlin threshold into a present, explicit `0.0` on the wire. With this PR that now hits the override path on every default `vad.detect()`/`openStream()` call, which fails against a loaded Silero model and can misclassify audio on the energy VAD. Dropping the `?: 0f` (mapping straight to the nullable `activationThreshold`) leaves an unset threshold unset on the wire.
2. `bindings/web/packages/core/src/Public/Extensions/RunAnywhere+VAD.ts:200` (and the `callOptions(options, 0)` calls in `detectVoice`/`streamVoiceActivity`) — Web sends a present `activationThreshold: 0` as a "no per-frame override" sentinel. After this PR that `0` is applied as a real override on every call, and the PR's own error-propagation change now surfaces the resulting Sherpa failure as a thrown exception. Leaving `activationThreshold` `undefined` there instead (and updating the two `toBe(0)` test assertions to `toBeUndefined()`) should fix it.

CodeRabbit also has an open point that still applies: https://github.com/RunanywhereAI/runanywhere-sdks/pull/813#discussion_r3874512475 — your explanation (no getter in `rac_vad_service_ops_t` without an ABI bump) checks out, so I've opened it as a follow-up (#896) instead of holding this PR on it, along with a second one (#897) for the Sherpa/Silero threshold floor.

Once the Kotlin and Web mappings are fixed we'll take another look. Thanks again!

sanchitmonga22:
> Replying to [this comment](https://github.com/RunanywhereAI/runanywhere-sdks/pull/813#issuecomment-5446767594)

Thanks for laying out both options! Failing outright (`RAC_ERROR_NOT_SUPPORTED` / propagating the setter's own error) is the right call here — it matches how the component path already handles a missing setter, and it's what's in the PR now. On the getter: let's keep it out of this PR — I opened #896 to track the restore/getter work separately, since it needs the ABI bump you flagged.

sanchitmonga22:
> Replying to [this comment](https://github.com/RunanywhereAI/runanywhere-sdks/pull/813#issuecomment-5591925906)

Agreed, please file it separately rather than folding it in here — I went ahead and opened #896 for the threshold-restore/getter follow-up so it's tracked now. Feel free to pick it up, or let us know if you'd rather someone else does.

ayaangazali:
Both taken, pushed above. You were right that this PR turns two existing sentinels into regressions, and neither was visible from the C++ side I was looking at.

**Kotlin.** `MappingOptions.kt` now maps straight across instead of through `?: 0f`. Checked the types rather than assuming the nullable mapping compiles:

```
public/api/Options.kt:164                      val activationThreshold: Float? = null
generated/.../v1/VADOptions.kt:49              public val activation_threshold: Float? = null
```

`Float?` to `Float?`, so an unset threshold stays unset on the wire.

**Web.** `callOptions` no longer defaults the threshold, and `detectVoice` / `streamVoiceActivity` pass `undefined` where they passed `0`, with the comment updated to say absence rather than zero is what means "no per-frame override".

One trap worth flagging, because it is the kind of thing that silently survives a review: this could not stay a defaulted parameter. With `activationThreshold = options?.activationThreshold ?? 0`, passing `undefined` at the call site re-triggers the default and puts the caller's own threshold straight back on the wire, which is close to the opposite of what those two sites want. The parameter is now required and explicit at both call sites.

**What I could not verify here, so you know where to look.** The Kotlin side I verified by reading the committed generated source above, but I could not compile it: `./gradlew compileDebugKotlin` fails in this checkout at `:modules:runanywhere-core-onnx:downloadJniLibs`, "Failed to download or validate RABackendONNX-android-arm64-v8a-v0.20.36.zip", which is unrelated to this diff. The Web side I could not typecheck at all, because `@runanywhere/proto-ts` is generated at build time and building it pulls the root workspace into a pre-existing `ERESOLVE` on `react-native-worklets` in `bindings/react-native/example`.

So the specific thing worth a second pair of eyes is whether the generated `VADOptions.activationThreshold` is `number | undefined`. If ts-proto emitted it as a required `number`, my `undefined` will not typecheck, and that would itself be the finding: Web would have no way to express "no threshold" and the generated type would need optionals enabled. CI's Web job will settle it either way, and I will fix it rather than leave it red.

ayaangazali:
Red check resolved, and the answer to my open question is better than I expected.

It was not a typecheck failure despite the job name. `web-typecheck` also runs the Web unit tests, and the generated `VADOptions.activationThreshold` does accept `undefined` (the typecheck itself passed). What failed were two existing assertions in `RunAnywhere+VADLifecycle.test.ts`:

```
220:  expect(request?.options?.activationThreshold).toBe(0);
253:  expect(counters.requests.every((r) => r.options?.activationThreshold === 0)).toBe(true);
```

Those tests were encoding the sentinel itself: they asserted the per-frame request carries a literal `0`. That is exactly the behaviour you asked me to remove, so the assertions had to move with it. Both now assert the per-frame override is absent.

Worth noting what I deliberately did not touch, since it is the part those tests were really protecting: line 213 still asserts `configurations[0].activationThreshold` is `0.2`, the caller's real threshold applied by `configureLifecycle`. That path is unchanged. The only thing that moved is the per-frame request, which now sends no override at all instead of `0`.

So the tests are slightly stronger than before: they pin the contract as "configure carries the threshold, the per-frame call carries nothing" rather than "the per-frame call carries zero".

PR #819, opened 2026-08-28
fix: correct the RunAnywhereMLXRuntime checksum for v0.20.31

ayaangazali:
## Description

`bindings/flutter/packages/runanywhere_mlx/ios/runanywhere_mlx.podspec` pins a checksum that does not match the archive v0.20.31 actually published:

```
committed  e2badb28a4496542a502fbfc5e019c784ed3cc109ac381586ec13c771bd0b3b5
published  7a24b950a53fe0e972584a6df016d700a246db69bae2285a2d3ce28ce4bafdbd
```

`git log -S` puts the committed value in #817, which synced checksums for v0.20.31. `RACommons` was fixed by that same PR and still matches, because it is byte-reproducible for a given source tree. `RunAnywhereMLXRuntime` is not, which is the whole reason `RAC_CHECKSUM_SKIP` exists, so when the release was re-run after #817 merged its archive got a new hash and the committed value went stale again. That is the superseded-build failure `AGENTS.md` describes, now on its fourth occurrence across 0.20.29 through 0.20.31.

**Nothing catches this.** `RAC_CHECKSUM_SKIP=RunAnywhereMLXRuntime` exempts this exact archive from `publish`'s `sync-checksums.sh --check`, which `AGENTS.md` already notes means it "would have shipped broken to Flutter consumers with no CI signal at all". pub.dev is still on 0.20.27, so it has not shipped yet, but the repo is wrong now and the next publish of `runanywhere_mlx` carries it.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring

## Testing

I did not trust the `.sha256` sidecar on its own. I downloaded the release asset and hashed it locally; both agree:

```
$ curl -sL -o mlxrt.zip .../v0.20.31/RunAnywhereMLXRuntime-ios-v0.20.31.zip
$ shasum -a 256 mlxrt.zip
7a24b950a53fe0e972584a6df016d700a246db69bae2285a2d3ce28ce4bafdbd
```

Per the "check every row of the checksum-location table, not only the one the CI error names" note in `AGENTS.md`, I checked the whole table against v0.20.31 rather than just this row:

```
Package.swift    RACommons / RABackendLLAMACPP / RABackendONNX
                 RABackendSherpa / RABackendNeuRT / RABackendMLX    all OK
mlx podspec      RABackendMLX                                       OK
                 RunAnywhereMLXRuntime                              MISMATCH  <-- this PR
                 RunAnywhereMLXMetal                                OK
                 RunAnywhereMLXResources                            OK
```

After the change all four MLX entries and all six SwiftPM targets resolve against v0.20.31. `scripts/validation/gates/check_release_version_coherence.sh` exits 0 (`[OK] release version coherence: 0.20.31`).

- [ ] Lint passes locally

No test: this is a single checksum literal whose correctness is the artifact comparison above, and there is no fixture that could assert it without network access to the release.

## Labels
- [x] `Flutter SDK`

## Checklist
- [x] Self-review completed

## Summary by CodeRabbit

* **Chores**
  * Updated the integrity verification for the iOS MLX runtime package.
  * No changes to download behavior or user-facing functionality.

### DIFF
--- a/bindings/flutter/packages/runanywhere_mlx/ios/runanywhere_mlx.podspec
+++ b/bindings/flutter/packages/runanywhere_mlx/ios/runanywhere_mlx.podspec
@@ -9,7 +9,7 @@
 
 mlx_checksums = {
   'RABackendMLX' => '0802fe58480c3e03e94498d46adb50fda7b9ade491807c7e0392a7fd68b83b59',
-  'RunAnywhereMLXRuntime' => 'e2badb28a4496542a502fbfc5e019c784ed3cc109ac381586ec13c771bd0b3b5',
+  'RunAnywhereMLXRuntime' => '7a24b950a53fe0e972584a6df016d700a246db69bae2285a2d3ce28ce4bafdbd',
   'RunAnywhereMLXMetal' => '17a2f8c4ce09ef691cde5e7d04171ce749fca89315205f90eb2eed5a76b682b1',
   'RunAnywhereMLXResources' => 'ace624c5cffa32f789cd3beee8d140740daadd793b1315c3583ab670410b3ca3'
 }.freeze


ayaangazali:
The red `centralization` here is not from this diff. Evidence rather than assertion:

**It is one step, and not a step this PR can influence.** Of the 14 steps in that job only step 11 failed, "Swift distribution repo (runanywhere-swift) is cut at this release". Steps 9 and 10, "Release-train version coherence" and its gate tests, both passed, and those are the ones that actually read the checksum this PR changes.

**The condition is documented in-tree.** `scripts/release/sync-versions.sh:616` says it plainly:

```
(scripts/validation/gates/check_swift_dist_repo_sync.sh fails every PR
 once v${NEW_VERSION} exists until that repo carries the ${NEW_VERSION} tag.)
```

`RunanywhereAI/runanywhere-swift` is currently tagged `0.20.30`, while this repo published `v0.20.31` at 2026-08-28T00:35:21Z. So the gate is correctly reporting that the distribution repo has not been cut for 0.20.31 yet.

**The pass/fail split follows the release boundary, not PR content:**

```
v0.20.31 published            2026-08-28T00:35:21Z

#819  centralization FAILURE  2026-08-28T16:27:17Z   (after)
#816  centralization SUCCESS  2026-08-27T19:49:10Z   (before)
#813  centralization SUCCESS  2026-08-27T17:55:15Z   (before)
#812  centralization SUCCESS  2026-08-27T17:46:31Z   (before)
#802  centralization SUCCESS  2026-08-26T17:43:01Z   (before)
```

Every run that postdates the release fails and every run that predates it passes, across PRs touching entirely unrelated files. Re-running #816 today would go red the same way.

I have not pushed anything for it, since the fix is cutting the dist repo (`bindings/swift/scripts/sync-dist-repo.sh` + push the 0.20.31 tag), which is a release step outside this PR.

Worth noting the two are related in one direction: `sync-dist-repo.sh` copies the same remote binaryTarget checksums, so cutting the dist repo before this correction lands would carry the stale `RunAnywhereMLXRuntime` value into the Swift distribution too.

ayaangazali:
Closing this: v0.20.32 superseded it.

The stale value this PR replaced is gone from `main` anyway, and the current entry matches the published artifact. I re-ran the same check that produced this PR, against the release's own `.sha256` sidecars rather than by eye:

```
RABackendMLX:           MATCH  c0fa8f3b5312e699c0bc37c648a4bcf0ff6b4634af1c874b74daf5aacc18e345
RunAnywhereMLXRuntime:  MATCH  b9b4c3316a0750b56b56ef10794be1d55a45b89e432e7304e4dcdb23f6a0d6ce
RunAnywhereMLXMetal:    MATCH  17a2f8c4ce09ef691cde5e7d04171ce749fca89315205f90eb2eed5a76b682b1
RunAnywhereMLXResources: MATCH ace624c5cffa32f789cd3beee8d140740daadd793b1315c3583ab670410b3ca3
```

All four `runanywhere_mlx.podspec` checksums on `main` now agree with `*-ios-v0.20.32.zip.sha256` on the release, so there is nothing left here to fix. Rebasing this onto 0.20.32 would leave an empty diff.

Worth noting for whoever owns the release path: the underlying gap that let 0.20.31 ship with a stale `RunAnywhereMLXRuntime` checksum was that `RAC_CHECKSUM_SKIP=RunAnywhereMLXRuntime` opts that archive out of both the update and the `--check` in `sync-versions.sh`, so the tooling cannot catch it. That is unchanged by this release, it just did not bite this time. Not reopening it here, only flagging it so the next occurrence is not a surprise.

Thanks for the release.

PR #829, opened 2026-09-01
docs(abi): the plugin ABI history stops at 9u but the define says 11u

ayaangazali:
## Description

`rac_plugin_entry.h` carries a version-by-version ABI history above the define. Every bump from `3u` to `9u` has an entry saying what changed and, crucially, why a plugin built against the previous version is now rejected:

```
 *   8u — revived reranking as a first-class primitive: promoted reserved
 *                 primitive wire value 11 to `RAC_PRIMITIVE_RERANK` and
 *                 renamed vtable reserved_slot_2 -> `rerank_ops` ...
 *   9u — LLM streaming token accounting: ... Engines compiled against v8
 *                 are rejected until rebuilt ...
 */
#define RAC_PLUGIN_API_VERSION 10u
```

v0.20.32 bumped the define to `10u` and did not add the matching entry, so the list runs 3, 4, 5, 6, 7, 8, 9 and then stops one short of the current value.

That is worth a line because of what this block is for. When an engine fails to register with `RAC_ERROR_ABI_VERSION_MISMATCH`, this is the file that tells its author what the new version requires. Right now a v9 engine is rejected and the document explaining every previous rejection is silent about this one.

The entry describes the v10 change and is sourced from the repo, not written from memory. `rac_engine_vtable.h:184` already documents the slot side of it:

```c
/** Image embedding (`RAC_PRIMITIVE_EMBED_IMAGE`). Promoted from reserved_slot_3 in ABI v10 —
 *  occupies the same binary offset, so the struct layout stayed stable across the promotion,
 *  exactly as rerank_ops did in v8. */
const struct rac_image_embedding_service_ops* image_embedding_ops;
```

and `rac_primitive.h:57` the wire side (`RAC_PRIMITIVE_EMBED_IMAGE = 12`, "Promoted from RAC_PRIMITIVE_RESERVED_12 in ABI v10").

Comment only, no code. I deliberately did not touch anything else: the functional side of the v10 promotion is complete, the `RAC_PRIMITIVE_TABLE` X-macro already carries its `X(RAC_PRIMITIVE_EMBED_IMAGE, image_embedding_ops, "embed_image")` row, so dispatch, capability reporting and manifest validation are all generated correctly. Only the history note was missing.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [x] Documentation update
- [ ] Refactoring

## Testing
- [ ] Lint passes locally

No test: this changes a block comment. Built anyway to confirm the header still compiles everywhere it is included.

```
$ cmake --build build -j 10
build exit=0, 0 errors
```

On lint: unchecked because `core/scripts/lint-cpp.sh` requires the repo's pinned `clang-format` and only Apple `clang-format 21` is available here.

## Labels
**SDKs:**

## Checklist
- [x] Self-review completed

## Summary by CodeRabbit

* **New Features**
  * Added support for image embedding and optical character recognition primitives in API versions 10u and 11u.
  * Engines may omit either capability when the corresponding operations are unavailable.

* **Bug Fixes**
  * Updated registration compatibility requirements: engines built against API versions 9u or 10u must be rebuilt to work with API version 11u.

### DIFF
--- a/core/include/rac/plugin/rac_plugin_entry.h
+++ b/core/include/rac/plugin/rac_plugin_entry.h
@@ -100,6 +100,19 @@ extern "C" {
  *                 Engines compiled against v8 are rejected until rebuilt;
  *                 NULL `get_stream_token_counts` means counts must be marked
  *                 estimated. Engine-vtable reserved_slot_3 is unchanged.
+ *  10u — promoted reserved primitive wire value 12 to image embedding
+ *                 (`RAC_PRIMITIVE_EMBED_IMAGE`) and renamed engine-vtable
+ *                 reserved_slot_3 -> `image_embedding_ops`, same binary
+ *                 offset, exactly as rerank_ops was promoted in v8. Engines
+ *                 compiled against v9 are rejected until rebuilt; a NULL
+ *                 `image_embedding_ops` means the engine does not serve the
+ *                 primitive.
+ *  11u — promoted reserved primitive wire value 13 to optical character
+ *                 recognition (`RAC_PRIMITIVE_OCR`) and renamed engine-vtable
+ *                 reserved_slot_4 -> `ocr_ops`, same binary offset, so the
+ *                 17-pointer tail is unchanged and only the ABI version gates
+ *                 it. Engines compiled against v10 are rejected until
+ *                 rebuilt.
  */
 #define RAC_PLUGIN_API_VERSION 11u
 


ayaangazali:
Rebased onto `488cf27b2` (0.20.35) and widened, because the gap grew while this sat.

When I opened this the define was `10u` with history up to `9u`. It is now `11u`, and neither entry was added, so the list runs 3, 4, 5, 6, 7, 8, 9 and stops two short:

```
$ sed -n '95,104p' core/include/rac/plugin/rac_plugin_entry.h
 *   9u — LLM streaming token accounting: ...
 *                 estimated. Engine-vtable reserved_slot_3 is unchanged.
 */
#define RAC_PLUGIN_API_VERSION 11u
```

This PR now adds both. Each is sourced from the promotion's own comment rather than written from memory:

- `10u`, from `rac_engine_vtable.h`: wire value 12 to `RAC_PRIMITIVE_EMBED_IMAGE`, `reserved_slot_3 -> image_embedding_ops`, same binary offset.
- `11u`, from `rac_primitive.h:58` and `rac_engine_vtable.h:200`: wire value 13 to `RAC_PRIMITIVE_OCR`, `reserved_slot_4 -> ocr_ops`, same binary offset, "so the 17-pointer tail is unchanged and only the ABI version gates it".

That last quote is the argument for this PR. The vtable layout is deliberately stable across these promotions, so the version number is the *only* thing that rejects a stale engine, and this block is the only place that explains what each version requires. Two consecutive bumps have now landed without it.

One thing I checked and want to correct on the record: I briefly thought the define had been left at `10u` while `test_rerank.cpp:184` asserted `== 11u`, which would have meant a failing test on `main`. That was my own error, I had grepped this branch's file instead of `main`'s. The define is `11u`, `test_rerank` passes on unmodified `main`, and there is no test failure here. Only the history entries are missing.

Comment-only, no code. Build and suite green on 0.20.35:

```
$ cmake --build build -j 10   # exit 0, 0 errors
100% tests passed out of 103
```

The earlier red on this PR was two separate non-issues: a DNS flake fetching protobuf at configure time, then the repo-wide Actions outage that failed every job on every PR (including #830) with zero steps run. Both are gone; this push is the first run with real signal.

ayaangazali:
Closing in favour of #849, which carries this finding along with the three other stale-reference fixes I had drip-fed as separate PRs.

Four docs PRs is four review threads for what should be one review. That was my mistake in splitting them; #849 has all four with the same evidence, re-verified on `45e91276e`. Nothing is dropped.

PR #831, opened 2026-09-02
chore: release 0.20.35 — ABI v11, OCR reachable through the SDK

sanchitmonga22:
OCR is the one modality NeuRT has a proven model for that **nothing can reach**: the detector
converts (Gate A 0.99993, 2.08× faster on ANE) and nvidia's recognizer bundles are published,
but there is no primitive, so no caller can drive them.

## Why a new primitive and not `vlm_ops`

I tested that first, because it would have avoided the ABI bump entirely. **It does not fit.**
The published recognizer is a CTC decoder over an 855-entry charset:

```
family: ocr   graphs: [recognizer]   host: charset.json
blank_token_id 0, eos_token_id 1, charset_size 855, d_model 256, 3 layers
```

No prompt, no tokenizer, no decode loop. `vlm_ops` is image + **prompt** → generated text.
Routing a line recognizer through it means handing it a prompt it cannot read and discarding
the box geometry OCR exists to produce — a contract the engine could only satisfy by lying.

## Promoted from the reserve, not grown

`RAC_PRIMITIVE_RESERVED_13` → `RAC_PRIMITIVE_OCR` (wire 13), `reserved_slot_4` → `ocr_ops` at
the same binary offset — exactly as `rerank_ops` did in v8 and `image_embedding_ops` in v10.
The **17-pointer tail is unchanged**, which is the assertion proving a promotion *renamed* a
slot rather than moving the layout under every prebuilt engine archive.

## Two verbs, because full-page OCR is two models

`recognize` reads an already-cropped line and every OCR engine must serve it. `read_page`
detects regions first, and an engine without a detector returns `NOT_SUPPORTED` — it must
**not** fall back to recognizing the whole page as one line, which returns fluent nonsense no
caller can distinguish from a real read. Regions carry a **four-corner quad**, not an
axis-aligned rect: FOTS emits rotated boxes, and flattening at the ABI throws the angle away
before anyone can use it.

## Validation

```
test_rerank        RAC_PLUGIN_API_VERSION == 11, OCR wire value 13, name "ocr"
test_engine_vtable ocr_ops immediately after image_embedding_ops;
                   reserved_slot_5 immediately after ocr_ops; tail still 17
ctest              127 passed, 0 failed   (NeuRT off — see below)
nm librac_commons  8 rac_ocr_* symbols exported
```

## ⚠️ Blocked, deliberately

This **cannot merge** until neurun ships an ABI-11 build. The guard already enforces it and
fired correctly on the first configure:

```
NeuRT prebuilt was built against RAC_PLUGIN_API_VERSION=10, but this repo is at 11.
Re-pin core/VERSIONS::NEURT_RELEASE_TAG.
```

Still to come: the NeuRT-side `g_neurt_ocr_ops` driver (CTC decode + charset + line
preprocessing, and teaching the manifest parser the `ocr` family and `recognizer` role), then
the Swift surface and catalog rows.

https://claude.ai/code/session_01R6p1R4pwTg5pvVjiEjq9ZA

## Summary by CodeRabbit

* **New Features**
  * Added OCR support for extracting text from RGB images.
  * Added page and line recognition with text, confidence scores, and region geometry.
  * Added OCR capability information and result-management APIs.
  * Integrated OCR as a routable engine feature with NeuRT support.

* **Compatibility**
  * Updated the plugin interface to version 11 to support OCR.

* **Release**
  * Updated SDK packages and documentation to version 0.20.35.


sanchitmonga22:
Corrected the contract after reading the actual model, before writing the driver against it.

I had `recognize` as the verb *"every OCR engine must serve"* — reasoning that reading one cropped line is the smaller, always-available half. **That is backwards for this family, and the driver would have proved it the expensive way.**

nvidia/nemotron-ocr's recognizer does not take a line image. Read off the published bundle:

```
in : crop     multiArrayType [1, 128, 8, 32]   fp16
out: var_703  multiArrayType [1, 1, 32, 858]   fp16
```

**128 channels** — that is a grid-sampled crop of the *detector's* feature map, not pixels. The recipe's own gold says so outright:

> "INPUTS ARE SYNTHETIC seeded randn crops — a real input is a grid-sampled crop of the detector's feature map, which needs the RegNet forward plus the C++ IndirectGridSample. So this gates EXPORT FIDELITY … **NOT end-to-end OCR accuracy on page images**."

So the recognizer cannot run standalone, and `recognize(line_image)` is unimplementable for the whole detector-coupled family. `read_page` is now the required verb; `recognize` is optional for engines whose recognizer genuinely consumes pixels (CRNN-style line models), and returns `NOT_SUPPORTED` otherwise — honest, where feeding pixels to a graph expecting 128 feature channels returns a confident string of nonsense. `supports_page_detection` inverted to `supports_line_recognition` for the same reason.

**This also resizes the remaining work.** OCR needs the detector forward *and* a grid-sample host op before one character can be read — the recognizer alone is not a shippable half. The detector is converted and gated (Gate A 0.99993); the grid-sample host op and the CTC decode are not written yet.

PR #836, opened 2026-09-03
fix(solutions): VoiceAgent knobs the loader parses are dropped by the expansion

ayaangazali:
## Description

A solution YAML can set barge-in, and the setting goes nowhere.

`config_loader.cpp:506-509` parses both knobs and puts them on `VoiceAgentConfig`:

```cpp
if (auto n = node.find("enable_barge_in"))
    cfg->set_enable_barge_in(to_bool(n->scalar));
if (auto n = node.find("barge_in_threshold_ms"))
    cfg->set_barge_in_threshold_ms(to_int(n->scalar));
```

`expand_voice_agent` then builds the `PipelineSpec` and never mentions either one. Since operator params are the only converter-to-operator channel, a host factory cannot see them. `enable_barge_in: false` in a config file is read, stored, and discarded.

The asymmetry is what makes this a bug rather than a design choice. Every other knob the loader parses for this solution is forwarded:

| parsed by config_loader | forwarded by expand_voice_agent |
|---|---|
| `sample_rate_hz` | yes, `vad` |
| `chunk_ms` | yes, `vad` |
| `system_prompt` | yes, `llm` |
| `max_context_tokens` | yes, `llm` |
| `temperature` | yes, `llm` |
| `emit_partials` | yes, `tts` |
| `enable_barge_in` | **no** |
| `barge_in_threshold_ms` | **no** |

This is not a knob with no meaning behind it either: the barge-in path is implemented in `voice_agent_feed_abi.cpp` (`kBargeInEchoMargin`, `kBargeInMinFrames`, `barge_in_armed`, per-frame arming). `solutions.proto:63` documents the flag as "Unset means enabled", so `false` is the only way to turn it off and it is exactly the value that gets lost.

Both now ride the `vad` op:

```cpp
if (cfg.has_enable_barge_in()) {
    (*vad->mutable_params())["enable_barge_in"] = cfg.enable_barge_in() ? "true" : "false";
}
if (cfg.barge_in_threshold_ms() > 0) {
    (*vad->mutable_params())["barge_in_threshold_ms"] = std::to_string(cfg.barge_in_threshold_ms());
}
```

`has_enable_barge_in()` rather than the value, because the field is `optional` and the documented default is enabled, so reading the value would turn "unset" into an explicit `false` and silently disable barge-in for everyone.

**Two judgement calls I want to flag rather than bury.**

First, `vad` and not `tts`. Barge-in is detected from user speech arriving during playback, and the existing implementation does that arming in the VAD/feed path, so the threshold is a VAD-domain value and it sits naturally beside `sample_rate_hz`/`chunk_ms`. If you would rather both landed on `tts` next to `emit_partials`, that is a one-line move and I will make it.

Second, there is no in-tree consumer to point at, and that is expected here rather than a gap in this PR. The header of this file says these expansions "intentionally reference engine-backed operator type names without installing factories for them", so operators are host-supplied and interpret params themselves. `emit_partials` has no in-tree reader either, for the same reason. This PR makes the two knobs reachable on the same terms as every other one.

I did not touch `audio_source` / `audio_file_path`, which the loader also parses without forwarding. Those describe how the host feeds audio into the pipeline rather than how an operator behaves, so they are a separate question and not obviously the same defect.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring

## Testing
- [ ] Lint passes locally

Added `voice_agent_barge_in_params_reach_the_vad_operator` to `core/tests/test_solution_runner.cpp`, beside the existing `voice_agent_solution_compiles`. It sets `enable_barge_in = false` (the value that cannot survive a value-read) plus a threshold, then asserts both land on the `vad` operator's params.

```
build exit=0, 0 errors
100% tests passed out of 101
```

Fails on the unfixed code. Verified by removing only the forwarding block and confirming the relink happened first:

```
[100%] Linking CXX executable test_solution_runner
[FAIL] core/tests/test_solution_runner.cpp:1101 enabled != params.end()
```

On lint: unchecked because `core/scripts/lint-cpp.sh` needs the repo's pinned `clang-format` and only Apple `clang-format 21` is available here. Both files this PR touches are completely clean under it (zero diff hunks), so there is no formatting drift either way.

The edited block is inside `#if defined(RAC_HAVE_PROTOBUF)`; protobuf-off builds do not compile it.

## Labels
**SDKs:**

## Checklist
- [x] Self-review completed

## Summary by CodeRabbit

* **New Features**
  * Voice agent configurations now support barge-in controls, including enablement and response threshold settings.
  * Reasoning output preferences are now honored, allowing configured reasoning content to be included in generated responses.

* **Bug Fixes**
  * Explicitly disabled barge-in settings are now preserved when configuring voice agents.

### DIFF
--- a/core/src/solutions/solution_converter.cpp
+++ b/core/src/solutions/solution_converter.cpp
@@ -81,6 +81,15 @@ void expand_voice_agent(const runanywhere::v1::VoiceAgentConfig& cfg, PipelineSp
         (*vad->mutable_params())["barge_in_threshold_ms"] =
             std::to_string(cfg.barge_in_threshold_ms());
     }
+    // `emit_thoughts` in the YAML lands on generation.reasoning.include_in_output
+    // (config_loader.cpp), and only system_prompt and temperature were being
+    // read back off `generation`, so it stopped at the proto like the barge-in
+    // pair. Guarded on has_reasoning() rather than the bool, which is a plain
+    // proto3 field with no presence of its own.
+    if (cfg.has_generation() && cfg.generation().has_reasoning()) {
+        (*llm->mutable_params())["emit_thoughts"] =
+            cfg.generation().reasoning().include_in_output() ? "true" : "false";
+    }
 
     add_edge(out, "vad.out", "stt.in");
     add_edge(out, "stt.final", "llm.in");
--- a/core/src/solutions/solution_converter.cpp
+++ b/core/src/solutions/solution_converter.cpp
@@ -69,6 +69,18 @@ void expand_voice_agent(const runanywhere::v1::VoiceAgentConfig& cfg, PipelineSp
         (*llm->mutable_params())["temperature"] = std::to_string(cfg.generation().temperature());
     }
     (*tts->mutable_params())["emit_partials"] = cfg.emit_partials() ? "true" : "false";
+    // Barge-in is detected on the VAD side (user speech arriving while the
+    // assistant is talking), which is why these ride the vad op alongside
+    // sample_rate_hz/chunk_ms rather than tts. config_loader.cpp parses both
+    // out of the solution YAML, so without forwarding them here a caller's
+    // `enable_barge_in: false` reaches the proto and then stops.
+    if (cfg.has_enable_barge_in()) {
+        (*vad->mutable_params())["enable_barge_in"] = cfg.enable_barge_in() ? "true" : "false";
+    }
+    if (cfg.barge_in_threshold_ms() > 0) {
+        (*vad->mutable_params())["barge_in_threshold_ms"] =
+            std::to_string(cfg.barge_in_threshold_ms());
+    }
 
     add_edge(out, "vad.out", "stt.in");
     add_edge(out, "stt.final", "llm.in");
--- a/core/tests/test_solution_runner.cpp
+++ b/core/tests/test_solution_runner.cpp
@@ -1085,6 +1085,7 @@ TEST(voice_agent_barge_in_params_reach_the_vad_operator) {
     va->set_vad_model_id("silero");
     va->set_enable_barge_in(false);
     va->set_barge_in_threshold_ms(250);
+    va->mutable_generation()->mutable_reasoning()->set_include_in_output(true);
 
     SolutionRunner runner(cfg);
     CHECK(runner.start() == RAC_SUCCESS);
@@ -1103,6 +1104,18 @@ TEST(voice_agent_barge_in_params_reach_the_vad_operator) {
     const auto threshold = params.find("barge_in_threshold_ms");
     CHECK(threshold != params.end());
     CHECK(threshold->second == "250");
+
+    // emit_thoughts rides `generation.reasoning`, which was the other knob the
+    // loader parsed and the expansion dropped.
+    const runanywhere::v1::OperatorSpec* llm = nullptr;
+    for (const auto& op : spec.operators()) {
+        if (op.name() == "llm")
+            llm = &op;
+    }
+    CHECK(llm != nullptr);
+    const auto thoughts = llm->params().find("emit_thoughts");
+    CHECK(thoughts != llm->params().end());
+    CHECK(thoughts->second == "true");
 
     runner.close_input();
     runner.wait();
--- a/core/tests/test_solution_runner.cpp
+++ b/core/tests/test_solution_runner.cpp
@@ -1067,6 +1067,48 @@ TEST(voice_agent_solution_compiles) {
 }
 
 // ---------------------------------------------------------------------------
+// 9b. VoiceAgent barge-in knobs reach the graph.
+//     config_loader.cpp parses enable_barge_in / barge_in_threshold_ms out of
+//     the solution YAML, so the expansion has to carry them onto an operator
+//     or a caller's setting stops at the proto. Explicit false is the case
+//     that matters: it is the only way to turn barge-in off, and it is
+//     indistinguishable from "unset" unless presence is honoured.
+// ---------------------------------------------------------------------------
+TEST(voice_agent_barge_in_params_reach_the_vad_operator) {
+    ScopedSolutionStandins standins;
+
+    SolutionConfig cfg;
+    auto* va = cfg.mutable_voice_agent();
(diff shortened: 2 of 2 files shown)


ayaangazali:
`kotlin-android` was flaky, not this diff. It is green now on an identical tree.

I re-ran it by amending the commit with no content change (`commit --amend --no-edit`), so the tree is byte-identical and only the SHA moved. Same code, opposite result:

```
before: kotlin-android  COMPLETED FAILURE
after:  kotlin-android  COMPLETED SUCCESS
```

Why I did not go looking for a fix first, beyond the re-run passing:

- **The C++ this PR changes did build in that failing run.** The job's own `Build Android JNI libs` step succeeded; what failed was the later `Kotlin SDK unit tests + debug assembly` step.
- **This diff cannot reach those tests.** It touches `core/src/solutions/solution_converter.cpp` and a C++ test, inside `#if defined(RAC_HAVE_PROTOBUF)`, reachable only through the solutions C ABI. `testDebugUnitTest` is a JVM unit run.
- **No Kotlin test asserts on any of it.** Nothing under `bindings/kotlin/src/test` references `barge_in` or operator params.
- **It was one run in thirty-seven.** `kotlin-android` was SUCCESS on my other 36 open PRs against the same base and FAILURE only here, which is what a low-rate flake looks like spread across independent runs rather than a regression.

I could not read the failing job's log to name the test: `gh run view --log`, the job logs endpoint, and the redirect all returned empty from here, and a local `testDebugUnitTest` stopped earlier at `downloadJniLibs` on my machine. So I am resting this on the re-run plus the reachability argument rather than claiming I identified the specific test.

The PR is otherwise unchanged: 25 checks passing, 3 pending, no failures.

ayaangazali:
Widened this by one field rather than opening a second PR for it.

After filing this I did the comparison properly, every field `config_loader.cpp` parses against every field each `expand_*` forwards, and `emit_thoughts` is the same defect in the same function. It lands on `generation.reasoning.include_in_output` (`config_loader.cpp:518`), and the expansion only ever read `system_prompt` and `temperature` back off `generation`, so it stopped at the proto exactly like the barge-in pair.

Splitting that into its own PR would have been drip-feeding one concern across two reviews, so it is here instead. The concern this PR now states is "knobs the loader parses that the expansion drops", which is the same shape as before, one field wider. Retitled to match.

Guarded on `has_reasoning()` rather than on the bool: `include_in_output` is a plain proto3 field with no presence of its own, so there is nothing to test on it directly. I had written `has_include_in_output()` first and it does not compile, which is how I noticed.

```cpp
if (cfg.has_generation() && cfg.has_generation() && cfg.generation().has_reasoning()) {
    (*llm->mutable_params())["emit_thoughts"] =
        cfg.generation().reasoning().include_in_output() ? "true" : "false";
}
```

The existing test now covers it too, and I revert-checked this field on its own, removing only the `emit_thoughts` block and leaving the barge-in fix in place:

```
[100%] Linking CXX executable test_solution_runner
[FAIL] core/tests/test_solution_runner.cpp:1117 thoughts != llm->params().end()
```

Full suite still 101/101.

Two things I checked and deliberately left out:

- `audio_source` / `audio_file_path` are also parsed without being forwarded, but they describe how the host feeds audio into the pipeline rather than how an operator behaves, so I do not think they are the same defect. Happy to be told otherwise.
- `agent_loop` and `time_series` have no `populate_*` in the loader at all, so they cannot be configured from YAML and have nothing to drop.

The same sweep did turn up four dropped fields in `expand_rag`, including `vector_store` while `vector_store_path` was forwarded. That is a different solution and a different set of consumers, so it is #839 rather than more scope here.

ayaangazali:
Pushed d28bd6e7 to remove a conflict this PR had with #812. No behaviour change, and the test is unchanged.

I now have eight open PRs touching `core/src/features/rag` and `core/src/solutions`, so I test-merged all eight onto `main` together rather than assuming pairwise cleanliness. Seven merged; this one conflicted:

```
CONFLICT (content): Merge conflict in core/tests/test_solution_runner.cpp
```

The cause was mine and entirely mechanical: this PR and #812 both added a test numbered `9b` at the same insertion point, immediately before `// 10. SolutionConfig (RAG)`. Nothing about the tests overlaps, they just landed on the same line.

Moved this one to its own slot, `11b`, before `// 12. C ABI end-to-end: YAML path`, and gave its `run_test_...()` registration a different neighbour. Re-ran the same eight-way probe afterwards:

```
merged OK: #742 #823 #824 #812 #836 #839 #844 #845
conflicts: NONE
build exit=0, ctest 103/103
```

So the eight can now land in any order without anyone resolving a merge for me. Flagging it here rather than silently repushing, since the branch moved.

sanchitmonga22:
> Replying to [this comment](https://github.com/RunanywhereAI/runanywhere-sdks/pull/836#issuecomment-5528864539)

That makes sense — thanks for explaining. `audio_source`/`audio_file_path` describe how the host feeds audio into the pipeline, not operator runtime behavior, so treating that as a different class of gap from the presence-tracked knobs this PR fixes holds up. Treating this as addressed.

sanchitmonga22 (approved):
Thanks a lot for this, @ayaangazali! This fixes `expand_voice_agent` silently dropping `enable_barge_in`, `barge_in_threshold_ms` and `emit_thoughts` at the proto boundary — they now reach the vad/llm operator params, with a new test locking in the presence-vs-default semantics.

Worth flagging for later, not holding this up: no in-tree operator reads these params yet (`DetectVoiceNode`/`GenerateTextNode` only look at their existing keys), so end-to-end behavior doesn't change until that catches up — opened #910 for the barge-in side and #911 for the wider gap between what expand_voice_agent writes and what operators read. Small optional add for whenever it's convenient: `GenerateTextNode` could read `emit_thoughts` directly in a follow-up — no need to hold this PR on it.

Checked: CodeRabbit reviewed the latest commit · CI green · built and linted locally merged into `main` (C++ commons and engines (core/**)) · two independent code reviews.

Follow-ups, not blocking: #910 (wire an operator to honor the barge-in params this PR forwards), #911 (audit every key expand_* writes against what operators actually read) — you're welcome to pick these up.

PR #841, opened 2026-09-04
fix(embeddings): an explicit normalize=false never reaches the engine

ayaangazali:
## Description

`rac_embeddings_options_from_proto` cannot express "do not normalize", so a caller asking for raw pooled vectors gets L2-normalized ones instead.

The adapter hardcodes the mode:

```cpp
// EmbeddingsOptions.normalize is a plain proto3 bool with no unset sentinel;
// the backends default to L2 unit vectors ...
// Disabling normalization is a component-configuration concern.
out->normalize = RAC_EMBEDDINGS_NORMALIZE_L2;
```

The premise in that comment is not true of the current schema. `normalize` is `optional`, so it is presence-tracked and `has_normalize()` separates "caller said false" from "caller said nothing":

```proto
// L2-normalize every vector to unit length (what cosine search expects).
// Unset = true. false returns the raw pooled vector.
optional bool normalize = 4 [(runanywhere.v1.rac_default) = "true"];
```

"false returns the raw pooled vector" is a documented behaviour with no way to reach it: every `EmbeddingsOptions` that crosses this adapter comes out as L2. The C ABI is not the limitation either, since the enum already carries the value the adapter never emits:

```c
RAC_EMBEDDINGS_NORMALIZE_NONE = 0, /**< No normalization */
```

The fix reads presence and leaves the default alone:

```cpp
out->normalize = (!in.has_normalize() || in.normalize()) ? RAC_EMBEDDINGS_NORMALIZE_L2
                                                         : RAC_EMBEDDINGS_NORMALIZE_NONE;
```

Unset stays L2, which is both `RAC_EMBEDDINGS_OPTIONS_DEFAULT.normalize` (already assigned two lines above) and the documented "Unset = true", so nothing changes for callers that do not set the field. Only an explicit `false` behaves differently, which is the point.

This is the same shape as #828: a value read justified by a comment describing a schema that has since changed, with the neighbouring line in this very function already doing it correctly (`out->truncate = in.has_truncate() ? ... : -1`).

I have not touched the second half of that comment, that disabling normalization is "a component-configuration concern". If the intent is genuinely that this knob should only exist at component-configuration level, then the field and its documentation are the things to change, and I would rather you make that call than have me delete a per-call option that reads as supported.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring

## Testing
- [ ] Lint passes locally

Extended `test_embeddings_options_mapping` in `core/tests/test_advanced_modality_proto_abi.cpp` with both explicit cases. The two assertions already there never set `normalize`, so they only ever covered the unset path and still pass unchanged.

```
build exit=0
100% tests passed out of 101
```

Fails on the unfixed code. Verified by restoring the hardcoded assignment, with the relink confirmed first:

```
[100%] Linking CXX executable test_advanced_modality_proto_abi
FAIL: an explicit normalize=false reaches the C ABI as NONE
      (...test_advanced_modality_proto_abi.cpp:934) raw.normalize == RAC_EMBEDDINGS_NORMALIZE_NONE
```

On lint: unchecked because `core/scripts/lint-cpp.sh` wants the repo's pinned `clang-format` and only Apple `clang-format 21` is here. `rac_proto_adapters.cpp` is fully clean under it, including my continuation-line alignment, which I corrected after it flagged a one-space difference.

The edited function is inside `#if defined(RAC_HAVE_PROTOBUF)`; protobuf-off builds do not compile it.

## Labels
**SDKs:**

## Checklist
- [x] Self-review completed

## Summary by CodeRabbit

* **Bug Fixes**
  * Embedding options now respect an explicitly disabled normalization setting, returning raw pooled vectors when selected.
  * Default and enabled normalization behavior continues to use L2 normalization.

* **Tests**
  * Added coverage for enabled, disabled, and default embedding normalization settings.

### DIFF
--- a/core/src/foundation/rac_proto_adapters.cpp
+++ b/core/src/foundation/rac_proto_adapters.cpp
@@ -849,11 +849,13 @@ bool rac_embeddings_options_from_proto(const ::runanywhere::v1::EmbeddingsOption
         return false;
     *out = RAC_EMBEDDINGS_OPTIONS_DEFAULT;
 
-    // EmbeddingsOptions.normalize is a plain proto3 bool with no unset sentinel;
-    // the backends default to L2 unit vectors (EmbeddingsConfiguration.normalize
-    // rac_default = true), so an explicit true and the proto default both resolve
-    // to L2 here. Disabling normalization is a component-configuration concern.
-    out->normalize = RAC_EMBEDDINGS_NORMALIZE_L2;
+    // EmbeddingsOptions.normalize is `optional` with rac_default "true", so
+    // presence separates "caller said false" from "caller said nothing"
+    // (embeddings_options.proto: "Unset = true. false returns the raw pooled
+    // vector"). Unset keeps the L2 that RAC_EMBEDDINGS_OPTIONS_DEFAULT already
+    // set; an explicit false is the only way to ask for the raw vector.
+    out->normalize = (!in.has_normalize() || in.normalize()) ? RAC_EMBEDDINGS_NORMALIZE_L2
+                                                             : RAC_EMBEDDINGS_NORMALIZE_NONE;
 
     switch (in.pooling()) {
         case ::runanywhere::v1::EMBEDDINGS_POOLING_STRATEGY_UNSPECIFIED:
--- a/core/tests/test_advanced_modality_proto_abi.cpp
+++ b/core/tests/test_advanced_modality_proto_abi.cpp
@@ -922,6 +922,25 @@ int test_embeddings_options_mapping() {
     CHECK(raw.truncate == 0, "explicit truncate=false maps exactly");
     CHECK(raw.batch_size == 32, "embedding batch size maps exactly");
 
+    // normalize is `optional` with rac_default "true": "Unset = true. false
+    // returns the raw pooled vector" (embeddings_options.proto). Explicit false
+    // is the only way to ask for the raw vector, so it has to survive the hop.
+    runanywhere::v1::EmbeddingsOptions no_normalize;
+    no_normalize.set_normalize(false);
+    raw = RAC_EMBEDDINGS_OPTIONS_DEFAULT;
+    CHECK(rac::foundation::rac_embeddings_options_from_proto(no_normalize, &raw),
+          "EmbeddingsOptions with normalize=false maps");
+    CHECK(raw.normalize == RAC_EMBEDDINGS_NORMALIZE_NONE,
+          "an explicit normalize=false reaches the C ABI as NONE");
+
+    runanywhere::v1::EmbeddingsOptions yes_normalize;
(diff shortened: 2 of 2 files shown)


ayaangazali:
Found supporting evidence for this while sweeping the bindings, and it upgrades the impact from "the ABI contradicts its own default" to "two shipping bindings already send this and it is being dropped".

Both TypeScript bindings put `normalize` on the wire today.

Electron, in `bindings/electron/src/api/data-abi.ts`, with a comment that describes precisely the symptom this PR fixes:

> `normalize` is the field that used to be lost: the public option existed and the component ABI had nowhere to put it, so **L2 and NONE produced the same vectors**. It is OPTIONAL on the wire, so unset is left absent and commons applies its own `rac_default`.

```ts
normalize: options.normalize ? options.normalize === NormalizeMode.L2 : undefined,
```

Web does the same in `Public/API/Mapping.ts:474`:

```ts
normalize: options?.normalize === undefined ? defaults.normalize : options.normalize === 'l2',
```

Both are correct: they leave the field absent when the caller says nothing, and send an explicit `false` for `'none'` / `NormalizeMode.NONE`. That is exactly the input `rac_embeddings_options_from_proto` was discarding, because it assigned `RAC_EMBEDDINGS_NORMALIZE_L2` unconditionally.

So an Electron or Web caller asking for unnormalized embeddings has been getting L2 vectors, and the binding layer is not where it is lost. That binding comment reads as someone hitting this symptom, fixing their half, and reasonably assuming commons honoured the wire value.

Nothing changed in the diff; I am adding this because it is the difference between a latent contract bug and one two SDKs are hitting.

sanchitmonga22:
> Replying to [this comment](https://github.com/RunanywhereAI/runanywhere-sdks/pull/841#issuecomment-5555767347)

That makes sense -- thanks for checking the actual binding sources. Confirmed: `bindings/electron/src/api/data-abi.ts:58` and `bindings/web/packages/core/src/Public/API/Mapping.ts:474-476` do send an explicit `normalize=false` for `NormalizeMode.NONE` today, and this PR's fix is exactly what makes that reach the engine. Treating this as addressed.

sanchitmonga22 (approved):
Thanks a lot for this, @ayaangazali! An explicit `normalize=false` from any SDK was being silently overwritten with L2 in the commons proto adapter -- this fixes it so callers actually get the raw pooled vector the proto documents, which the Electron and Web bindings are already sending on the wire today.

Checked: CodeRabbit reviewed the latest commit · CI green · built and linted locally merged into `main` (C++ commons core, proto adapter + tests) · two independent code reviews.

Follow-ups, not blocking: #918 (ONNX/QHexRT still ignore `normalize=false`), #919 (Python SDK can't request it) -- you're welcome to pick these up.

PR #844, opened 2026-09-06
fix(solutions): an apostrophe in a YAML scalar swallows the trailing comment

ayaangazali:
## Description

A solution YAML that contains an apostrophe in prose silently absorbs the trailing comment into the value.

```yaml
voice_agent:
  system_prompt: Don't use markdown  # keep replies short
```

parses `system_prompt` as:

```
Don't use markdown  # keep replies short
```

The comment stripper in `config_loader.cpp` tracks quote state so that a `#` inside a quoted string is not treated as a comment, which is correct and worth keeping:

```cpp
if (!in_dq && c == '\'') { in_sq = !in_sq; clean.push_back(c); continue; }
...
if (!in_sq && !in_dq && c == '#') break;
```

The problem is that an apostrophe in ordinary prose is indistinguishable from an opening quote. `Don't` flips `in_sq` to true and nothing ever flips it back, so every remaining character on the line counts as quoted and the `#` is never reached.

`system_prompt` is exactly where this bites, because it is the one field in these configs that holds English prose, and "don't", "user's" and "it's" are natural things to write in a voice-agent instruction. The value then reaches the model with a stray `# keep replies short` appended to it.

The fix keeps the quote tracking and adds a fallback for the case that proves it was wrong. An unterminated quote at end of line means the tracking mis-parsed that line, so it re-scans using YAML's own rule that an inline comment is a `#` preceded by whitespace:

```cpp
if (in_sq || in_dq) {
    clean.clear();
    for (size_t i = 0; i < line.size(); ++i) {
        if (line[i] == '#' && (i == 0 || line[i - 1] == ' ' || line[i - 1] == '\t'))
            break;
        clean.push_back(line[i]);
    }
}
```

That leaves the balanced cases alone: `"has # inside"` still keeps its `#`, and a `#` with no leading whitespace (`a#b`) is still not a comment.

I considered dropping the quote tracking entirely and always cutting at a whitespace-preceded `#`, which is simpler. I did not, because it would break `system_prompt: "explain the # operator"`, a case the current code handles correctly. This keeps that working and only changes lines the old logic got wrong.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring

## Testing
- [ ] Lint passes locally

Added `yaml_apostrophe_does_not_swallow_a_trailing_comment` to `core/tests/test_solution_runner.cpp`, driving the real parser through the public `load_solution_from_yaml`. It asserts both directions: the apostrophe line parses as `Don't use markdown`, and a properly quoted `"has # inside"` keeps its `#`.

```
build exit=0, 0 errors
100% tests passed out of 101
```

Fails on the unfixed code. Verified by removing only the fallback block, confirming `config_loader.cpp` recompiled and the test binary relinked first:

```
[ 69%] Building CXX object core/CMakeFiles/rac_commons.dir/src/solutions/config_loader.cpp.o
[ 98%] Linking CXX executable test_solution_runner
[FAIL] core/tests/test_solution_runner.cpp:1165 prompt == "Don't use markdown"
[yaml] system_prompt = Don't use markdown  # keep replies short
```

That last line is the bug printed verbatim. The quoted-`#` assertion kept passing under the revert, so the two halves are independent.

On lint: unchecked because `core/scripts/lint-cpp.sh` needs the repo's pinned `clang-format` and only Apple `clang-format 21` is available here. `test_solution_runner.cpp` is entirely clean under it; `config_loader.cpp` has one pre-existing hunk at line 534 (the `emit_thoughts` line), well outside my 125-139 range.

The edited code is not inside any `#if`, so it compiles in every configuration.

## Labels
**SDKs:**

## Checklist
- [x] Self-review completed

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed YAML parsing for unquoted text containing apostrophes, ensuring trailing comments are handled correctly.
  - Preserved `#` characters within tokens and quoted values while continuing to remove actual trailing comments.
  - Improved handling of comments following text that includes apostrophes.

- **Tests**
  - Added coverage for apostrophes in ordinary YAML text and for correctly quoted values containing `#`.
  - Verified comment handling across multiple quoted and unquoted value formats.

### DIFF
--- a/core/src/solutions/config_loader.cpp
+++ b/core/src/solutions/config_loader.cpp
@@ -107,14 +107,26 @@ class YamlParser {
             // keeps us honest for strings containing '#'.
             std::string clean;
             bool in_sq = false, in_dq = false;
-            for (char c : line) {
+            for (size_t i = 0; i < line.size(); ++i) {
+                const char c = line[i];
+                // A quote can only OPEN where a YAML token can start: begin of
+                // line, or after whitespace, ':' or '-'. It still CLOSES
+                // anywhere. Without this, the apostrophe in `Don't` opened a
+                // quote, and a second apostrophe later on the line (commonly in
+                // the comment, `# don't forget`) closed it again, leaving the
+                // tracking balanced and wrong so the '#' was never seen.
+                const char prev = i == 0 ? '\0' : line[i - 1];
+                const bool can_open =
+                    i == 0 || prev == ' ' || prev == '\t' || prev == ':' || prev == '-';
                 if (!in_dq && c == '\'') {
-                    in_sq = !in_sq;
+                    if (in_sq || can_open)
+                        in_sq = !in_sq;
                     clean.push_back(c);
                     continue;
                 }
                 if (!in_sq && c == '"') {
-                    in_dq = !in_dq;
+                    if (in_dq || can_open)
+                        in_dq = !in_dq;
                     clean.push_back(c);
                     continue;
                 }
--- a/core/src/solutions/config_loader.cpp
+++ b/core/src/solutions/config_loader.cpp
@@ -122,6 +122,21 @@ class YamlParser {
                     break;
                 clean.push_back(c);
             }
+            // An apostrophe in ordinary prose (`Don't use markdown`) opens a
+            // quote state nothing closes, so the loop above treats the rest of
+            // the line as quoted and keeps the trailing comment as part of the
+            // value. An unterminated quote means the tracking was wrong for
+            // this line, so fall back to YAML's own rule: an inline comment is
+            // a '#' preceded by whitespace. That still leaves `a#b` alone and
+            // still protects a balanced "has # inside".
+            if (in_sq || in_dq) {
+                clean.clear();
+                for (size_t i = 0; i < line.size(); ++i) {
+                    if (line[i] == '#' && (i == 0 || line[i - 1] == ' ' || line[i - 1] == '\t'))
+                        break;
+                    clean.push_back(line[i]);
+                }
+            }
             // Trim trailing WS.
             while (!clean.empty() &&
                    (clean.back() == ' ' || clean.back() == '\t' || clean.back() == '\r')) {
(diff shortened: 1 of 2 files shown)


sanchitmonga22:
Thanks for this, @ayaangazali! Fixing the comment-stripper so a plain `system_prompt` like `Don't use markdown` no longer picks up the trailing comment is a real improvement.

1. `core/src/solutions/config_loader.cpp:132` -- the fallback only runs when a quote is still open at end of line, so the same bug survives whenever the value *and* the comment each contain an apostrophe, e.g. `system_prompt: Don't use markdown  # don't forget` -- the two apostrophes balance out, `in_sq`/`in_dq` are both false again by end of line, and the comment still gets appended. Letting `'`/`"` open a quote only at a token boundary (start of line, or right after whitespace, `:` or `-`), while still closing anywhere, should close this gap -- and it'd be great to add that two-apostrophe case to the new test alongside the ones already there.

ayaangazali:
Caught me. Fixed in the push above, using the rule you suggested.

You are right that the fallback was the wrong shape: it only ran when a quote was still open at end of line, so two apostrophes cancelled out and the original bug walked straight through. Confirmed your exact case before changing anything, with the parser reverted to what is on this branch today:

```
system_prompt: Don't use markdown  # don't forget
  ->  "Don't use markdown  # don't forget"
```

A quote now opens only where a YAML token can start (begin of line, or after whitespace, `:` or `-`) and still closes anywhere, so the apostrophe in `Don't` never opens one and the `#` is seen normally. I kept the unterminated-quote fallback rather than deleting it, since it still does real work for a genuinely unterminated quote like `system_prompt: 'oops  # x`.

Added the two-apostrophe case to the test as you asked, and one more going the other way: a properly quoted scalar whose *comment* also contains an apostrophe (`system_prompt: "has # inside"  # don't forget`), so the fix cannot be "stop tracking quotes" and the `#` inside the quotes still survives.

Verified the test actually catches the regression rather than just passing: reverted only `config_loader.cpp`, confirmed the binary relinked, and the new assertion fails with exactly the wrong value above. Restored the fix and it passes. Full run 103/103.

PR #845, opened 2026-09-06
fix(rag): rerank snippets are cut mid-UTF-8 character

ayaangazali:
## Description

`rerank_llm_pointwise` builds the scorer prompt from truncated passages, and the truncation stops at a byte count:

```cpp
constexpr size_t kMaxChunkChars = 360;

std::string flatten_and_truncate(const std::string& text, size_t max_chars) {
    std::string out;
    for (char c : text) {
        if (out.size() >= max_chars)   // bytes, not characters
            break;
        out.push_back((c == '\n' || c == '\r' || c == '\t') ? ' ' : c);
    }
    return out;
}
```

So a passage in any non-Latin script is cut mid-sequence and the snippet handed to the model ends in an incomplete character. Running that exact function over realistic inputs:

```
                bytes  codepoints  valid_utf8
ASCII             360         360     yes
CJK (aligned)     360         120     yes
CJK (+1 byte)     360         121     NO
Russian           360         191     NO
emoji             360         292     NO
```

The aligned CJK row is the trap: with a passage whose length happens to be a multiple of 3 the cut lands on a boundary and everything looks fine, so this does not reproduce reliably unless you shift the input by a byte.

This is the same defect the codebase already fixed one directory over, and the comment there explains why it was worth fixing:

```cpp
// core/src/features/rag/rag_backend.cpp:307
// Truncate the source preview at a UTF-8 character boundary, not a raw
// byte offset: a mid-sequence cut emits invalid UTF-8, which strict proto
// decoders (e.g. SwiftProtobuf) reject ...
```

This one feeds a prompt rather than a proto field, so the consequence is milder: the scorer sees a mangled trailing character rather than a decode failure. It is still worth correcting, because it is a per-passage corruption that only affects non-English content and the fix is the same three lines.

The fix mirrors `rag_backend.cpp` exactly, walking back off continuation bytes. The whitespace substitution above is one byte for one byte, so offsets into `out` and `text` still line up.

**One thing I deliberately did not change.** `kMaxChunkChars` is named for characters but counts bytes, so a CJK passage gets 120 characters of context where an English one gets 360. That is arguably the bigger quality issue for non-English RAG, but fixing it would roughly triple the prompt size for those languages, which is a context-budget and cost decision rather than a bug fix. Flagging it here rather than making that call unasked.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring

## Testing
- [ ] Lint passes locally

No test added. `flatten_and_truncate` is a file-local helper in an anonymous namespace, and the only exported entry point that reaches it, `rerank_llm_pointwise`, needs a live LLM handle. Making it testable would mean widening the module's surface to suit the test, which is a worse trade than the fix is worth. The existing `test_rag_rerank.cpp` deliberately covers `parse_rerank_scores` and `reorder_by_scores` "directly, without a" model, and this helper does not fit that shape.

Instead I verified against a standalone copy of the exact function, before and after:

```
before:  Russian bytes=360 valid_utf8=NO    emoji bytes=360 valid_utf8=NO
after:   Russian bytes=359 valid_utf8=yes   emoji bytes=357 valid_utf8=yes
         ASCII   bytes=360 valid_utf8=yes   (unchanged)
```

Full suite green on the real build:

```
$ cmake --build build -j 10 && ctest --test-dir build -j 10
100% tests passed out of 101
```

On lint: unchecked because `core/scripts/lint-cpp.sh` needs the repo's pinned `clang-format` and only Apple `clang-format 21` is available here. Under it this file is completely clean, including the line I added, which I reformatted after it flagged my initial wrapping.

The edited code is not inside any `#if`, so it compiles in every configuration.

## Labels
**SDKs:**

## Checklist
- [x] Self-review completed

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented truncated text from ending with incomplete UTF-8 characters.
  - Preserved the existing byte-size limit when shortening text.

### DIFF
--- a/core/src/features/rag/rag_rerank.cpp
+++ b/core/src/features/rag/rag_rerank.cpp
@@ -30,6 +30,17 @@ std::string flatten_and_truncate(const std::string& text, size_t max_chars) {
             break;
         out.push_back((c == '\n' || c == '\r' || c == '\t') ? ' ' : c);
     }
+    // Back off to a UTF-8 character boundary. The budget is a byte count, so
+    // a passage in any non-Latin script is cut mid-sequence and the snippet
+    // handed to the scorer ends in an incomplete character. The whitespace
+    // substitution above is 1 byte for 1 byte, so offsets into `out` and
+    // `text` still line up. Same walk as rag_backend.cpp's source preview.
+    size_t cut = out.size();
+    while (cut > 0 && cut < text.size() &&
+           (static_cast<unsigned char>(text[cut]) & 0xC0) == 0x80) {
+        --cut;
+    }
+    out.resize(cut);
     return out;
 }
 
--- a/core/src/features/rag/rag_rerank.cpp
+++ b/core/src/features/rag/rag_rerank.cpp
@@ -36,8 +36,7 @@ std::string flatten_and_truncate(const std::string& text, size_t max_chars) {
     // substitution above is 1 byte for 1 byte, so offsets into `out` and
     // `text` still line up. Same walk as rag_backend.cpp's source preview.
     size_t cut = out.size();
-    while (cut > 0 && cut < text.size() &&
-           (static_cast<unsigned char>(text[cut]) & 0xC0) == 0x80) {
+    while (cut > 0 && cut < text.size() && (static_cast<unsigned char>(text[cut]) & 0xC0) == 0x80) {
         --cut;
     }
     out.resize(cut);


ayaangazali:
Correcting the record on one line of this PR's description.

The body said the file was clang-format clean "including the line I added, which I reformatted after it flagged my initial wrapping". That was true of my working tree but not of what I had actually pushed: I fixed the wrapping and then ran `commit --amend` without staging the change, so the amend picked up nothing and the branch kept the wrapped version.

Fixed in d9d9fe3a. The condition is now on one line as clang-format wants it, and the pushed tree matches what I checked locally:

```cpp
while (cut > 0 && cut < text.size() && (static_cast<unsigned char>(text[cut]) & 0xC0) == 0x80) {
```

Suite still 101/101 after the change. Nothing else in the diff moved.

sanchitmonga22 (approved):
Thanks a lot for this, @ayaangazali! Rerank passages in non-Latin scripts were being cut off mid-UTF-8-character before reaching the scorer -- this backs the cut off to the last complete character, matching the idiom already used for source previews.

Checked: CodeRabbit reviewed the latest commit (d9d9fe3a) with zero actionable comments · CI green across all workflows · built and tested locally at the merge base (core C++ commons -- core/src/features/rag/**, 104/104 tests) · two independent code reviews, both traced the UTF-8 boundary walk by hand across the multi-byte edge cases.

Follow-ups, not blocking: #920 (UTF-8-safe splitting in the document chunker) -- you're welcome to pick it up.

PR #848, opened 2026-09-08
fix(vlm): streamed VLM chunks are invalid UTF-8 for any character the vocab splits

ayaangazali:
## Description

`vlm_stream_token_callback` forwards each backend piece straight to the caller:

```cpp
// core/src/features/vlm/vlm_module.cpp:573 on main
if (!cleaned.empty() && ctx->token_callback) {
    return ctx->token_callback(cleaned.c_str(), ctx->user_data);
}
```

A tokenizer piece is not guaranteed to be a whole character, so some of those chunks are not valid UTF-8.

**Why it is wrong.** Byte-level BPE vocabularies carry single-byte fallback tokens, so any character without a whole-character token must be emitted as several pieces. Measured against the real Qwen2.5 vocabulary (`tokenizer.json`, 151643 tokens):

- 1448 tokens are not valid UTF-8 on their own
- all 64 bare continuation bytes `0x80`-`0xBF` are present as single-byte tokens

Whether a given character splits is decidable without reference to merge order, by asking whether the whole character exists as one token:

| character | whole-character token in vocab | consequence |
|---|---|---|
| U+1FAE0 melting face | no | must split mid-character |
| U+1F979 | no | must split mid-character |
| U+9F9C (rare CJK) | no | must split mid-character |
| U+1F600, U+65E5 | yes | safe |

That last row is why this has gone unnoticed. The characters used in casual testing have whole-character tokens; the ones that do not are the failure cases.

**What the caller sees.** The C callback is consumed directly by `bindings/python/native/module.cpp:917`, whose `stream_token_cb` hands the `std::string` to pybind11 for a strict UTF-8 decode:

```
>>> b'\xf0\x9f'.decode('utf-8')
UnicodeDecodeError: 'utf-8' codec can't decode bytes in position 0-1: unexpected end of data
```

So this does not degrade to replacement characters on that binding, it raises, and `module.cpp:743` re-throws the exception on the main thread. Lenient consumers substitute U+FFFD instead, which is the same mangling `rag_backend.cpp` already documents for RAG previews.

The LLM decode loop already solves this engine-side with a UTF-8 DFA (`engines/llamacpp/llamacpp_backend.cpp`, `Utf8State` plus `partial_utf8_buffer`). Notably the *same* Python `stream_token_cb` is used for LLM streaming and is safe there, purely because the engine buffers before calling back. The VLM streaming loop has no equivalent, and its own comment says it "Mirrors the LLM backend's stop_window pattern in run_decode_loop" while omitting the boundary walk that pattern depends on.

**What this changes.** A chunk is delivered only up to its last complete character; a trailing partial sequence is held until the rest arrives.

The guard goes in `vlm_module.cpp` rather than in the llamacpp engine because the component token callback is the single funnel every VLM backend's stream passes through, so one guard also covers backends that do not exist yet. The new `features/common/utf8_stream_boundary.h` matches the shape of the existing `features/common/special_token_filter.h`, the other per-stream chunk-shaping helper in that directory.

Accumulation into `cleaned_text` is deliberately not delayed, so the final result text is byte-for-byte unchanged. Only chunk boundaries move.

One judgment call worth flagging: if generation stops mid-character (max_tokens or cancellation), the leftover bytes are not delivered on the stream. They cannot render, and emitting them would put an invalid chunk on the stream for the sake of an incomplete character. `cleaned_text` still contains them, so the final result is unaffected.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring

## Testing
- [ ] Lint passes locally

`cmake -B build -DRAC_BUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Debug` plus `ctest`: builds with 0 errors, 101/101 pass.

Both boxes above are left unticked deliberately:

- **Lint.** I do not have the repo's pinned `clang-format` here. The version I do have (23.1.0) reports 1467 violations on the *untouched* `main` copy of `vlm_module.cpp`, so it disagrees with whatever CI uses and its verdict on the whole file is meaningless. What I can report: the new header is clean under it, and the one line of mine it flagged was rewritten to a form that does not depend on the version.
- **Tests.** No test is committed. The sibling helper in the same directory (`special_token_filter.h`) has no dedicated test either, so adding one here would exceed the convention for this class of header. I verified the logic with a throwaway harness instead (8 boundary assertions plus a simulation of the funnel): the U+1FAE0 piece sequence yields 2 chunks, both valid UTF-8, text preserved; without the guard 3 of 3 chunks are invalid. Happy to commit that as a test file if you would prefer it in-tree.

Not run: any on-device VLM streaming check. I have no VLM model or device here, so I have not observed the corrupted caption end to end. The claim rests on the vocabulary measurement above plus the code path, not on a live capture.

`core/` guard check: all four edited regions are outside any `#if`, and the new header includes only `<string>`, so no build configuration sees a different result.

None. This is native `core/` code with no binding or app changes.

## Labels
**SDKs:**
- [ ] `Swift SDK`
- [ ] `Kotlin SDK`
- [ ] `Flutter SDK`
- [ ] `React Native SDK`
- [ ] `Web SDK`

## Checklist
- [x] Self-review completed

The header carries the rationale, so no separate docs change.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved VLM streaming to prevent incomplete or malformed UTF-8 characters from appearing in incremental output.
  - Streaming callbacks now receive only complete characters, reducing display issues when text is split across streamed chunks.
  - Final accumulated responses remain unchanged while partial output is handled more reliably.

- **Documentation**
  - Clarified streaming callback behavior, including how text chunks are combined or held until complete UTF-8 characters are available.

### DIFF
--- a/bindings/swift/Sources/RunAnywhere/CRACommons/include/rac_vlm_component.h
+++ b/bindings/swift/Sources/RunAnywhere/CRACommons/include/rac_vlm_component.h
@@ -163,7 +163,12 @@ RAC_API rac_bool_t rac_vlm_component_supports_streaming(rac_handle_t handle);
  * @param image Image input
  * @param prompt Text prompt
  * @param options Generation options (can be NULL for defaults)
- * @param token_callback Called for each generated token
+ * @param token_callback Called with the next run of decoded text. Not one call
+ *                       per token: a chunk is emitted only up to its last
+ *                       complete UTF-8 character, so a token whose bytes finish
+ *                       a character started by the previous token arrives
+ *                       merged with it, and a token that only starts one is
+ *                       held until it is complete.
  * @param complete_callback Called when generation completes
  * @param error_callback Called on error
  * @param user_data User context passed to callbacks
--- a/bindings/swift/Sources/RunAnywhere/CRACommons/include/rac_vlm_types.h
+++ b/bindings/swift/Sources/RunAnywhere/CRACommons/include/rac_vlm_types.h
@@ -370,7 +370,9 @@ typedef struct rac_vlm_info {
 /**
  * @brief Simple VLM streaming callback
  *
- * Called for each generated token during streaming.
+ * Called with successive runs of decoded text during streaming, each ending on
+ * a complete UTF-8 character. This is not one call per token: see
+ * rac_vlm_component_process_stream.
  *
  * @param token The generated token string
  * @param user_data User-provided context
--- a/core/include/rac/features/vlm/rac_vlm_component.h
+++ b/core/include/rac/features/vlm/rac_vlm_component.h
@@ -174,7 +174,12 @@ RAC_API rac_bool_t rac_vlm_component_supports_streaming(rac_handle_t handle);
  * @param image Image input
  * @param prompt Text prompt
  * @param options Generation options (can be NULL for defaults)
- * @param token_callback Called for each generated token
+ * @param token_callback Called with the next run of decoded text. Not one call
+ *                       per token: a chunk is emitted only up to its last
+ *                       complete UTF-8 character, so a token whose bytes finish
+ *                       a character started by the previous token arrives
+ *                       merged with it, and a token that only starts one is
+ *                       held until it is complete.
  * @param complete_callback Called when generation completes
  * @param error_callback Called on error
  * @param user_data User context passed to callbacks
--- a/core/include/rac/features/vlm/rac_vlm_types.h
+++ b/core/include/rac/features/vlm/rac_vlm_types.h
@@ -406,7 +406,9 @@ typedef struct rac_vlm_info {
 /**
  * @brief Simple VLM streaming callback
  *
- * Called for each generated token during streaming.
+ * Called with successive runs of decoded text during streaming, each ending on
+ * a complete UTF-8 character. This is not one call per token: see
+ * rac_vlm_component_process_stream.
  *
  * @param token The generated token string
  * @param user_data User-provided context
--- a/core/src/features/common/utf8_stream_boundary.h
+++ b/core/src/features/common/utf8_stream_boundary.h
@@ -44,6 +44,52 @@ inline size_t incomplete_utf8_tail(const std::string& text) {
     return 0;  // three continuations with no lead in reach: already complete
 }
 
+/**
+ * Per-stream UTF-8 assembler. `feed()` returns the bytes that are safe to hand
+ * on now, holding back a trailing partial character until the rest arrives;
+ * `flush()` releases whatever is left at end of stream.
+ *
+ * Shaped like `rac::tokens::StreamFilter` on purpose, and held alongside one:
+ * both VLM stream funnels need the same per-stream shaping, so the logic lives
+ * in one type rather than being written out at each callback.
+ *
+ * Both return a reference to a buffer owned by the assembler, valid until the
+ * next `feed()` / `flush()` on the same instance.
+ */
+class Utf8Assembler {
+   public:
+    const std::string& feed(const std::string& chunk) {
+        out_.clear();
+        pending_ += chunk;
+        const size_t hold = incomplete_utf8_tail(pending_);
+        if (pending_.size() > hold) {
+            out_.assign(pending_, 0, pending_.size() - hold);
+            pending_.erase(0, pending_.size() - hold);
+        }
+        return out_;
+    }
+
+    /**
+     * End of stream. A remainder here is a character the backend never
+     * finished emitting, so it cannot render; dropping it keeps every
+     * delivered chunk valid UTF-8. Callers that accumulate the unassembled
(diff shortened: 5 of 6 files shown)


ayaangazali:
Fair catch, fixed in the push above.

`@param token_callback Called for each generated token` no longer described the contract once a partial character is held back, since a chunk can carry bytes from two tokens or be withheld entirely while a character completes. Updated `rac_vlm_component.h` and `rac_vlm_types.h`, plus the Swift CRACommons copies of both, so the four VLM declarations agree.

Worth noting the wording was already imprecise before this PR: the sentinel `StreamFilter` in the same funnel merges and withholds chunks too, so "each generated token" has not been literally true for a while. This change widens the gap, so it is the right place to correct it.

I left the LLM headers alone even though `rac_llm_component.h:199` has the identical line and the identical caveat. That is a separate finding in a path this diff does not touch, and folding it in here would make the PR two concerns. Happy to open it separately if you want it tracked.

Build 0 errors, ctest 101/101 after the change.

sanchitmonga22:
Thanks for this, @ayaangazali! The `rac_vlm_component_process_stream` path (Python bindings, Electron) no longer splits a UTF-8 character across streamed chunks — a real fix, and the new `incomplete_utf8_tail` helper handles partial, complete and malformed tails correctly.

1. `core/src/features/vlm/vlm_module.cpp:1191` — `generated_stream_token_trampoline`, the callback `rac_vlm_stream_proto` actually registers (the path Swift, Kotlin, Flutter, React Native and Web stream VLM through), still sends `display` straight into the protobuf token event with no UTF-8 boundary check — same gap in `flush_held_stream_text`'s end-of-stream flush. A character the vocab splits still reaches those five SDKs as invalid UTF-8. Could you add the same `utf8_tail` / `incomplete_utf8_tail` buffering there (or assemble UTF-8 once inside `rac_vlm_llamacpp_process_stream` so both call paths share it)?

Not blocking this PR: I opened #923 for the identical stale docstring in `rac_llm_component.h:199` you flagged — thanks for catching that. And no pressure on the regression test either way — happy to add it ourselves if you'd rather not.

Once the `rac_vlm_stream_proto` path is covered too we'll take another look. Thanks again!

sanchitmonga22:
Thanks for the offer! Since the fix still needs to cover `rac_vlm_stream_proto`'s callback (see the comment above), it'd be great to land the regression test alongside that update — a fake backend emitting a character split across pieces through both call paths would cover it well. No pressure either way if you'd rather we add it.

sanchitmonga22:
> Replying to [this comment](https://github.com/RunanywhereAI/runanywhere-sdks/pull/848#issuecomment-5590685799)

Good idea — that docstring is stale in exactly the same way. I opened #923 for it so it doesn't get lost; thanks for spotting it. Treating this as addressed.

ayaangazali:
You are right, and thank you for pinning down the exact callback. Fixed in the push above.

I had the wrong mental model of the funnel. I fixed `vlm_stream_token_callback` believing it was the single point every VLM backend's stream passes through. It is not: `rac_vlm_component_process_stream` reaches the engine through `rac_vlm_process_stream`, while the proto path calls `ref.ops->process_stream` directly at `vlm_module.cpp:1471` and lands in `generated_stream_token_trampoline`. Two funnels over one engine, and I only covered the one the Python and Electron bindings use.

I took your second suggestion rather than repeating the buffering, though one level up from where you put it. The hold-back logic is now `rac::tokens::Utf8Assembler`, shaped exactly like the `StreamFilter` that *both* stream contexts already hold, and each context now holds one of each. One implementation, both paths, and the symmetry is visible at the struct rather than being two copies of the same four lines.

I did consider assembling inside `rac_vlm_llamacpp_process_stream` as you suggested, which is genuinely the lowest common point. I went with commons instead because `engines/llamacpp` only sees the public headers under `core/include`, so putting it there would mean either exporting an internal utility into the public C surface or duplicating the helper per engine, and a future non-llamacpp VLM backend would have to remember to do it. Happy to move it if you would rather the engine own it.

Covered in this push:
- `generated_stream_token_trampoline`: `generation.token` and the VLM stream event now carry assembled text.
- `flush_held_stream_text`: feeds the sentinel filter's tail through the assembler, then releases what completes.

`ctx->text` and `cleaned_text` still accumulate the unassembled text, so final result text is byte-for-byte unchanged; only chunk boundaries move.

Checks: guard audit says the trampoline edits sit inside `#if defined(RAC_HAVE_PROTOBUF)` and the C-callback edits outside it, matching the code around them. Build 0 errors, ctest 101/101. Assembler exercised directly on the U+1FAE0 piece sequence: 2 chunks out, both valid UTF-8, text preserved; a stream truncated mid-character emits nothing invalid; a stray continuation byte does not stall it.

Thanks for opening #923 for the LLM docstring rather than asking me to fold it in here.

PR #800, opened 2026-08-26
docs: npm install commands pin a version that was never published

ayaangazali:
## Description

Every Web and React Native README pins its install command to `0.20.11`:

```
npm install @runanywhere/web@0.20.11 @runanywhere/web-llamacpp@0.20.11 @runanywhere/web-onnx@0.20.11
```

That version does not exist on npm. `@runanywhere/web` went `0.20.10` straight to `0.20.16`, so this command has never worked for anyone who copied it:

```
$ npm view @runanywhere/web@0.20.11 version
npm error code E404
npm error 404 No match found for version 0.20.11
```

Published `0.20.x` versions: `0.20.9, 0.20.10, 0.20.16, 0.20.17, 0.20.18, 0.20.19, 0.20.24, 0.20.27`.

This is not ordinary staleness. These nine files sit outside both `check_release_version_coherence.sh`'s doc list and `sync-versions.sh`'s rewrite set, so no release has ever moved them and none will.

Repins to `0.20.27`, the current latest for every affected package.

`packages/qhexrt/README.md` is the one exception: `@runanywhere/qhexrt` last published `0.20.19`, so that file pins core and qhexrt together at `0.20.19` rather than pairing `core@0.20.27` with a qhexrt version that does not exist. The Web README's own guidance is that "mismatched core/backend versions may fail peer dependency checks."

Root `README.md` is deliberately untouched — #798 is already refreshing its version line, and I did not want to collide with it.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [x] Documentation update
- [ ] Refactoring

## Testing

Docs only, no code paths touched. Verified every version this PR documents actually resolves:

```
@runanywhere/web@0.20.27           -> 0.20.27
@runanywhere/web-llamacpp@0.20.27  -> 0.20.27
@runanywhere/web-onnx@0.20.27      -> 0.20.27
@runanywhere/core@0.20.27          -> 0.20.27
@runanywhere/llamacpp@0.20.27      -> 0.20.27
@runanywhere/onnx@0.20.27          -> 0.20.27
@runanywhere/core@0.20.19          -> 0.20.19
@runanywhere/qhexrt@0.20.19        -> 0.20.19
```

- [ ] Lint passes locally
- [ ] Added/updated tests for changes

No linter covers these files and no test asserts README contents, so neither box is ticked rather than claiming a check I did not run.

## Labels
- [x] `React Native SDK`
- [x] `Web SDK`

## Checklist
- [x] Code follows project style guidelines
[...]

PR #849, opened 2026-09-08
docs: four stale references (skills rcli, npm pins, packaging, ABI history)

ayaangazali:
## Description

Four documentation findings, each of which I had opened as its own PR. That was my mistake: four separate docs PRs is four review threads for what should be one. Consolidating them here and closing #822, #800, #758 and #829 in favour of this.

All four re-verified on `45e91276e` (0.20.36).

**1. Skill docs walk through a binary that is not in this repo.** `sdk-publish`, `sdk-test-starters` and `telemetry-e2e` all give `rcli` commands. There is no `rcli` directory in the tree; it moved out. Both the `.agents/` and `.claude/` copies are updated, since they are separate files rather than symlinks.

**2. npm install commands pin a version that was never published.** The React Native and Web READMEs say `@0.20.11`:

```
$ npm view @runanywhere/core versions          # 0.20.11 present? False
$ npm view @runanywhere/web-llamacpp versions  # 0.20.11 present? False
```

Copying the documented command fails. Changed to `0.20.27`, which is the current published release for all three packages.

`bindings/react-native/packages/qhexrt/README.md` deliberately keeps `0.20.19`: that is qhexrt's own latest published version, not a stale pin.

**3. Two stale references in packaging docs.** `bindings/python/PACKAGING.md` tells you to follow the `python-linux` CI job; `grep -c python-linux .github/workflows/pr-build.yml` returns 0, because #757 removed it. The manylinux wheel therefore has no CI recipe left to copy, which the doc now says. Separately, `scripts/validation/verify_default_pool.sh` tells you to run `idl/codegen/generate_defaults_pool.sh`; the file on disk is `generate_defaults_pool.py`.

**4. The plugin ABI history stops two versions short.** `RAC_PLUGIN_API_VERSION` is `11u` but the history block's last entry is `9u`, so the two most recent ABI breaks are undocumented. Added both, in the style of the surrounding entries: 10u for image embedding (`reserved_slot_3` to `image_embedding_ops`) and 11u for OCR (`reserved_slot_4` to `ocr_ops`).

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [x] Documentation update
- [ ] Refactoring

## Testing
- [ ] Lint passes locally

Neither applies and neither is ticked. Every added line in the one non-markdown source file (`rac_plugin_entry.h`) is a comment, verified mechanically: of the added lines, 0 are not comment lines. `verify_default_pool.sh` changes one comment line. No behaviour to test and nothing for a linter to judge beyond prose.

None. Documentation and comments only.

## Labels
**SDKs:**
- [ ] `Swift SDK`
- [ ] `Kotlin SDK`
- [ ] `Flutter SDK`
- [x] `React Native SDK` - README install commands
- [x] `Web SDK` - README install commands
- [x] `Commons` - ABI history comment in `core/include`

## Checklist
- [x] Self-review completed

### DIFF
--- a/.agents/skills/sdk-publish/SKILL.md
+++ b/.agents/skills/sdk-publish/SKILL.md
@@ -375,21 +375,24 @@ This is enforced, not just documented: once `v$VERSION` is tagged on `runanywher
 
 **If you force-move a tag, SwiftPM/Xcode consumers cache the OLD commit in several places — clearing just `Package.resolved` is not enough.** Verifying your own fix (or having a downstream consumer verify it) can silently keep failing even after the fix is correctly pushed, because: (1) any consumer repo's own tracked `*.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved` pins the exact old resolution and `xcodebuild -resolvePackageDependencies` respects it rather than re-resolving; (2) `~/Library/Caches/org.swift.swiftpm/repositories/<pkg>-<hash>/` is a global shared clone cache keyed by repo, oblivious to a moved tag; (3) `~/Library/org.swift.swiftpm/security/fingerprints/<pkg>-<hash>.json` records the fingerprint SwiftPM saw for that tag and can silently keep using it after a mismatch instead of failing loud. To actually re-resolve after moving a tag: delete the consumer's tracked `Package.resolved`, `rm -rf .build .swiftpm` and any `~/Library/Developer/Xcode/DerivedData/<Scheme>-*`, AND clear the matching entries under both global `org.swift.swiftpm` cache directories for every affected package identity — then resolve fresh.
 
-## 5. macOS rcli notarization
+## 5. rcli ships from RunanywhereAI/RCLI, not from this release
 
-CI (`native_rcli_macos` in `release.yml`) usually already produces a notarized, stapled artifact when Developer ID + notary secrets are configured in repo secrets. **Check the release assets first** before assuming you need to notarize anything yourself locally:
+rcli is not built, versioned, or released here. `release.yml` says so in its own header:
+
+> Official CLI bottles (rcli) are NOT produced here — they ship from
+> RunanywhereAI/RCLI, which consumes those kits via `find_package(RunAnywhere)`.
+
+So when cutting a release of *this* repo:
+
+- **Do not look for rcli assets on `v$VERSION`.** `v0.20.25` was the last release that carried any (`rcli-macos-arm64-v0.20.25.tar.gz` and friends); `v0.20.30` and `v0.20.31` carry none, and there is no `native_rcli_macos` job in `release.yml` to produce them.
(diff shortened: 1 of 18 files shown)


ayaangazali:
Two separate things here. One is a mistake in my PR description, the other is a suggestion I do not think holds.

**My description was wrong.** I wrote "Changed to unpinned installs so the docs cannot drift out of sync with the registry again." The diff does no such thing: it changes `0.20.11` to `0.20.27`. I have corrected the body. The finding and the diff are both fine, the sentence describing them was not.

**On aligning everything to `0.20.36`:** that would break the commands. `0.20.36` is not on npm:

```
@runanywhere/web          latest=0.20.27   has 0.20.36=False
@runanywhere/web-llamacpp latest=0.20.27   has 0.20.36=False
@runanywhere/core         latest=0.20.27   has 0.20.36=False
@runanywhere/qhexrt       latest=0.20.19   has 0.20.27=False
```

`0.20.27` is the current published release for those three, which is why the diff targets it. Documenting `0.20.36` would reproduce exactly the bug this PR fixes, just one version further along.

Same reason `bindings/react-native/packages/qhexrt/README.md` keeps `0.20.19`: that is qhexrt's latest published version, so it is correct rather than stale.

The `>=0.20.36` peer range you found is a real inconsistency, but it runs the other way: the in-repo manifests declare a version that has not shipped. Installing the published `0.20.27` packages resolves against the published `0.20.27` peer ranges, not the unreleased in-repo ones, so the documented commands work today. That gap between manifest and registry is a release-process question rather than a docs one, and it belongs in its own issue instead of this thread.

Worth saying plainly: this drift is the argument for not pinning at all in a README, and my original instinct there was right even though the diff does not implement it. I have kept the diff matching the registry rather than expanding scope mid-review, but if you would rather these commands were unpinned, say so and I will make that change instead.

sanchitmonga22:
Thanks for this, @ayaangazali! The core fixes here are solid — the npm pin correction (0.20.11 → 0.20.27/0.20.19) checks out against the live registry, and the ABI 10u/11u history addition matches the real header. Because this PR also touches the `.agents`/`.claude` skill docs that other agents follow for releasing and testing the SDK (`sdk-publish`, `sdk-test-starters`, `telemetry-e2e`), a maintainer will give those a final look before merging.

Two small things worth a look while it's open: `sdk-test-starters/SKILL.md:154` still has a "for rcli instead" block a few lines after the new sentence saying rcli isn't in this repo anymore, and `telemetry-e2e/SKILL.md:265` still gives a `cmake --preset rcli-macos-release` build command that no longer exists here.

sanchitmonga22:
> Replying to [this comment](https://github.com/RunanywhereAI/runanywhere-sdks/pull/849#issuecomment-5590689314)

That makes sense — thanks for checking the registry yourself. Pinning to what's actually published (0.20.27 for core/llamacpp/web/web-llamacpp, 0.20.19 for qhexrt) is the right call here; CodeRabbit's suggested 0.20.36 isn't installable. Treating that thread as addressed. On whether to unpin these commands instead — that's a call for whoever clears the hold on this PR; I've left it open for them in the review notes.

Issue #869, opened 2026-09-11
Add CI compile coverage for the Flutter/RN Android OkHttp transport copies


sanchitmonga22:
Follow-up from #741 by @shubhamsinnh — thanks again for that PR!

## What
PR CI only compiles the Kotlin SDK's `OkHttpHttpTransport.kt` (the `kotlin-android` job, scoped to `bindings/kotlin`). The Flutter plugin's copy (`bindings/flutter/packages/runanywhere/android/src/main/kotlin/com/runanywhere/sdk/httptransport/OkHttpHttpTransport.kt`) and the React Native core's copy (`bindings/react-native/packages/core/android/src/main/java/com/runanywhere/sdk/httptransport/OkHttpHttpTransport.kt`) are never compiled by any PR CI job — `flutter-pubget` only runs Dart analysis/tests and `rn-typecheck` only runs TypeScript, a C++ header compile, ESLint and Jest.

## Why it matters
A future change to either copy (this PR's fix included) can silently break the Flutter or RN Android build; nothing in CI catches it before merge, and it would only surface downstream in the RN/Flutter example apps or on-device.

## Suggested approach
Add a CI step (or job) that runs a Gradle compile of the module owning `OkHttpHttpTransport.kt` under `bindings/flutter/packages/runanywhere/android` and `bindings/react-native/packages/core/android` — even a `compileDebugKotlin`/`assembleDebug` scoped to that module — alongside the existing `kotlin-android` job.

## Done when
- [ ] CI fails if either the Flutter or RN Android `OkHttpHttpTransport.kt` copy fails to compile.

Not blocking #741. @shubhamsinnh, you know this code well now — you're welcome to take this one if you're interested.

_Opened with help from Claude Code and Codex._

_Reviewed with help from Claude Code and Codex._

PR #741, opened 2026-08-18
fix(android): map OkHttp stream cancellation to RAC_ERROR_CANCELLED

shubhamsinnh:
## Description
On Android, canceling an in-flight streaming/resume HTTP request was reported as a generic network error instead of a cancellation, diverging from iOS.

`OkHttpHttpTransport.streamInternal` (backing the `request_stream` and `request_resume` vtable slots) declares the `Call` *inside* the `try`, and its `catch (e: Throwable)` hardcodes `cancelled = false`. When `Call.cancel()` fires while `call.execute()` is still waiting for response headers — the normal shape of SDK teardown via `cancelAllStreams()`, or a cancel racing stream startup — OkHttp throws `IOException: Canceled`, and the C++ adapter's `map_kotlin_transport_error(...)` collapses that to `RAC_ERROR_NETWORK_ERROR`.

The C++ adapter already returns `RAC_ERROR_CANCELLED` when the `StreamResponse.cancelled` field is `true` (it checks `f_sr_cancelled` before the error-message path). The flag just was never set on this path. Swift's `URLSessionHttpTransport.mapTransportError` maps `NSURLErrorCancelled` → `RAC_ERROR_CANCELLED`, so this restores cross-SDK parity.

## Fix
Hoist the created `Call` into a nullable function-scoped variable and read `call.isCanceled()` in the catch:

- `var activeCall: Call? = null` before the `try`
- `activeCall = call` after `clientForCall.newCall(request)`
- `cancelled = activeCall?.isCanceled() == true` in the `catch`

This also fixes the cancel-races-startup path (`cancelAllStreams()` landing before `execute()`). Applied identically to the three copies of `OkHttpHttpTransport` (`bindings/kotlin`, `bindings/flutter/.../android`, `bindings/react-native/.../android`).

## Scope note
The **buffered** `executeRequest` path still has no `Call` handle and no cancellation mechanism (it's a fluent `clientForCall.newCall(request).execute().use { ... }`), so it cannot be cancelled by the SDK at all — a separate, larger parity gap (Swift's buffered path is cancellable via `URLSessionTask.cancel()`). That is intentionally out of scope for this change.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring

## Testing
- [x] Lint passes locally
- [ ] Added/updated tests for changes (no new tests per repo guidance)

Local:
- `git diff --check` — clean.
[...]

PR #930, opened 2026-09-11
fix(android): classify host-dispatcher stream cancellation as RAC_ERROR_CANCELLED

### LINKED ISSUE #870: Stream cancel via a host client's own dispatcher isn't classified as RAC_ERROR_CANCELLED
Follow-up from #741 by @shubhamsinnh — thanks again for that PR!

## What
`bindings/kotlin/src/main/kotlin/com/runanywhere/sdk/httptransport/OkHttpHttpTransport.kt:584` (and the matching lines in the Flutter and React Native Android copies) now classify a cancelled stream by checking only `slot.cancelRequested`, which is set solely by the SDK's own cancel paths (the chunk callback returning `false`, or `cancelAllStreams()`). If a host app installs its own `OkHttpClient` via `setHttpClient()` and cancels calls directly through that client's shared `Dispatcher` (e.g. `dispatcher.cancelAll()`) instead of the SDK's `cancelAllStreams()`, the streaming call is still cancelled, but `slot.cancelRequested` is never set — the transfer aborts with `IOException: Canceled`, which now maps to `RAC_ERROR_NETWORK_ERROR` and gets retried by the download orchestrator, instead of `RAC_ERROR_CANCELLED`.

## Why it matters
This is a narrow, unsupported-path edge case (the SDK's own `cancelAllStreams()` is the documented way to cancel), but a host that reaches into its installed client can still trigger a spurious retry instead of a clean cancel.

## Suggested approach
Combine both signals while still excluding OkHttp's own timeout-triggered cancel, e.g. `cancelled = slot.cancelRequested || (call.isCanceled() && e !is InterruptedIOException)` (timeouts surface as `InterruptedIOException`/`SocketTimeoutException`).

## Done when
- [ ] A stream cancelled through a host-installed client's dispatcher reports `RAC_ERROR_CANCELLED`, and a call timeout still reports `RAC_ERROR_TIMEOUT`, in all three Android transport copies.

Not blocking #741. @shubhamsinnh, you know this code well now — you're welcome to take this one if you're interested.

shubhamsinnh:
## Description

Follow-up to #741 (issue #870). A streaming call cancelled through a host-installed `OkHttpClient`'s own `Dispatcher` (for example `dispatcher.cancelAll()` after `setHttpClient()`) aborts with an `IOException`, but the SDK-owned `StreamSlot.cancelRequested` marker is never set — so the failure was reported as a network error and handed to the download orchestrator's retry policy instead of surfacing as `RAC_ERROR_CANCELLED`.

All three Android transport copies now classify cancellation the same way:

| File | Change |
|---|---|
| `bindings/kotlin/src/main/kotlin/com/runanywhere/sdk/httptransport/OkHttpHttpTransport.kt` | shared `isStreamCancellation` helper + `java.io.InterruptedIOException` import |
| `bindings/flutter/packages/runanywhere/android/src/main/kotlin/com/runanywhere/sdk/httptransport/OkHttpHttpTransport.kt` | same |
| `bindings/react-native/packages/core/android/src/main/java/com/runanywhere/sdk/httptransport/OkHttpHttpTransport.kt` | same |

```kotlin
private fun isStreamCancellation(failure: Throwable, call: Call?, slot: StreamSlot): Boolean {
    if (synchronized(slot) { slot.cancelRequested }) return true
    return call != null &&
        call.isCanceled() &&
        failure is IOException &&
        failure !is InterruptedIOException
}
```

Used by both failure sites so the setup-time catch and the body-read catch cannot drift:

- `streamInternal` catch → `cancelled = isStreamCancellation(e, activeCall, slot)`
- `drainBody` read catch → clean cancellation when the helper matches, otherwise rethrown unchanged

`SocketTimeoutException` is covered through its `InterruptedIOException` inheritance; no message matching or raw error strings are introduced. The SDK's own cancel paths (`cancelAllStreams()` and the chunk callback returning `false`) and every other `IOException` are unchanged. No C++ error mapping, retry policy, or public API changes.

## Type of Change

- [x] Bug fix

## Testing

- [x] Lint passes locally
- [ ] Added/updated tests for changes — per the plan for this issue: no unit tests; validated with the existing checks plus focused behavioral review

Validation performed:

- ktlint 1.5.0 (repo pin) over the three files, on the working tree and on a pristine `main` checkout: identical results (30/94/94 findings), so the change adds no ktlint violations.
- detekt 1.23.8 (repo pin) with `bindings/kotlin/detekt.yml` on the Kotlin copy: exit 0, no findings.
- Compiled all three transport owners with kotlinc 2.0.21 against real okhttp 4.12.0 + okio 3.6.0 (project/Android types stubbed outside the repo): exit 0 for each.
- Behavioral probe against real OkHttp 4.12.0:

| Scenario | exception | `call.isCanceled()` | classified as cancellation |
|---|---|---|---|
| `Call.cancel()` | `IOException` | true | yes |
| host `dispatcher.cancelAll()` in flight | `SocketException` | true | yes |
| OkHttp `callTimeout` | `InterruptedIOException` | true | no |
| connect timeout | `SocketTimeoutException` | false | no |

- JNI contract re-confirmed unchanged: `cancelled = true` → `RAC_ERROR_CANCELLED`; a `SocketTimeoutException`/`InterruptedIOException` error-message prefix → `RAC_ERROR_TIMEOUT`; otherwise `RAC_ERROR_NETWORK_ERROR`.

## Labels

- [x] `Kotlin SDK` - Changes to Kotlin SDK (`bindings/kotlin`)
- [x] `Flutter SDK` - Changes to Flutter SDK (`bindings/flutter`)
- [x] `React Native SDK` - Changes to React Native SDK (`bindings/react-native`)

Note: this PR changes only Kotlin source under `bindings/...`; the Flutter and React Native Android copies are compiled by CI once #869 lands (that PR adds focused `compileDebugKotlin` coverage for both). Until then, CI compiles the Kotlin SDK copy via the existing `kotlin-android` job.

Fixes #870.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of cancelled streaming requests, including cancellations initiated outside the SDK.
  - Prevented cancelled streams from being incorrectly reported as transport failures.
  - Preserved timeout reporting so timed-out requests continue to be identified as timeout errors.

### DIFF
--- a/bindings/flutter/packages/runanywhere/android/src/main/kotlin/com/runanywhere/sdk/httptransport/OkHttpHttpTransport.kt
+++ b/bindings/flutter/packages/runanywhere/android/src/main/kotlin/com/runanywhere/sdk/httptransport/OkHttpHttpTransport.kt
@@ -37,6 +37,7 @@ import okhttp3.Request
 import okhttp3.RequestBody
 import okhttp3.RequestBody.Companion.toRequestBody
 import java.io.IOException
+import java.io.InterruptedIOException
 import java.time.Duration
 import java.util.concurrent.Callable
 import java.util.concurrent.ConcurrentHashMap
@@ -464,7 +465,7 @@ object OkHttpHttpTransport {
                 statusCode = 0,
                 headers = emptyArray(),
                 errorMessage = "${e.javaClass.simpleName}: ${e.message ?: "unknown"}",
-                cancelled = activeCall != null && synchronized(slot) { slot.cancelRequested },
+                cancelled = isStreamCancellation(e, activeCall, slot),
             )
         } finally {
             inFlightStreams.remove(streamId)
@@ -494,7 +495,11 @@ object OkHttpHttpTransport {
                     try {
                         input.read(buffer)
                     } catch (io: IOException) {
-                        if (synchronized(slot) { slot.cancelRequested }) {
+                        // A cancelled call surfaces as an IOException — treat
+                        // that as a clean cancellation rather than a transport
+                        // failure. Timeouts stay transport errors (see
+                        // [isStreamCancellation]).
+                        if (isStreamCancellation(io, call, slot)) {
                             cancelled = true
                             break
                         }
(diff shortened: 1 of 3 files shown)


shubhamsinnh:
Hi @sanchitmonga22 , 
Please have a look when you get a chance.
It addresses the issue :  https://github.com/RunanywhereAI/runanywhere-sdks/issues/870
 
 thanks.

PR #933, opened 2026-09-12
fix(bindings): surface directory iterator errors

### LINKED ISSUE #890: Duplicated list_dir bindings swallow directory-iterator errors
Follow-up from #808 by @shubhamsinnh — thanks again for that PR!

## What
The four binding copies of `list_dir` (`bindings/electron/native/posix_platform_adapter.cpp:296`, `bindings/electron/native/win32_platform_adapter.cpp:264,281-282`, and the matching Python files) already drift from `core/src/desktop/desktop_adapter.cpp`. Most visibly, their `directory_iterator` loop ends silently on an error and the function still returns `RAC_SUCCESS` with a partial or empty listing, whereas the core adapter returns `filesystem_error_to_rac(ec, ...)` (`desktop_adapter.cpp:283-285,320-322`). The core adapter also filters hidden entries, masks `is_dir` on a type error, and logs skipped oversized names — none of which the four binding copies do.

## Why it matters
Today the blast radius is small: `model_registry_refresh.cpp` treats an iterator error and an empty success listing the same way (it just skips the directory). But #808 needed the exact same 3-line fix pasted into four files because there's no shared implementation, so this class of bug will keep recurring.

## Suggested approach
Share one `list_dir` implementation across the Electron and Python bindings, or at minimum port the core adapter's iterator-error handling into the four binding copies. Note the Electron `RAC_DESKTOP_ADAPTER=OFF` lane (win-arm64 / QHexRT) doesn't link the core desktop adapter, so sharing it outright isn't a drop-in change.

## Done when
- [ ] All four binding `list_dir` implementations return an error result (not `RAC_SUCCESS` with a partial listing) when the directory iterator itself fails, matching the core desktop adapter's contract.

Not blocking #808. @shubhamsinnh, you know this code well now — you're welcome to take this one if you're interested.

shubhamsinnh:
## Description

- Mirror the core desktop adapter's structured filesystem-error mapping in all four Electron and Python directory-listing adapters.
- Return mapped errors from both the directory probe and iteration instead of reporting failures as missing directories or successful partial listings.

Fixes #890

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring

## Testing

- [ ] Lint passes locally (clang-format is unavailable on this host)
- [ ] Added/updated tests for changes (no new test work requested)
- [x] git diff --check
- [x] C++20 syntax compilation for both Win32 adapter translation units
- [ ] POSIX syntax compilation (no POSIX toolchain or WSL distribution is installed on this host)

## Checklist

- [x] Self-review completed
- [x] Documentation updated (not needed)

## Screenshots

Not applicable; native error-handling change only.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved directory-listing error reporting across supported platforms.
  - Missing paths, permission failures, and storage exhaustion now return more specific errors.
  - Directory checks no longer incorrectly classify probe failures as missing directories.
  - Errors encountered while reading directory contents are no longer treated as successful completion.

### DIFF
--- a/bindings/electron/native/posix_platform_adapter.cpp
+++ b/bindings/electron/native/posix_platform_adapter.cpp
@@ -285,6 +285,13 @@ rac_result_t posix_get_memory_info(rac_memory_info_t* out, void*) {
 // Directory enumeration + probe (portable — same as win32's std::filesystem)
 // ---------------------------------------------------------------------------
 
+rac_result_t filesystem_error_to_rac(const std::error_code& ec, rac_result_t fallback) {
+    if (ec == std::errc::no_such_file_or_directory) return RAC_ERROR_FILE_NOT_FOUND;
+    if (ec == std::errc::permission_denied) return RAC_ERROR_PERMISSION_DENIED;
+    if (ec == std::errc::no_space_on_device) return RAC_ERROR_STORAGE_FULL;
+    return fallback;
+}
+
 rac_result_t posix_list_dir(const char* dir_path, rac_directory_entry_t* out_entries,
                             size_t* in_out_count, void*) {
     if (!dir_path || !in_out_count) return RAC_ERROR_INVALID_ARGUMENT;
@@ -310,6 +317,7 @@ rac_result_t posix_list_dir(const char* dir_path, rac_directory_entry_t* out_ent
         }
         ++written;
     }
+    if (ec) return filesystem_error_to_rac(ec, RAC_ERROR_STORAGE_ERROR);
     *in_out_count = out_entries ? written : total;
     return RAC_SUCCESS;
 }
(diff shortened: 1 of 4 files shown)


shubhamsinnh:
Hey @sanchitmonga22, please rerun the failed `swift-spm` check when you get time.

The GitHub-hosted runner failed while downloading `nlohmann/json` because it couldn’t resolve `github.com` through DNS:

```text
fatal: unable to access 'https://github.com/nlohmann/json.git/':
Could not resolve host: github.com
```

All other checks passed, so this appears to be a transient runner network issue rather than a code failure. Thanks!

PR #943, opened 2026-09-13
fix(web): distinguish known exhausted storage and RAM probes from unknown (#899)

### LINKED ISSUE #899: Distinguish a known-and-exhausted storage probe from an unknown one
Follow-up from #816 by @ayaangazali — thanks again for that PR!

## What
`can_fit` (and `can_run`) now treat `available_storage <= 0` (and `available_ram <= 0`) as "unknown, so let it through" (`core/src/infrastructure/model_management/model_compatibility.cpp:69`), matching `idl/model_types.proto:637-641`'s documented "0 = unknown" contract. But some platform probes also return 0 when the answer is known and the volume is actually full: the Web probe returns 0 whenever `quota - usage <= 0` (`bindings/web/packages/core/src/Public/API/Namespaces/models.ts:63-66`). After #816, an exhausted browser quota now reads as `canFit=true` instead of "not enough storage."

## Why it matters
On native platforms this is mostly cosmetic: the download planner independently re-probes free space and fails closed when it's unknown (`core/src/infrastructure/download/download_orchestrator.cpp`, the non-`__EMSCRIPTEN__` branch). But that same file sets `require_free_space = false` under `__EMSCRIPTEN__`, so on Web nothing else catches it — the download starts and only fails later with a raw browser quota error, instead of the clear "not enough storage" message this whole compatibility path exists to show. The same 0-means-both-things ambiguity already existed for RAM before this PR; this just makes it visible for storage too.

## Suggested approach
Let a probe report "known and exhausted" distinctly from "unknown" — for example, have the Web probe return a small positive sentinel (e.g. 1 byte) when the quota is known but exhausted rather than clamping to 0, or give `ModelCompatibilityRequest` field presence (an `optional` field) instead of overloading 0 for both meanings. Whichever approach is chosen, apply it symmetrically to `available_ram_bytes` and `available_storage_bytes`.

## Done when
- [ ] A quota-exhausted Web probe (or a genuinely-zero-free native probe) reports `canFit=false`/`canRun=false`, distinguishable from an unresolved probe.
- [ ] A platform whose probe cannot run at all still passes through as "unknown" exactly as today — no regression for that case.

Not blocking #816. @ayaangazali, you know this code well now — you're welcome to take this one if you're interested.

Adityakk9031:
﻿## Description
Follow-up to #816. Addresses #899:
- In `bindings/web/packages/core/src/Public/API/Namespaces/models.ts`, distinguish between an unknown probe (returns 0, which commons interprets as unknown and allows to pass) and a probe that is known but exhausted (quota or memory <= 0).
- When `navigator.storage.estimate()` succeeds with `quota > 0` but free space (`quota - usage`) is <= 0, return a 1-byte sentinel (`KNOWN_EXHAUSTED_BYTES_SENTINEL = 1`) rather than clamping to 0. Since models require > 1 byte, commons compatibility correctly evaluates `canFit = false`.
- Symmetrically, when `navigator.deviceMemory` is reported as <= 0, return `KNOWN_EXHAUSTED_BYTES_SENTINEL = 1` so commons evaluates `canRun = false`.
- If the browser probe cannot run or throws, 0 is returned to preserve the fail-open "unknown passes through" behavior.
- Added unit tests in `models.probes.test.ts` verifying exhausted, positive, and unknown states for both probes.

Closes #899.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring

## Testing
- [x] Added unit tests in `packages/core/tests/unit/Public/API/Namespaces/models.probes.test.ts` (10/10 pass)
- [x] Full Web SDK unit test suite passes (252/252 tests)
- [x] `npm run typecheck` passes
- [x] ESLint passes on `packages/core`

## Checklist
- [x] Self-review completed

## Summary by CodeRabbit

* **Bug Fixes**
  * Memory and storage availability checks now distinguish between unknown capacity and confirmed exhaustion.
  * Exhausted memory or storage is reported consistently when available capacity is zero or negative.
  * Invalid or non-finite memory readings are now treated as unknown instead of being reported as available.

* **Tests**
  * Added coverage for available capacity, exhausted resources, unknown values, invalid readings, missing platform APIs, and storage estimation errors.


Adityakk9031:
@sanchitmonga22 opened some more please have a look

Issue #886, opened 2026-09-11
docs: rcli was renamed to wally — update README install instructions


sanchitmonga22:
Follow-up from #798 by @ziyaad-mallick — thanks again for that PR!

## What
`RunanywhereAI/RCLI` has been renamed to `RunanywhereAI/wally` (the old URL now redirects). The README's rcli section (`README.md:380-415`, including the quick-start block at `README.md:111-118` and the PowerShell line this PR adds at `README.md:412`) still documents `rcli` commands and links `RunanywhereAI/RCLI`. That URL still resolves and both `install.sh` and `install.ps1` still run, but they now install a binary called `wally` (`wally.exe` on Windows, `wally-<version>-*` release assets), so every `rcli ...` example in the README fails for someone who follows it end to end. The README's Homebrew install path should be checked against the same rename.

## Why it matters
Every documented rcli command in the README (quick-start, the dedicated rcli section, and the SDK support table) breaks or misleads a reader the moment they try it.

## Suggested approach
Decide whether this repo's README should track the tool's new name (`wally`) or pin to an older release that still ships `rcli`, then update the quick-start block, the rcli details section (binary name, repo link, install commands) and the SDK table row consistently. Reconcile with the Homebrew formula too.

## Done when
- [ ] README's rcli/wally references (binary name, repo link, install commands) are internally consistent
- [ ] The documented install commands (brew, curl|sh, irm|iex) each produce a binary whose name matches what the README tells the reader to run

Not blocking #798. @ziyaad-mallick, you know this code well now — you're welcome to take this one if you're interested.

_Opened with help from Claude Code and Codex._

_Reviewed with help from Claude Code and Codex._

ziyaad-mallick:
I'll take this one.

Before starting I checked what each documented install path actually produces today:

- `install.sh` and `install.ps1` on wally `main` (`efcf54b`) install `wally` / `wally.exe` from `RunanywhereAI/wally` (`install.sh:16`, `install.sh:124`, `install.ps1:144`).
- `brew install runanywhereai/tap/rcli` still installs `rcli`: `Formula/rcli.rb` in `homebrew-tap` is at 0.20.24, pulls the `rcli-*-v0.20.24` archives from this repo's releases, and symlinks `rcli` (last tap commit `e9c777b`, Aug 18). wally's own repo already carries `Formula/wally.rb`, but the tap only has `rcli.rb`.

[...]

PR #685, opened 2026-08-13
refactor(qhexrt): use explicit URL registration

shubhamsinnh:
## Description
Follow-up to [#603](https://github.com/RunanywhereAI/runanywhere-sdks/pull/603), incorporating the QHexRT ownership direction discussed in [#532](https://github.com/RunanywhereAI/runanywhere-sdks/pull/532). This is also prerequisite work for the generic catalog registration and refresh flow requested in [#523](https://github.com/RunanywhereAI/runanywhere-sdks/issues/523); that batch API remains a separate follow-up.

The earlier implementation mixed two responsibilities: QHexRT owned product catalog data, while example applications also carried catalog metadata. This change keeps product catalog ownership in applications and leaves QHexRT responsible for backend registration, device capability probing, and architecture-variant resolution.

Applications register their selected artifact references through the shared model-registration API. Commons still resolves logical QHexRT bundle references to the supported device variant (`v75`, `v79`, or `v81`); already pinned variant references remain unchanged.

This PR intentionally contains only the SDK-repository portion of the work:

- remove the native QHexRT product catalog, catalog-only bridges, and catalog-only tests
- remove catalog-specific model selection from QHexRT while retaining device-aware bundle resolution
- register application-owned Flutter and React Native QHexRT entries through the shared SDK API
- preserve download size, context length, source, description, and capability metadata across Swift, Kotlin, Flutter, and React Native registration paths
- preserve native registration failure codes through Kotlin JNI without introducing a throwing C++ allocation on the error path
- keep download size independent from runtime memory requirements
- retain device probing and backend-registration responsibilities in QHexRT

The standalone Android application changes are intentionally excluded and should be reviewed separately in `RunanywhereAI/runanywhere-android`.

React Native archive/multi-file and Web, Electron, and Python registration-metadata parity remain tracked separately in #857 and are not part of this PR.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [x] Refactoring

## Testing
- [ ] Lint passes locally
- [x] Added/updated tests for changes

Validation performed on the current branch:

- `git diff --check` — passed
- `git diff --check origin/main...HEAD` — passed
- branch ancestry check — current branch is rebased onto the latest fetched `origin/main`
- focused review of the Swift metadata path, Kotlin JNI error ownership, QHexRT bundle policy, and catalog bridge removals — passed
- verified no registration path defaults download size from memory requirements — passed
- platform builds and SDK checks — pending on the new CI run triggered by this update

[...]

PR #951, opened 2026-09-15
fix(release): keep the root README's install pins on core/VERSION

### LINKED ISSUE #885: docs: keep root README.md version pins in sync with core/VERSION
Follow-up from #798 by @ziyaad-mallick — thanks again for that PR!

## What
Nothing keeps the root `README.md`'s version numbers in sync with `core/VERSION`. `scripts/release/sync-versions.sh` and `scripts/validation/gates/check_release_version_coherence.sh` only touch `bindings/**` files — never the root README. Today the "all SDKs ship on one version line" sentence (`README.md:434`) and the per-package install pins (`README.md:198-201, 235-239, 271, 306, 375`) already disagree with each other and with what's actually published: `npm view @runanywhere/core@0.20.11` and `@runanywhere/web@0.20.11` both 404, and `https://pypi.org/pypi/runanywhere/json` 404s too, so the documented `pip install runanywhere==0.20.11` and `npm install @runanywhere/core@0.20.11 ...` commands fail for anyone who copies them today.

## Why it matters
A reader who follows the README's own install commands hits a 404 on the first try, which is a bad first impression and generates avoidable support questions.

## Suggested approach
Add the root `README.md` version line and install-command pins to `sync-versions.sh` so a release bump updates them automatically, and add a check to `check_release_version_coherence.sh` so a stale README fails the gate the same way a stale binding does.

## Done when
- [ ] `sync-versions.sh` updates README.md's version line and install-command pins on a version bump
- [ ] `check_release_version_coherence.sh` fails when README.md's version references drift from `core/VERSION`

Not blocking #798. @ziyaad-mallick, you know this code well now — you're welcome to take this one if you're interested.

ziyaad-mallick:
Closes #885

Opened as a draft because of the open question at the bottom: as written, this pins the README to a version no registry has yet. I'll mark it ready once you pick a source of truth.

## What changed

- `scripts/release/sync-versions.sh`: after the release-doc loop, the root `README.md` is bumped by five pin shapes: the Maven coordinates (`io.github.sanchitmonga22:runanywhere-*:`), the pub.dev constraints (`runanywhere*: ^`), the npm specs (`@runanywhere/*@`), `pip install runanywhere==`, and the "one version line, currently **x.y.z**" sentence. Each shape matches any `x.y.z`, not only the current version, so a bump also fixes a pin that is already stale. It reuses `bump_all_on_line`, so it validates before writing, replaces every pin on a line (the npm lines carry two), and aborts if a shape stops matching.
- `scripts/validation/gates/check_release_version_coherence.sh`: for each of the same five shapes, every match in `README.md` must be exactly `core/VERSION`. Any pin that isn't gets its own `[FAIL]` line. So does a shape that no longer matches at all.
- `README.md`: produced by running the updated sync with `0.20.36`. That changed 14 pins on 12 lines (3 Kotlin, 5 Flutter including the commented ones, 4 npm, 1 pip, 1 sentence) and nothing else.

## Why `expect_literal` wasn't enough

The binding docs use `bump_all_on_line "$doc" "$CURRENT_VERSION_REGEX"` plus `expect_literal "$doc" "$VERSION"`. That doesn't work for the root README for two reasons:

1. Sync: it only rewrites the *current* version. The README had 0.20.11 and 0.20.29 and no 0.20.36, so the bump would abort with "version pattern not found". Even after a one-time hand fix, it could never repair a pin someone edits later.
2. Gate: `expect_literal` passes once a single occurrence exists. With the second npm pin on the React Native line set back to 0.20.11, `0.20.36` is still on 12 lines of the README, so that check stays green.

## Evidence (run locally)

- `bash -n` passes on both scripts. (shellcheck isn't installed here.)
- Green: on this branch the gate prints `[OK] release version coherence: 0.20.36`.
- Red, one change at a time against the synced README, each exiting 1 with exactly one mismatch:
  - second npm pin on the React Native line -> `[FAIL] README.md: expected 0.20.36, found '@runanywhere/llamacpp@0.20.11'`
  - version-line sentence -> `found 'version line, currently **0.20.29'`
  - commented Flutter line -> `found 'runanywhere_onnx: ^0.20.11'`
- The gate against the current `main` README lists all 14 stale pins (`14 mismatch(es)`).
- The five patterns match exactly those 14 pins. They don't touch `0.83.1+`, `22.12+`, `127.0.0.1`, or the unpinned `pip install runanywhere` on line 98.

I didn't add cases to `scripts/validation/gates/test_release_version_coherence.py`, since CONTRIBUTING asks not to add unit tests unless the change calls for them. I can add README cases there if you want the gate covered the same way as the podspec checks.

## Open question: pins vs. what is actually published

This keeps the README consistent with `core/VERSION` (0.20.36), as the issue asks. But 0.20.36 isn't published on any registry yet. Checked on Sep 15:

- npm: `latest` is 0.20.27 for `@runanywhere/core`, `llamacpp`, `web` and `web-llamacpp`. None of the four has 0.20.36, and none has 0.20.11 either, so those pins already 404 on `main`.
- Maven Central: `runanywhere-sdk` and `runanywhere-llamacpp` top out at 0.20.27. 0.20.11 is there, so the Kotlin pins resolve today and won't after this change.
- pub.dev: `runanywhere` and `runanywhere_llamacpp` top out at 0.20.27.
- PyPI: `https://pypi.org/pypi/runanywhere/json` returns 404.

On #849 you went with pinning to what's published. If the root README should do the same, the sync and gate here need the published version as their source instead of `core/VERSION`. Which should it follow?


ziyaad-mallick:
The two red checks (`android-arm64`, `kotlin-android`) aren't caused by this diff. Both fail in `setup-toolchain` → `android-actions/setup-android@v4`, before any repo code runs:

```
Warning: Failed to find package 'tools'
Error: The process '/usr/local/lib/android/sdk/cmdline-tools/20.0/bin/sdkmanager' failed with exit code 1
```

Google dropped the deprecated `tools` package on about Sept 15 (android-actions/setup-android#537). setup-android fixed it in v4.0.2 on Sept 17 (#538), and the floating `@v4` tag now points past that fix. PR matrix runs for #950 (Sept 15) and #685 (Sept 17) failed the same way, and from Sept 18 on the Android jobs pass again — including runs on `main` today. This run happened on Sept 15, inside that window.

The PR only touches `README.md`, `scripts/release/sync-versions.sh` and `scripts/validation/gates/check_release_version_coherence.sh`, and the `validation` job that runs the coherence gate is green.

I don't have permission to re-run jobs here. Could a maintainer re-run the two failed jobs? They should pass now.

One scheduling note: #886 (rcli → wally) edits the same README section near the version line (around line 431), so whichever of the two lands second needs a small rebase.

